PR #3, opened 2019-06-27
Feature dropout

George3d6:
Added a dropout probability to the data source which should allow us to selectively dropout some of the features before they are feed into the mixer, in order to train the model on cases when some of the features are missing.


George3d6:
Closing due to it already being implemented in a better way by @torrmal on `master`

PR #8, opened 2019-07-09
New MLP mixer

George3d6:
* Added a new MLP mixer which seems to yield better or similar performance on existing datasets and is currently used when the input is > 3000 values. The purpose of this is to allow running datasets where input features are encoded into very long input tensors. This reduced training for the non-public dataset we were concerned about (see slack) from 2+ days and counting to ~20 minutes.

* Added an argument to `.learn` that allows us to stop training after a given number of second (the argument is `stop_training_after_seconds` and defaults to 5 days)

* Added a condition that allows the model to stop training in the edge case where it happens to get close to 100% performance on both the train and test datasets.

* Added a condition that allows the model to stop training if no new lowest error is found after a relatively large number of epochs. Arguable if this is the approach we want to take in the long run, it might be worth trying a new mixer, reseting training with different initial weights or changing hyperparams instead. However, it is a quick fix to the issue of models that can't either converge nor get biased to the training set.

### DIFF
--- a/lightwood/mixers/nn/nn.py
+++ b/lightwood/mixers/nn/nn.py
@@ -102,7 +102,6 @@ class NnMixer:
             # forward + backward + optimize
             outputs = self.net(inputs)
             loss = self.criterion(outputs, labels)
-            loss.backward()
 
             # print statistics
             running_loss += loss.item()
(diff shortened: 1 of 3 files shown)


lightwood/mixers/nn/nn.py
```diff
@@ -102,7 +102,6 @@ def error(self, ds):
             # forward + backward + optimize
             outputs = self.net(inputs)
             loss = self.criterion(outputs, labels)
-            loss.backward()
```
torrmal: why did you remove this?

torrmal: @George3d6 ^

PR #12, opened 2019-07-13
Make install work on windows & linux

George3d6:
* Added platform specific dependencies in `setup.py` to allow the install to work on windows and linux.
* Removed a way of sourcing the version in `setup.py` which would cause a fresh install via `pip3 install -e .` to fail (since a dependency required to import the file in which the version resided needed to be installed beforehand).

### DIFF
--- a/lightwood/__init__.py
+++ b/lightwood/__init__.py
@@ -1,4 +1,6 @@
 import sys
+
+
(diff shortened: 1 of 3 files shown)


George3d6:
Closing this in favor of: https://github.com/mindsdb/lightwood/pull/14

PR #14, opened 2019-07-14
CI tests & Install fix

George3d6:
* Added CI test and a .travis config to run it on all supported platform (could be useful to also add python 3.6 and 3.5 on linux if we wish to support them)
* Added platform specific dependencies in setup.py to allow the install to work on windows and linux.
* Removed a way of sourcing the version in setup.py which would cause a fresh install via pip3 install -e . to fail (since a dependency required to import the file in which the version resided needed to be installed beforehand).
* Added buttons for version & tests to README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,8 @@ cache: pip
 install:
   - pip3 install --upgrade pip
   #- travis_wait pip3 install -e .
-  - travis_wait if [ "$TRAVIS_OS_NAME" = "windows" ]; then python setup.py install; fi
-  - travis_wait if [ "$TRAVIS_OS_NAME" != "windows" ]; then python3 setup.py install; fi
+  - if [ "$TRAVIS_OS_NAME" = "windows" ]; then travis_wait python setup.py install; fi
+  - if [ "$TRAVIS_OS_NAME" != "windows" ]; then travis_wait python3 setup.py install; fi
   - travis_wait pip3 install pytest
 
 before_script: cd tests/ci_tests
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -1,26 +1,58 @@
...
+
+    #dependency_links.append('https://download.pytorch.org/whl/cpu/torch-1.1.0-cp37-cp37m-win_amd64.whl#egg=torch-1.1.0.0')
+    #dependency_links.append('https://download.pytorch.org/whl/cpu/torchvision-0.3.0-cp37-cp37m-win_amd64.whl#egg=torchvision-0.3.0.0')
+    dependency_links.append('https://download.pytorch.org/whl/cu100/torch-1.1.0-cp37-cp37m-win_amd64.whl#egg=torch-1.1.0.0')
+    dependency_links.append('https://download.pytorch.org/whl/cu100/torchvision-0.3.0-cp37-cp37m-win_amd64.whl#egg=torchvision-0.3.0.0')
+
 setuptools.setup(
     name="lightwood",
-    version=lightwood_version,
+    version='0.7.1',
```
torrmal: @George3d6  why here? if anything move it to an  about.py like in mindsdb,  but we may want to use the version number else where in the code base, its a good pracrtice that the version can be read from a common file.

.travis.yml
```diff
@@ -1,15 +1,37 @@
 language: python
 matrix:
   include:
-    - python: 3.5
-    - python: 3.6
```
torrmal: should keep 3.6

PR #15, opened 2019-07-14
Fix MindsDB installation on Google Colab

abitrolly:
Fixes https://github.com/mindsdb/mindsdb/issues/267
because there are no Linux wheels for https://pypi.org/project/torch/1.1.0.post2/#files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ matrix:
 cache: pip
 # install dependencies
 install:
-  - travis_wait pip install lightwood
+  - travis_wait pip install -r requirements.txt
 # run tests
 script:
   - pytest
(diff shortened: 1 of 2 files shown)


George3d6:
I was unaware of the 
```
torch >= 1.1.0.post2; sys_platform == 'darwin'
torch >= 1.1.0;       sys_platform != 'darwin'
``` 
syntax

But, as it stands, I dynamically insert OS & version specific dependencies in `setup.py` and that fixed all the OS-specific bugs, so this PR is somewhat redundant due to #14 :/

Thanks for the tip regarding that requirements.txt syntax and thanks for the intention to help, I'm sorry this coincided with my PR fixing the same thing.

abitrolly:
@George3d6 now you can try to do the same with the [environment markers(https://www.python.org/dev/peps/pep-0508/#environment-markers).

PR #20, opened 2019-07-19
Docs update

ZoranPandovski:
Minor typos fixed int the documentation.


George3d6:
`You would need a python 3.5 or higher.` should just be `You need python 3.5 or higher`, I'll merge like this and correct in the dropout fix PR.

PR #31, opened 2019-08-09
Run on cpu

George3d6:
* The `USE_CUDA` config option is now properly used to train the model on the GPU (when true) and on the CPU (when false). Default value is `False` when no GPU is detected and `True` when a GPU is detected.
* Updated and cleaned up the Sklearn mixer (it wasn't working before), it can now be specified from the config
* Added some fundamental building blocks for automatically constructing basic FCNN shapes
* Generating the basic FCNN in `default_net` based on the size of the input/output
* Added AdamW and training with an lr curve using  AdamW (currently disabled but can be enabled via setting the `adamw` flag to `True`). This seemed to increase training speed and sometimes even accuracy on some models, but I won't enable it for now, introducing this into production might happen after a longer design session with @torrmal 
* Dropped `nan` support, will now crash if a `nan` value is detected in the numerical encoder

### DIFF
--- a/lightwood/mixers/nn/helpers/shapes.py
+++ b/lightwood/mixers/nn/helpers/shapes.py
@@ -23,15 +23,14 @@ def rectangle(in_size,out_size,depth):
 
 def rombus(in_size,out_size,depth,max_size=None):
     if max_size is None:
-        max_size = in_size*2
+        max_size = max(in_size,out_size)*2
     if depth < 3:
         print('Depth must be at least 3 for the rombus function to work correctly, setting it to 3')
         depth = 3
 
     funnel_size = math.ceil(depth/2)
 
-    first_funnel = funnel(max_size,in_size,funnel_size)
-    first_funnel.reverse()
+    first_funnel = funnel(in_size, max_size,funnel_size)
 
     if depth % 2 == 1:
         first_funnel = first_funnel[:-1]
--- a/lightwood/mixers/nn/helpers/shapes.py
+++ b/lightwood/mixers/nn/helpers/shapes.py
@@ -1,0 +1,43 @@
+import math
+
+
+def funnel(in_size, out_size, depth):
+    if depth < 2:
+        print('Depth must be at least 2 for the funnel function to work correctly, setting it to 2')
+        depth = 2
+
+    step_denominator = depth - 1
+    layers = list( range(out_size, in_size, round((in_size - out_size)/step_denominator) ) )
+    layers.reverse()
+    layers = [in_size, *layers]
+    return layers
+
+def rectangle(in_size,out_size,depth):
+    if depth < 2:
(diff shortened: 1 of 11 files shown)


lightwood/mixers/nn/helpers/shapes.py
```diff
@@ -0,0 +1,42 @@
+import math
+
+
+def funnel(in_size, out_size, depth):
+    if depth < 2:
+        print('Depth must be at least 2 for the funnel function to work correctly, setting it to 2')
```
torrmal: please change this to logging.warning

lightwood/mixers/nn/helpers/shapes.py
```diff
@@ -0,0 +1,42 @@
...
+
+    step_denominator = depth - 1
+    layers = list( range(out_size, in_size, round((in_size - out_size)/step_denominator) ) )
+    layers.reverse()
+    layers = [in_size, *layers]
+    return layers
+
+def rectangle(in_size,out_size,depth):
+    if depth < 2:
+        print('Depth must be at least 2 for the rectangle function to work correctly, setting it to 2')
```
torrmal: same logging.warning

lightwood/mixers/nn/helpers/shapes.py
```diff
@@ -0,0 +1,42 @@
...
+
+    layers = [in_size for x in range(depth - 1)]
+    layers = [*layers,out_size]
+    return layers
+
+def rombus(in_size,out_size,depth,max_size=None):
+    if max_size is None:
+        max_size = max(in_size,out_size)*2
+    if depth < 3:
+        print('Depth must be at least 3 for the rombus function to work correctly, setting it to 3')
```
torrmal: logging.warning

lightwood/mixers/nn/nn.py
```diff
@@ -27,8 +28,14 @@ def __init__(self):
 
         self.criterion = nn.MSELoss()
         self.epochs = 120000
-        self.optimizer_class = optim.Adadelta
-        self.optimizer_args = {'lr': 0.1}
+
+        if self.dynamic_adamw:
```
torrmal: this will never check True, as the class is initialized and then self.*<-->config attributes are mapped.

George3d6: Ah yes, I know, I left it in there for now so that I can play around with it (and maybe ask other people to enable it for internal testing). I want to read a bit more about how to adapt the learning rate and how to get very-fast convergence using adamw before I expose it to normal users and allow it to be enabled via config or based on the data.

lightwood/mixers/nn/helpers/shapes.py
George3d6: fixed

PR #34, opened 2019-08-21
Img from url

George3d6:
Allow urls to be used to image instead of paths


George3d6:
small change, so merging w/o review

PR #35, opened 2019-08-25
Unbalanced loss

George3d6:
* Added loss functions that accepts weights for each of the target values as part of the configuration
* Fixed the way the data source encodes values, due to an error there, dropout and any other custom value based encoding wasn't working. This means dropout was actually NOT working until now, hopefully now that it's working this will provide some improvement for missing column scenarios and improve accuracy overall


George3d6:
As all points raised during the review were addressed, I will go ahead and merge

PR #37, opened 2019-09-08
Various major changes to model building & training

George3d6:


lightwood/mixers/nn/nn.py
```diff
@@ -1,48 +1,164 @@
...
-import torch
 import math
+import itertools as it
+
+import torch
+from torch.optim.optimizer import Optimizer
+
+
+
+class Ranger(Optimizer):
```
torrmal: moce this to another helper file and import it

George3d6: oh, yes... I kept it in nn.py for easier debugging and forgot to move it, will do

PR #48, opened 2019-10-15
Large data

### LINKED ISSUE #46: AX Optimization causes transofrmer error during training
Sometimes, when running the ax optimization (happened on a proprietary dataset, doesn't seem to happen on other),  transformer's `self.feature_len_map` mysteriously gets cleared before the first call to the callback function during train.

I'm not sure what's causing this, we need to look into it further.

### LINKED ISSUE #47: ValueError: Input contains NaN, infinity or a value too large for dtype('float64')
Your Environment

    Python version: 3.7.4
    Pip version: 19
    Operating system: Ubuntu 18.04
    Python environment used (e.g. venv, conda): venv
    Mindsdb version you tried to install: 1.6.8

Describe the bug
ValueError is thrown when training

To Reproduce
Steps to reproduce the behavior, for example:

    Use this example
    You should see the error: ValueError: Input contains NaN

Additional context
Screenshot from 2019-10-08 13-41-47: https://user-images.githubusercontent.com/7192539/66394176-aa3d6680-e9d4-11e9-9e1b-d61fe50b26c4.png

George3d6:
Main feature:

* Allow for cache disabling via the `CONFIG`, this is useful for running large dataset, it does decrease speed but it doesn't seem to affect it that significantly. We should control this option dynamically from mindsdb in the future based on the computer's available memory and the dataset size (PR incoming tonight or tomorrow).

In the process of doing this I've removed the option to disable the cache for a particular column only, since it proved to cumbersome to implement properly and it was hardly working as intended. Might be worth looking at implementing it again when we have more time and/or when we restructure some of the dataset logic. But I don't think now is the correct moment.

Various fixes:

* Changed the r2 accuracy score for numerical target variables to use the encoded values, which fixed #47. Unlike the previous implementation though, we take the decoded predictions and encode them, that way flags (the positions that represent isnull, iszero and sign) will be either 0 or 1, rather than `0.0x` or `9.9x`, which was decreasing the r2 score significantly.

* Changed an internal logic of fitting a data source to a model so that the transformer doesn't get replaced with a new one every time, if the data source already has a trasnformer. This fixes #46 

* Fix to accuracy score so that it works well when the panda df interprets one of the categorical column as a number

### DIFF
--- a/lightwood/api/predictor.py
+++ b/lightwood/api/predictor.py
@@ -327,7 +327,6 @@ class Predictor:
         return self
 
 
-
     def predict(self, when_data=None, when=None):
         """
         Predict given when conditions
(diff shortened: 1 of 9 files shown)


lightwood/api/predictor.py
```diff
@@ -68,7 +70,7 @@ def evaluate_mixer(mixer_class, mixer_params, from_data_ds, test_data_ds, dynami
 
         if max_training_time is None and max_epochs is None:
             logging.error("Please provide either `max_training_time` or `max_epochs` when calling `evaluate_mixer`")
-            exit()
+            sys.exit(1)
```
torrmal: should we thorw an exception instead?

George3d6: yes, fixed

PR #50, opened 2019-10-28
Bayesian NN mixer

George3d6:
Added a bayesian NN mixer to lightwood, in addition to some re-factoring to remove a bit of useless code from the default `NnMixer` and share helpers between the `BayesianNnMixer` and the `NnMixer`.

***

**Important caveat:** 

While the network seems to be "working" it performs very poorly compared to the default mixer on the datasets on which I've tested it, I recommend reviewing the training code and seeing if you spot any obvious errors. 

Otherwise, I'd say we should merge this even though it has the bad accuracy, so that we don't have to keep a separate branch, this mixer is not used by default so it won't interfere with anything for now.   

I will look into making it get better accuracy next week.

Reference used for the implementation:

https://github.com/paraschopra/bayesian-neural-network-mnist
https://github.com/rakshitsareen/Bayesian-Neural-Networks
https://towardsdatascience.com/making-your-neural-network-say-i-dont-know-bayesian-nns-using-pyro-and-pytorch-b1c24e6ab8cd
http://docs.pyro.ai/en/stable/index.html

***

The Bayesian NN Mixer, in addition to returning an array of predictions (under the `predictions` key) in the output dictionary, will aslo return:

* output_dict -> output_column -> confidence   |   Confidence in the prediction (from 0 to 1)
* output_dict -> output_column -> possible_predictions   |  A list of all possible predictions according to the various pyro models
* output_dict -> output_column -> possible_predictions_confidence   |   Confidence for each of the predictions in the `possible_predictions` key

The predictions returned under the `predictions` key will be those in which the network has the highest confidence

To test the network and see the output run: `python3 tests/development/bayesian_nn_test.py`

Additional notes:

* The implementation for numerical predictions is kinda "demo"~ish in nature, partially because I wanted to test on a few numerical datasets to see what approach would work best (in terms of histogram bins and whether or not to join buckets together when confidence is low and take the confidence-adjusted mean)

* Might be worth looking into this contributor extension to pyro: http://docs.pyro.ai/en/stable/contrib.bnn.html  --  I did not managed to get it to work and found no examples on a summary search of the internet, so I gave up for the moment but I'll get back to it at some point.

* While Pyro is open source, it seems to be mainly contributed to by Uber AI employees and documentation/examples/discussions are not 1/1000 as plentiful as those for Pytorch. One thing I'm afraid of here is that this library is to "niche" to rely upon as the backbone for our main mixer. In addition to that it uses very high level abstractions and doesn't provide a lot of hooks into them (e.g. it's not painless to adjust a custom optimizer to work with it). But I don't see this as a killer flaw, just a red flag in moving forward with using it.


lightwood/mixers/bayesian_nn/bayesian_nn.py
```diff
@@ -0,0 +1,307 @@
...
+        when_data_source.transformer = self.transformer
+        when_data_source.encoders = self.encoders
+        data_loader = DataLoader(when_data_source, batch_size=len(when_data_source), shuffle=False, num_workers=0)
+
+        self.net.eval()
+        data = next(iter(data_loader))
+        inputs, _ = data
+        inputs = inputs.to(self.net.device)
+
+        sampled_models = [self.pyro_guide(None, None) for _ in range(len(inputs))]
```
torrmal: i dont understand, this should not be a function of the len of inputs. but rather function of a constant NUMBER_OF_MODELS to generate a distribution from (lets start with 100)

torrmal: @George3d6 ^

George3d6: Yes... I'm not sure how that slipped in there, and the length of the input was rather small so I didn't catch it

PR #53, opened 2019-10-31
Fixes for: No cache, memory leak, network shape

torrmal:


lightwood/mixers/helpers/default_net.py
```diff
@@ -46,21 +54,23 @@ def __init__(self, ds, dynamic_parameters):
...
         elif (not large_output) and large_input:
             shape = funnel(self.input_size,self.output_size,depth)
 
         elif (not large_input) and large_output:
             shape = rectangle(self.input_size,self.output_size,depth - 1)
-
         else:
             shape = rectangle(self.input_size,self.output_size,depth - 2)
 
+        print(shape)
```
torrmal: print

George3d6: Was still debugging this branch and adding stuff, removed now

George3d6:
* Separate the encoder preparation into a `prepare_encoders` method of the data source which should be called as soon as possible during training on the training datasource. In the previous implementation this was happening during the first pass of a column through `get_encoded_column_data`, but when we we're training without cache (and thus we only feed data to `get_encoded_column_data` this resulted in encoders being prepared with only a single value)

* Fixed a memory leak during testing and error evaluation caused by torch/autograd storing gradients for the model on predictions and error determination forward passes (`predict` and `error`), these aren't used, since there's no backwards pass and they consume memory needlessly. Noticeable speedup in training and prediction speed on large datasets, especially on the GPU.

* Smaller changes:
   1. Removed some branching which was never used in network shape determination. 
    2. Explicitly  setting network in eval and train mode when appropriate. This disables things like dropout layers, not needed for now but good to have in case we add anything where eval and train behavior are different, it was already used in one place so I though either use it everywhere or remove it.
   3. Added some comments to the code that was added/modified

PR #54, opened 2019-11-01
Plinear

torrmal:
It implements a probabilistic linear activation function such that the weights are sampled from a normal distribution, and the backpropagation step tries to optimize the variance and mean of each weight.


lightwood/mixers/helpers/default_net.py
```diff
@@ -74,7 +75,7 @@ def __init__(self, ds, dynamic_parameters):
 
         layers = []
         for ind in range(len(shape) - 1):
-            layers.append(torch.nn.Linear(shape[ind],shape[ind+1]))
+            layers.append(PLinear(shape[ind],shape[ind+1]))
```
George3d6: We should not remove the Linear layer completely, instead, we should add a flag to toggle the linear layer on/off or mass `layer_type` to the default network when we initialize it (I think the former is better).

If we decide that the PLinear layer is better we can make it a default later.

lightwood/mixers/helpers/plinear.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        if self.w_sampler == self.w_normal:
+            init.kaiming_uniform_(self.mean, a=math.sqrt(5)) # initial means (just as in original linear)
+            init.uniform_(self.sigma, a=0.01, b=0.1) # initial sigmas from 0.5-2
+        elif self.w_sampler == self.w_discrete_normal:
+            init.kaiming_uniform_(self.mean, a=math.sqrt(5))  # initial means (just as in original linear)
+            init.uniform_(self.sigma, a=0.05, b=0.2)
+
+        if self.bias is not None:
+            fan_in, _ = init._calculate_fan_in_and_fan_out(self.mean)
```
George3d6: Is this bias initialization any good for FC layers ? As far as I know this method is usually used with CNNs.
Either way, minor point, if you don't have strong references on whether or not it works either we can just test different bias initialization later and see if it makes any difference.

lightwood/mixers/helpers/plinear.py
```diff
@@ -0,0 +1,171 @@
...
+            bound = 1 / math.sqrt(fan_in)
+            init.uniform_(self.bias, -bound, bound)
+
+    def w_discrete_normal(self):
+        """
+        Sample a w matrix based on a discrete normal distribution
+        :return: w
+        """
+
+        bucket_seed = random.uniform(0,1) # gett a number that can help determines which bucket of the distribution can numbers fall in
```
George3d6: The code from here to line 81 could be replaced with:

`sigma_multiplier = np.random.choice([1,2,3],p=[0.33,0.47,0.2])`

and it would probably make it easier to understand and shorter.

lightwood/mixers/helpers/plinear.py
```diff
@@ -0,0 +1,171 @@
...
+    def forward(self, input):
+        '''
+        Forward pass of the function.
+        The goal is to generate values such as if they weights of the linear operation are sampled from a normal distribution
+        '''
+
+        return F.linear(input, self.w_sampler(), self.bias)
+
+
+if __name__ == "__main__":
```
George3d6: I assume this was an old test, maybe replace with a test for the PLinear class itself instead :?

lightwood/mixers/helpers/plinear.py
```diff
@@ -0,0 +1,171 @@
...
+            if bucket_seed > 0.8:
+                sigma_multiplier = 4
+
+        # generate the initial tensor, this will ultimately transforms in to the weights
+        w = torch.Tensor(self.out_features, self.in_features)
+        # make sure that they are evently distributed between -1, 1
+        init.uniform_(w, a=-1, b=1)
+
+        # adjust based on sigma
+        w = self.mean*( 1+ w * torch.abs(self.sigma) * sigma_multiplier )
```
George3d6: How do we go about extracting some sort of confidence / interval of possible prediction in the last layer (which is presumably a big part of the goal here). Do we add an option to remove the `self.mean*` operation for the last layer and then look at the probability distribution here ?

lightwood/mixers/helpers/default_net.py
torrmal: agree

lightwood/mixers/helpers/plinear.py
torrmal: I put it there just to make sure that it supports the same API that normal linear does, so it can be swapped seamlessly

lightwood/mixers/helpers/plinear.py
torrmal: you need to sample results at the end to calculate the probability distribution of the net outcome

tests/ci_tests/full_test.py
```diff
@@ -38,4 +41,31 @@ def iter_function(epoch, error, test_error, test_error_gradient, test_accuracy):
 
 predictor.learn(from_data=df, callback_on_iter=iter_function, eval_every_x_epochs=2, stop_training_after_seconds=120)
 
-print(predictor.predict(when={'number_of_rooms':3, 'number_of_bathrooms':2, 'sqft':700, 'location':'great'}))
+#print(predictor.predict(when={ 'number_of_bathrooms':2, 'sqft':700, 'location':'great'}))
+
+predictor.save('/tmp/test.lw')
```
George3d6: oh, you have to use `os.join` to build the path here else it fails on windows, see the travis error

tests/ci_tests/full_test.py
```diff
@@ -38,4 +41,31 @@ def iter_function(epoch, error, test_error, test_error_gradient, test_accuracy):
...
+
+#exit()
+predictor = Predictor(load_from_path='/tmp/test.lw')
```
George3d6: Use `os.join`

lightwood/mixers/helpers/plinear.py
George3d6: Oh, ok, linear does indeed do the same thing by default in pytorch: https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/linear.py#L82, my bad

George3d6:
Fails when training on GPU, pressumably some tensors aren't set the the correct device

PR #57, opened 2019-11-14
Oversample

George3d6:
Added option for lightwood to handle unbalanced dataset via over sampling.

Minor changes:
* Using the mean of all other weights as the default weight for values not present in the weight map.
* Arranged and commented config flags

### DIFF
--- a/lightwood/config/config.py
+++ b/lightwood/config/config.py
@@ -2,16 +2,19 @@ import torch
 
 
 class CONFIG:
+    """General"""
     USE_CUDA = False
     if torch.cuda.device_count() > 0:
         USE_CUDA = True
 
     USE_DEVICE = None
+    USE_CACHE = True
+    # Enable deterministic cuda flag and use seeds everywhere (static or based on features of the dataset)
+    DETERMINISTIC = True
+    OVERSAMPLE = True
 
-    DETERMINISTIC = False
-
+    """Probabilistic FC layers"""
     USE_PROBABILISTIC_LINEAR = False # change weights in mixer to be probabilistic
 
-    USE_CACHE = True
-
+    """Bayesian Network"""
     NUMBER_OF_PROBABILISTIC_MODELS = 2
--- a/lightwood/config/config.py
+++ b/lightwood/config/config.py
@@ -11,8 +11,7 @@ class CONFIG:
     USE_CACHE = True
     # Enable deterministic cuda flag and use seeds everywhere (static or based on features of the dataset)
     DETERMINISTIC = True
-    #OVERSAMPLE = True
-    OVERSAMPLE = False
+    OVERSAMPLE = True
 
     """Probabilistic FC layers"""
     USE_PROBABILISTIC_LINEAR = False # change weights in mixer to be probabilistic
--- a/lightwood/config/config.py
+++ b/lightwood/config/config.py
@@ -8,10 +8,10 @@ class CONFIG:
         USE_CUDA = True
 
     USE_DEVICE = None
-    USE_CACHE = True
+    CACHE_ENCODED_DATA = True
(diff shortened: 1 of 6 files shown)


lightwood/config/config.py
```diff
@@ -2,16 +2,19 @@
 
 
 class CONFIG:
+    """General"""
     USE_CUDA = False
     if torch.cuda.device_count() > 0:
         USE_CUDA = True
 
     USE_DEVICE = None
+    USE_CACHE = True
```
torrmal: what are we caching?

lightwood/mixers/nn/nn.py
```diff
@@ -24,6 +26,7 @@ def __init__(self, dynamic_parameters, is_categorical_output=False):
         self.optimizer_class = None
         self.optimizer_args = None
         self.criterion = None
+        self.sampler = None
```
torrmal: do we really ant to serialize this on model save, i dont think I am understanding this PR @George3d6

lightwood/config/config.py
George3d6: This is just a name I re-arranged in the config file, it was here before under the name `USE_CACHE`, arguably could be renamed to `CACHE_ENCODED_DATA`, since it's the toggle for caching encoded data.

Will change it.

lightwood/mixers/nn/nn.py
George3d6: I guess it doesn't need to be saved with the model, I didn't realize that implication of making it part of `self`, I just wanted it to be used by both the `iter_fit` and `error` methods without having to re-create it each time.

I will explain the purpose of the PR bellow.

George3d6:
So, to be more explicit as per your comment @torrmal , the purpose of the PR is  to allow using pytorch's oversampling sampler to handle unbalanced datasets.

That is to say, instead of using `weights` for the loss function or adding more rows with the under-represented target variable in mindsdb, we use oversampling via the DataLoader (by making it sample via the `WeightedRandomSampler` being created).

This is preferable to adding more rows with the under-represented target variable since it doesn't incur the same memory cost but provides the same beneifts.

This can sometimes pe preferable to using weights in the loss function, despite being somewhat slower, since if our underrepresented target variable is very scarce (e.g. 1 in 5000), then the weighted loss function will not be equivalent to giving the underrepresented target variable equally accuracy to the over-represented one. 

This is because we might only see the under-represented target variable every few batches  (in this case, since we use batches of 200, only once every 20 batches), so we end up with some batches ( in this example 19 out of 20 batches) over-fitting the gradient towards the over-represented variables and some batches ( in this case 1 out of 20) over-fitting the gradient towards the under-represented variable (since it's weight is so high). 

Arguably, even if the difference between batch size and underrepresented variable occurrence is smaller (e.g. we see the underrepresented variable in 1 out of 400 rows and the batch size is 400), this might still lead to worst performance than over-sampling, but here I'm going on hear-say mainly, I don't know how pytroch's batch gradient computation works well enough to defend this claim.

Overall, if you aren't sure this approach has credit we can leave it in an only enable it manually for the time being.

George3d6:
Also, one thing I should note: Training with oversampling enabled significantly improves the accuracy on two of the very unbalanced benchmark datasets: Default on Credit and Pulsar Star (by more than 10% that is). So "in practice" the changes seem to have the intend effect.

George3d6:
So, sampler should no longer be saved now, check manually and it's not in the saved object. Also added a new test case to validate that we can re-train from the saved model (before we were only doing this for predict).

Not pickling the value could be done in a fancier way by modifying the state get/set method of the objects (see https://stackoverflow.com/questions/2345944/exclude-objects-field-from-pickling-in-python) ... but I don't see it providing any benefits and it could cause trouble with other things (e.g. copying), so I think this solution is simpler and preferable.

PR #59, opened 2019-11-22
Memory aware model

George3d6:
**Important**: This PR will have to be refactored a bit after we merge the Oversample and Selfaware PR, I recommend merging those first as this PR is a bit more complicated and arguably less relevant to our goals (getting explainable models) than Selfaware and improves accuracy on benchmark datasets less than Oversample.

In short:

With this PR the architectures we select for a model are now dependent on how much memory the user has left on the device he chose (cpu == ram and cuda == memory of a given gpu).

More in depth:

* Made some corrections to the way we generate a "funnel" shape for the network, this shouldn't affect most models but the error came in when we had a network with small depth (<4), very large input size and very small input size.

* Implemented an algorithm that iterates through ever-increasing network sizes until either:
a) We run OOM (in which case we safely recover and use the last largest network size we got)
b) We end up with more than 40% of the memory on the device being taken up by the model
c) We reach a "maximum" size for the network past which it would help very little to increase it

Initially we iterate through network shapes: Funnel, Square and then rombus... so essentially (for most cases):

```
# Example here corresponds to a network of 4 inputs and 2 outputs and showcases how we would "grow" the shape with this PR if available memory existed on the machine
                                                                                              *
   
                                                                                              *

*                                            *       *                                *       *      * 
      *                                       
*           *                                *       *      *                         *       *      * 
      *                  =>                                             =>          
*           *                                *        *     *                         *       *      * 
      *
*                                             *        *                               *       *      * 

                                                                                                *
                                                                                                *
```

After that, we grow the depth of the network from 3 layer (1 hidden) to 6 (4 hidden). Once we reach that point, even if there's enough memory left, we stop

* Benchmarked this new implementation, it gives better results on CIFRAR1000 and worst/slightly worst results on the default on credit, pulsar star and cancer50 datasets. However, these results are being compared to the Oversample branch (which is the one that obtains oversample results). I'd like to benchmark this after merging with Oversample and see if the results end up being better or worst.

* Overall, there's room for improvement, the memory usage and available memory determination algorithms are pretty weak at the moment. However, determining available memory is more complicated than I thought and requires some extra dependencies (some of which can't be installed via pip, e.g. nvidia-sim), so before sinking even more time into it, I'd rather see the performance with the current one once merged with Oversample and deciding whether or not this approach is overall a good one to go with.

### DIFF
--- a/lightwood/__about__.py
+++ b/lightwood/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'lightwood'
 __package_name__ = 'mindsdb'
-__version__ = '0.11.3'
+__version__ = '0.11.4'
 __description__ = "Lightwood's goal is to make it very simple for developers to use the power of artificial neural networks in their projects."
 __email__ = "jorge@mindsdb.com"
 __author__ = 'MindsDB Inc'
--- a/lightwood/__about__.py
+++ b/lightwood/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'lightwood'
 __package_name__ = 'mindsdb'
-__version__ = '0.11.6'
+__version__ = '0.11.7'
 __description__ = "Lightwood's goal is to make it very simple for developers to use the power of artificial neural networks in their projects."
 __email__ = "jorge@mindsdb.com"
 __author__ = 'MindsDB Inc'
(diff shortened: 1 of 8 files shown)


George3d6:
Closing this as relevant changes from this branch have already been merged as part of other branches.

PR #62, opened 2019-11-28
Selfaware and autoencoding fixes

George3d6:
Benchmarks are still running for this PR, I'll update it with the results when they are finished.

### Selfaware related (Rather important)

* If the losses get too large (inf or nan) we halve the learning rate and re-start training at that (resetting all the weights)
* If the overall error gets small enough (arbitrarily picked as `< 1`), then we switch the loss combination operator that combines the awarenessloss with the prediction loss is switched from 
* Extracting a confidence for the prediction from the awareness output and giving it to the user (this is computed at line 93 in nn.py), this is probably the most critical thing to review and possibly suggest changes to.

### Autoencoder related changes (Not that important)

Basically, the categorical autoencoder wasn't playing along very well with a lot of the unbalanced-categories related features and with the CrossEntropy loss we use for categorical output. Rather than making the necessary changes to adjust of auto-encoded categorical output, I thought that it might be better to just use one-hot for categorical outputs indifferent of their dimension. In order to achieve this three changes were made:

* The `is_target` argument was added to all the encoder constructors (was already present in most anyways)
* The encoder initialization now passes the `is_target` argument as `True` when initializing an encoder for an output column
* The categorical autoencoder always falls back to one-hot encoding when `is_target == True` (previously it only did this if the categorical column contained less than 100 distinct values, see `max_encoded_length`).

### DIFF
--- a/lightwood/mixers/nn/nn.py
+++ b/lightwood/mixers/nn/nn.py
@@ -6,6 +6,7 @@ import torch.optim as optim
 from torch.utils.data import DataLoader
 import numpy as np
 import gc
+import operator
 
 from lightwood.mixers.helpers.default_net import DefaultNet
 from lightwood.mixers.helpers.transformer import Transformer
@@ -33,7 +34,7 @@ class NnMixer:
         self.nn_class = DefaultNet
         self.dynamic_parameters = dynamic_parameters
         self.awareness_criterion = None
-
+        self.loss_combination_operator = operator.add
 
         self._nonpersistent = {
             'sampler': None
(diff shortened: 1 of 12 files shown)


lightwood/mixers/nn/nn.py
```diff
@@ -67,29 +70,42 @@ def predict(self, when_data_source, include_encoded_predictions = False):
...
 
         output_trasnformed_vectors = {}
-
-        for output_vector in outputs:
+        confidence_trasnformed_vectors = {}
+
+        for i in range(len(outputs)):
+            if awareness_arr is not None:
+                confidence_vector = awareness_arr[i]
+                transformed_confidence_vectors = when_data_source.transformer.revert(confidence_vector,feature_set = 'output_features')
```
torrmal: not sure, what the logic is here, but the confidence vector isn't the actual value but rather the distance to the target, lets discuss intervals

George3d6:
Benchmarks are done running and it seems that this PR would cause a considerable drop in performance on both the default of credit and cancer50 datasets. I think it's worth investigating and fixing this before merging

George3d6:
Ok, after some small logic and magic number changing I got it to perform about the same as lightwood 0.11.5 (the oversample version with not selfaware), still getting worst performance on cifrar and default on credit, but not that significantly worst (accuracy down by about 3% on doc and 4% on cifrar) but actually better performance of cancer (+2.5%) and pulsar start (+3%).

So on the whole, I think this should safely bring back lightwood to about the same level of accuracy as before the selfaware changes.

Ideally, we'd obviously like it to be better in all cases, and dynamically lowering the learning rate is obviously not a free performance gain since it increases training time, but I think I'd be comfortable merging this for now.

PR #79, opened 2019-12-05
Fix windows install

### LINKED ISSUE #28: Windows install via pypi repo fails
`pip install lightwood` will fail on windows due to us sourcing the torch dependency from outside of pypi.

Not much we can do about it at the moment, since the pypi version of torch for windows seems to fail installing most of the time (or install a version that's way too old).

we should look into fixing when new releases of torch arrive on pypi and/or new updates for the various libraries involved with torch appear on windows.

Temporary "fix" is just asking people to install from github.

ZoranPandovski:
This PR fixes #28 issue for Windows installation. It is critical to have lightwood installation works with `pip` because the mindsdb GUI depends on it, and many users are using it on Windows. Currently, this is tested through TestPyPI and it works.


George3d6:
The old install methodology was there because installing pytroch via pypi on windows wasn't working. This seems to pass the unit-tests now, and if you're saying you tested it on your window machine and it does work... maybe it was fixed (?).

If this is so then this would make the whole install process much simpler, I'll leave it for @torrmal to merge though. If this works we can and should also update install instructions for mindsdb on windows (since it would allow installing from pypi directly, as previously we had to install via github, since lightwood had to be installed via github, since it had the 3rd party torch and torchvision dependencies).

When you tested the install via TestPyPi on your machine, did you start with a new environment and/or remove the previous installations of pytroch ? (@ZoranPandovski , that is)

ZoranPandovski:
Yes, now following the PyTorch official documentation, users can install (torch and torchvision) on windows with providing -find link parameter. For testing purposes, I had a new virtual environment without any dependency installed. Also, this is checked in Linux too just in case this didn't break anything.
About the documentation you are right, I will need to update it because we are specifying install from GitHub.

George3d6:
Ok then, nice find, I completely missed this. Please also update the version in `__about__.py` as with any other PR,  make it `0.13.0` since it's a major user-facing change as it modifies how we are installing and if you have time, once this is merge, update the mindsdb docs.

@torrmal unless you have anything against this please merge as soon as possible, it would help a lot to be able to install via pip on windows again.

PR #81, opened 2019-12-09
Transform text encoders

George3d6:
Text encoder based on DistilBERT from huggingface (basically a BERT implementation but made to have fewer weights and thus quicker).

At the moment it behaves in two possible ways:

a) If the target is anything but a single categorical variable, it generates a vector of 768 embeddings for each value in  a text column. These seem to be pretty decent for training the mixer, they allowed for 84% accuracy on the imdb movie reviews dataset, similar to ludwig, where as the previous encoder got us to ~73% and was much slower

b) If the target is a single categorical variable, train a model to predict the target and feed the logits into the mixer (note, this only works for one-hot encoding, but we only use one-hot encoding for categorical output variables at the moment anyway). In theory this could be extended to more output types, but getting it to work well in practice is trickier.  I'm running some benchmarks on this approach now, it seems to perform decently on a sample of imdb movie reviews.

Additional changes:

* Increase mixer's network size if outputs & inputs are very small.
* Added a `training_data` optional argument to `prepare_encoder`.
* Added the ability to pass the encoded/decoded target to the input encoders when the datasource prepares all encoders.

### DIFF
--- a/lightwood/__about__.py
+++ b/lightwood/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'lightwood'
 __package_name__ = 'mindsdb'
-__version__ = '0.12.3'
+__version__ = '0.13.2'
 __description__ = "Lightwood's goal is to make it very simple for developers to use the power of artificial neural networks in their projects."
 __email__ = "jorge@mindsdb.com"
 __author__ = 'MindsDB Inc'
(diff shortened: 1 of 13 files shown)


George3d6:
Hmh, merging stuff around I seem to have reverted some important changes, opening a new PR from a new branch.

PR #85, opened 2019-12-13
Update docs

ZoranPandovski:
- Remove specific install option for windows
- Add PyPI download stats for lightwood


George3d6:
Hmh, a version update here will actually cause the tests to fail again, because it will try to redeploy to pypi but no changes => it will fail to deploy a new version.

We need to figure out some workaround this... maybe. For now I'll just merge your branch into the one I'm about to PR and we should be fine.

George3d6:
Closing, will merge together with the next PR, see above.

Issue #73, opened 2019-12-03
K-Fold cross validation


torrmal:
Lets do k fold cross validation instead of https://github.com/mindsdb/lightwood/issues/67

read:

https://machinelearningmastery.com/k-fold-cross-validation/

George3d6:
Posed a few questions in slack regarding this, will formalize the implementation I plan once those are answered.

George3d6:
Implemented as of #101 merge

Issue #95, opened 2019-12-17
Weird segfault issue on import


George3d6:
Basically, there's a weird segfault happening when importing lightwood... in some cases.

Examples:
```
Segmentation fault:
import transformers

Segmentation fault:
import mindsdb
import lightwood

Segmentation fault:
import mindsdb
import transformers

Works:
import lightwood
import transformers

Works:
import lightwood
import mindsd
```
Doesn't happen on all machine. No idea of the cause, investigating now.

George3d6:
By a weird stroke of luck, this bug is fixed if lightwood is imported before mindsdb... which we now do in the `__init__.py` of mindsdb. So all the cases in which the segfault happens have been fixed.

The reason as to why this is the case though...  I still do not know that :/

Leaving the issue open, though maybe it should be mindsdb related.

George3d6:
So, some extra help here: https://github.com/pytorch/pytorch/issues/4111

Also, interesting that this crashes:

```
>>> import sklearn
>>> import torchvision
```

Might be an issue related to the latest torch/trochvision release.

George3d6:
It seems like a good way of fixing this is making sure that  `torchvision` is always imported first and foremost, that fixes all other crashes seemingly.

George3d6:
Note: Check whether the new torch/torchvision version fixes this on all environments.

ZoranPandovski:
@George3d6 Sure, I will add update if something popsup.

George3d6:
Should be fixed now.

PR #101, opened 2020-01-06
K-fold cross validation

### LINKED ISSUE #93: Add generic training loop
Currently a lot of the elements from the training look happening inside the nn mixer's `iter_fit` and in the callback that the `predictor` api passes to it are re-used in the categorical autoencoder and the text encoder. It might be worth while abstracting away a few of the things an re-using them in all 3 places.

George3d6:
Resolves #93 and #73

### #93 (Add generic training loop) related
1. Added a `Gym` class which contains a basic environment with the objects needed to train a model (model, optimizer, scheduler, loss function...etc) and a `fit` function that, given some data, trains the actual model (very similar logic to that in `predictor.py`).
2. Used said `Gym` class to simplify the training logic for the text and categorical neural-network based encoders.

### #73 (K-Fold cross validation) related
3. Added a `SubSet` class that allows us to treat only some indexes of a `Dataset` as a dataset in of themselves. This class is mostly a view over the dataset it's constructed on and should have very little memory or access overhead (note, some similar things already existed, but I tried using them and they were lacking in various ways)
4. Added `create_subsets` that creates a given number of random subsets from a `Dataset`
5. Switched up training in `predictor.py` to train the model on these subsets (one train subset and one test subset, containing mutually exclusive rows from the other subsets). Currently we train with each subsets once until we reach our usual stop-training condition on each and then we train again on the full dataset until the stop-training condition is reached.  I don't think this is an ideal approach, but it seems to yield good accuracy results, however it's rather slow. I wanted to discuss a bit more with @torrmal  as to how exactly we want to train the model using these subsets.

## Todo

* Add benchmark results... not ready yet, since I wrestled some environment-related issues today, running now and should be included by tomorrow.
* Decide how many subsets we want to have and how we want to loop through them in order to train the model. I'm thinking a good approach would be:
1. Train+test on every subset pair until we reach a "decent' accuracy
2. Train and test on every subset pair until we reach the stop-condition (~0 loss or model stops improving over a number of iterations) on the test subset
3. Get loss & accuracy on the test set, if this is close to the loss & accuracy we got on the independent subsets, the model is good. If not maybe go back to step 2. and re-itterate number of times.

But I'm not sure this approach is ideal, again @torrmal , I'm curios to hear what you think we should do here.


lightwood/__init__.py
```diff
@@ -1,3 +1,4 @@
+import torchvision
```
torrmal: why this?

lightwood/api/predictor.py
```diff
@@ -177,15 +173,18 @@ def type_map(col_name):
         mixer_params = {}
 
         if 'mixer' in self.config:
-            mixer_class = self.config['mixer']['class']
+            if 'class' in  self.config['mixer']:
+                mixer_class = self.config['mixer']['class']
             if 'attrs' in  self.config['mixer']:
                 mixer_params = self.config['mixer']['attrs']
         else:
             mixer_class = NnMixer
```
torrmal: perhaps move this to line 174

torrmal: mixer_class = NnMixer
mixer_params = {}

if 'mixer' in self.config: ...

lightwood/__init__.py
George3d6: This change was made as a hack in order to fix this issue: https://github.com/mindsdb/lightwood/issues/95

We already do this in `mindsdb` in order to not have it crash (since it's easier to update that rather than update lightwood), but this fixes the issue in case we use lightwood independently.

It's a hack, but it works, might not be required once we update to the next torch version, but it's pretty harmless so I'd rather have it in for now.

Forgot I added it to this PR so I didn't explain this in the description, my bad.

lightwood/api/predictor.py
George3d6: Sure

PR #106, opened 2020-01-15
Extra mixers

George3d6:
* Added a gradient boosting mixer for lightwood, only predicts the categorical and numerical outputs.
* Using said gradient boosting mixer when it's accuracy on the testing dataset is 5% higher than that of the nn mixer
* Corrected the autoencoder data loader to handle columns of mixed string&int type and no longer rely on the torch DataLoader


George3d6:
Closing in favour of #112

PR #107, opened 2020-01-16
[WIP] Fix lint issues

djKooks:
Hello,
this is the result PR of lint-fix.

Please give suggestion on configuration, and settings for lint-check.

Thanks.


George3d6:
@djKooks looking at the linter's result, it seems that nothing interesting was caught by it. It's just modifications adding and removing a ` ` or `\n` characters.

lightwood/__init__.py
```diff
@@ -1,16 +1,15 @@
+from lightwood.__about__ import __package_name__ as name, __version__
+from lightwood.encoders import BUILTIN_ENCODERS
+from lightwood.mixers import BUILTIN_MIXERS
+from lightwood.api.predictor import Predictor
+import lightwood.model_building
+import lightwood.constants.lightwood as CONST
```
George3d6: their old location was as it was intended

djKooks:
@George3d6 I have some question while updating.

1. This is some issues shows in linter(ignore line number...):
```
...
lightwood/encoders/text/rnn.py:17:33: F405 'torch' may be undefined, or defined from star imports: lightwood.encoders.text.helpers.rnn_helpers
lightwood/encoders/text/rnn.py:30:28: F405 'Lang' may be undefined, or defined from star imports: lightwood.encoders.text.helpers.rnn_helpers
...
Do you prefer using star imports, or it will be good to fix this?

2. This is another lint issue:
```
...
lightwood/model_building/__init__.py:1:1: F401 'lightwood.model_building.basic_ax_optimizer.basic_ax_optimizer.BasicAxOptimizer' imported but unused
```
what is this part for?

**Comment by George3d6** [comment]
1.
I prefer not using start imports but it's not that big an issue and they do work in some situations.
In that case I think `Lang` should be imported explicitly and `torch` should be imported separately, as it stands it will be missing. However that code is currently "dead" so we never stumbled upon the issue.

2. I think that import was supposed to server to "expose" `BasicAxOptimizer` without using the full path to where it is. So I think it should stay

3. Again, same think, those should be there but there's a lot of dead code there that hasn't been used in half a year, or ever, so we never stumbled upon it. Feel free to ignore `rnn_helpers.py` and `rnn.py` as they will probably undergo partial-removal and refactoring whenever we have time.

4. General rule I use at the moment is class == camelCase, anything else == underscore_case. But some old code doesn't respect that rule, but again, rnn_helpers.py can safely be ignored for now. If this convention is not respected everywhere that's fine, I usually refactor it as I go along making other changes.

**Comment by djKooks** [comment]
@George3d6 thanks! I'll keep checking on, and let you know after done.

**Comment by djKooks** [comment]
@George3d6 hello,
I've update some more lint result, so hope you to check before going on because you'll have your own coding style, which can be important than PEP8 rule to you. 
So If you find some change you don't prefer, please let me know so I could ignore that rule.

Also, I've update some rule to ignore, which include 'star import' because seems you are thinking of refactoring which can effect this change. We can do it later.

This is the rule I've setup for ignore. If you have some rule prefer to be added, please let me know.
- F403: https://www.flake8rules.com/rules/F403.html 
- F405: https://www.flake8rules.com/rules/F405.html
- W504: https://www.flake8rules.com/rules/W504.html
- E722: https://stackoverflow.com/questions/51034424/pep8-does-not-allow-try-except-block/51034440

Thanks!

**Comment by George3d6** [comment]
Can you please tell me all of the commands you used to run the linter and check the reuslts ?

Other than that, I'll look over this when I have time and merge it together with the linter configuration.

**Comment by George3d6** [comment]
@djKooks 

Actually, two things:
* Please add back the `random` import in the autoencoder and other imports that are necessary, so that the tests don't fail
* PLease resolve the merge conflicts in rnn and datetime.

**Comment by djKooks** [comment]
@George3d6 thanks for comment!
You can test with:
```
$ pip install flake8
...
$ flake8 lightwood/**/*.py
...
```

George3d6:
Tests are passing, great, I'll try resolving the conflicts and merging this as soon as I have time. Might take a while since w'e are a bit busy and I want to look over every change just in case, but rest assured it will get merged.

Thank you for your help @djKooks

djKooks:
@George3d6 oh thanks! Because there are still few lints left, so if you're busy and need my help for solving conflict, let me know~

George3d6:
Sorry for the delay, fixed the conflicts, will merge tonight

PR #108, opened 2020-01-16
Initial CI tests

ZoranPandovski:
Initial structures for the CI tests. @George3d6  Note that full path is added to runpay as scripts so it can execute only __main__ logic. Also, there are few tests that are failing and are currently commented out.


George3d6:
@ZoranPandovski 
Please:
1. Bump the version
2. Add the `full_test` logic to `ci_tests.py`, we still want to run that along side the unit tests.

Other than that, this seems to be doing what we want, so all is well. I'll look into why those tests are failing and improve the logging as we go along, but this is a good start.

ZoranPandovski:
@George3d6 done.

Issue #110, opened 2020-01-16
Travis build not failing for failed tests


ritwik12:
**Describe the bug**
Even if there are failures in test scripts then also the travis build will pass
**To Reproduce**
Run travis build

**Expected behavior**
If there is an error in travis unit test scripts then the build should fail

**Additional context**
This is just the replication of similar issue in mindsdb/mindsdb  https://github.com/mindsdb/mindsdb/issues/343

ritwik12:
Good to close

Issue #109, opened 2020-01-16
Ignore deployment for Doc type change


ritwik12:
**Describe the bug**
Travis deploy is running for Docs also

**To Reproduce**
send any PR with a Doc change such as .md, LICENSE or .travis.yml

**Expected behavior**
Deployment should be skipped

**Additional context**
This is similar to what was requested in mindsdb/mindsdb, just a replication of https://github.com/mindsdb/mindsdb/issues/352

ritwik12:
Good to close

PR #111, opened 2020-01-16
Improvements to Travis

ritwik12:
* Ignore Doc change #109
* Fail travis build for errors in scripts (unit tests) #110

@George3d6 This is just the same replication from mindsdb/mindsdb and I think will be useful as I was going through Lightwood

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ install:
 before_script: cd tests/ci_tests
 # run tests
 script:
+  - set -e
   - if [ "$TRAVIS_OS_NAME" = "windows" ]; then travis_wait python full_test.py; fi
   - if [ "$TRAVIS_OS_NAME" != "windows" ]; then travis_wait python3 full_test.py; fi
   - cd ../..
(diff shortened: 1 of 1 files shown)


George3d6:
We need to figure out why the OSX tests is failing, a similar thing was happening to mindsdb. Could the `set -e` not work on OSX ?

The logs themselves say "timeout" but the tests didn't run for very long ( less than 4 minutes).

Any ideas @ritwik12 ?

ZoranPandovski:
The problem that we have found with -e also is that it exits immediately if a simple command exits with a non-zero status, and there are no additional logs from inside Travis.

ritwik12:
@ZoranPandovski  That'll be natural I guess. I suggest using pytest instead of running scripts.

ritwik12:
```
My build fails unexpectedly #

One possible cause for builds failing unexpectedly can be calling set -e (also known as set errexit), either directly in your .travis.yml, or sourceing a script which does. This causes any error causing a non-zero return status in your script to stop and fail the build immediately.

Note that using set -e in external scripts does not cause this problem, as the errexit is effective only in the external script.
```
For further details
https://docs.travis-ci.com/user/common-build-problems/

ritwik12:
@George3d6 I guess`set -e` breaks in Linux.
https://github.com/travis-ci/docs-travis-ci-com/issues/1672

ritwik12:
As suggested in above docs. We can use `set +e` later at some point based on the requirements of mindsdb

George3d6:
@ritwik12  I'll just remove the `set -e` for now and make sure to throw errors whenever our test scripts fail.

ritwik12:
@George3d6 Thanks, that will be great!!

PR #117, opened 2020-02-02
Weight balancing forst booster and accuracy function

George3d6:
* Added weight to gradient boosting helper mixer when predicting unbalanced categories
* Added weight to error function when applied to unbalanced categories
* Made boosting mixer training mandatory (maybe consider not doing this ? Or doing so only for datasets bellow a certain size ?), as it usually takes little time and there's a chance they perform when the network utterly fails to get us anything.

### DIFF
--- a/lightwood/config/config.py
+++ b/lightwood/config/config.py
@@ -25,6 +25,3 @@ class CONFIG:
 
     # Flags bellow are deprecated but still here in case we want to use the feature again
     OVERSAMPLE = False
-
-    """Bayesian Network"""
-    NUMBER_OF_PROBABILISTIC_MODELS = 2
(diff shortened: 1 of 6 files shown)


lightwood/config/config.py
```diff
@@ -12,17 +12,19 @@ class CONFIG:
...
-
     """Encoder options"""
     TRAIN_TO_PREDICT_TARGET = True
     MAX_ENCODER_TRAINING_TIME = 3600 * 2
+
+    # Flags bellow are deprecated but still here in case we want to use the feature again
+    OVERSAMPLE = False
+
+    """Bayesian Network"""
+    NUMBER_OF_PROBABILISTIC_MODELS = 2
```
torrmal: should we remove this in favor of USE_PROBABILISTIC_LINEAR

George3d6: Bayesian net is already removed, but I will remove the flag completely, sure.

I kept forgetting to look at improving accuracy with PLinear, I'll look into it once I'm done and/or give up on embedding layers + trainable encoders things (other PR, which is not ready)

PR #121, opened 2020-02-16
tsFresh time series encoder

surendra1472:
REF: https://github.com/blue-yonder/tsfresh/blob/master/notebooks/robot_failure_example.ipynb


George3d6:
So, few issues in order of importance:

## 1. Doesn't seem to behave similar to current timeseries format

Currently it seems like `test_data/train.csv` isn't the timeseries format we support

As in, the timeseries format we use is an array, so what the encoder would receive is something like:

`1 2 3 4 5 6 7 8` rather than 8 rows each with one numbers.

Though since ts_fresh supports encoding multiple columns we might instead pass an array o arrays (e.g. for the timeseries column and all the columns it depends upon)

Maybe I'm miss-understand what exactly is happening here though.

We could change this format... but that would be a whole other mess to handle in mindsdb and in lightwood, it's not as easy as changing a single encoder I don't think

## 2. The encoder won't work

The current interface is not supported for timeseries encoders, this will outright fail to run as it stands without modifying it to respect the encoder interface we currently have or modifying lightwood & mindsdb. Maybe the modifications are worth making though... I'm unsure on that front @torrmal maybe chip in if you have the time.

I'd be partial towards keeping things as we have them right now and adding features as we go, as opposed to re-vamping the way we do timeseries (especially until we have some good benchmarks for how different approaches to timeseries work).

## 3. Tests should be a bit more clear

I.e. why are `F_x,F_y,F_z,T_x,T_y,T_z` in the test file if they are not used anywhere ? Or are those values used somewhere ?

And there should be some assertions as to what exactly is happening. It might be best to just copy the test from the CesiumTS encoder for now and testing with that, since that will guarantee the two encoders are compatible.

BUT

Overall this scaffolding is more than enough, @surendra1472 so I think I might just pull the branch locally and do the modifications needed to integrate it on Monday. Unless you want to handle that yourself, in which case look at: https://github.com/mindsdb/lightwood/blob/master/lightwood/encoders/time_series/cesium_ts.py and use an example such as https://github.com/mindsdb/mindsdb-examples/tree/master/others/twitter for testing.

Let me know what you prefer, if you've got other things to work on don't worry, this should be easy enough for me to handle. If you want to finish this feature yourself and got the time for it then do tell and I'll leave you to it.

George3d6:
Ok, I think this is ready now, just gonna run it through some datasets to see if it's any good.

In the meanwhile @torrmal feel free to review it.

PR #127, opened 2020-02-26
Loss for multiple targets

George3d6:
## Changed the way we compute loss for targets

*Issue:* When we had more than one target, we use a single combined loss (MSELoss), this was especially bad if we had to predict different types (e.g. one categorical, one numerical) which could have encoding of different sizes

*Solution:* Have on loss per target and sum them together to get the overall loss.

## Selfaware changes

Basically, the selfaware network wasn't working, loss was increasing and then oscillating at very high value.

The change made is that instead of predicting the loss for each individual element, we predict the total loss (when call .item()) for every target. We use the inverse of this number as the confidence. Currently mindsdb does the normalization but I'll be switching that to lightwood in the next PR probably

In order to do this we apply the loss twice to the outputs, once with reduction (e.g. mean) so we can call .backwards on it, once without reduction so that the selfaware network is trained on each example individually, rather than on the overall accuracy for the batch.

This is how the loss progression for selfaware looks now, so it's presumably learning something:

newplot(2)

## Added confidence estimation based on the activation in the last layer

This is only implemented for categorical value, it requires storing some data during training to normalize the confidence.

The way we do it right now might not be perfect but we can optimize once we have benchmarks for it (see https://github.com/mindsdb/mindsdb/issues/400)

This is currently done with the `estimate_confidence` as part of the criterion object (in this case `TransformCrossEntropyLoss`), though I think that we could move it to a separate object. For now however I think it's easier to keep it there, since we already associate 1 criterion per output and the functionality is rather tiny + it's basically another way to evaluating the output, so it makes some sense for it to be part of the criterion, even though we're basically evaluating the outputs against themselves.

Note, all of these changes are backwards compatible, here's an example output on a problem predicting tow targets, one numerical one categorical:

```
{'location': {'predictions': ['great', 'good', 'poor', 'great', 'great', 'great', 'poor', 'good', 'great', 'great', 'good', 'poor', 'poor', 'good', 'poor'], 'selfaware_confidences': [1.236175616464377, 1.6827618081678242, 1.7715204173625982, 1.1813380147805825, 1.043548509390665, 1.390010890777802, 1.7798637064396776, 1.6411176202623818, 1.2842910952932183, 1.2639855179576753, 1.8544907290474373, 1.5884443557410584, 1.9437405743930276, 1.4949034313013976, 1.9257777593611143], 'loss_confidences': [0.9296639760705506, 1, 1, 1, 0.6482666342437551, 0.6895045756185776, 1, 1, 0.8183272935578144, 1, 1, 1, 1, 1, 1]}, 'rental_price': {'predictions': [0, 7043, 8114, 0, 25198, 5528, 0, 7889, 2839, 14854, 10872, 4722, 0, 8067, 0], 'selfaware_confidences': [0.8869535501792387, 0.9868010224907979, 0.9455396038309388, 0.8897142885628135, 0.8704289062724551, 0.90318393405211, 0.8641412523371225, 0.9846134199373522, 0.8962504977473028, 0.8556350110098413, 0.9666207013530211, 0.9430927287064579, 0.9092147459867603, 0.9678780765082856, 0.9389329288050975]}}
```

## Other things

* Should make it much easier to add a custom loss function for target types
* Simplified the label-encoding for cross entropy loss and moved it into a custom wrapper `TransformCrossEntropyLoss` (so it can be used similar to how we use MSE loss)
* Added a skeleton for Quantile loss ,though this is not used anywhere yet
* Removed oversampling in the DataSource (not needed now that we are properly computing weights in mindsdb and handling them in the CrossEntropyLoss)
* Added the ability to encode numerical targets as ranges to the Numeric encoder (disabled by default)


lightwood/api/predictor.py
```diff
@@ -454,6 +448,7 @@ def predict(self, when_data=None, when=None):
                         if 'confidences' in helper_mixer_predictions[output_column] and helper_mixer_predictions[output_column]['confidences'] is not None:
                             main_mixer_predictions[output_column]['confidences'] = list(helper_mixer_predictions[output_column]['confidences'])
 
+        print(main_mixer_predictions)
```
torrmal: prints

lightwood/api/data_source.py
```diff
@@ -60,6 +60,8 @@ def __init__(self, data_frame, configuration):
         self.enable_dropout = False
         self.disable_cache = not CONFIG.CACHE_ENCODED_DATA
         self.subsets = {}
+        self.out_indexes = None
+        self.out_types = None
```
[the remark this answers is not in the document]
George3d6: I can't get the output indexes (`out_indexes`) because those are only determined once we transform the data.

I can get the output type (`out_types`) out of the config... but, it's rather annoying to get to them and I'm relying on the ordering here (as in, the output indexes correspond to the types in order), and that's not decided in the config, that's decided in the transform. So I think it's better to source them this way. 

Overall, this part is a bit messy, but all other solutions were messy as well so I ended up going with this approach.

lightwood/api/predictor.py
George3d6: Will remove, will also remove a bunch of other redundant stuff in the numerical encoder and the quantile loss since now I have a separate branch for predicting numerical ranges.

PR #131, opened 2020-03-09
Example train script: Default of credit

surendra1472:
@George3d6 please verify the format. If it is fine,  ill add few more data sets. Thank you.


George3d6:
This is supposed to be an example, to that extent it should showcase as many features as possible with as little extra clutter as possible.

Thus I'd think we need to 

a) Simplify `calculate_test_score`

This:

```
def calculate_test_score():
    def convert_to_int(x):
        try:
            return int(x)
        except Exception:
            return 0

    test = pd.read_csv('dataset/test.csv')
    actual = list(test['default.payment.next.month'])

    test.drop(['default.payment.next.month'], axis=1)
    predictor = Predictor(load_from_path='test.pkl')
    predictions = predictor.predict(when_data=test)
    predicted = list(predictions['default.payment.next.month']['predictions'])
    predicted = list(map(convert_to_int, predicted))

    balanced_accuracy = balanced_accuracy_score(list(map(int, actual)), list(map(int, predicted)))
    print(f'Balacned accuracy score of {balanced_accuracy}')
    accuracy = accuracy_score(list(map(int, actual)), list(map(int, predicted)))
    print(f'accuracy score of {accuracy}')
```

Can be just this:

```
    test = pd.read_csv('dataset/test.csv')
    actual = [str(x) for x in test['default.payment.next.month']]
    
    predictor = Predictor(load_from_path='test.pkl')
    predictions = predictor.predict(when_data=test)
    predicted = [str(x) for x in predictions['default.payment.next.month']['predictions']]

    balanced_accuracy = balanced_accuracy_score(actual, predicted)
    print(f'Balacned accuracy score of {balanced_accuracy}')
```

b) To get proper results on this dataset weights should be used for the `1` and `0` outputs, see the interface of `learn` on how to do that, it's basically just a dictionary, same as that passed to mindsdb

c) Again, since it's an example I think we should KIS as much as possible, some remove any extra code that we don't need... remove the second lightwood import and use lightwood.Predictor instead, remove the `if __name__ == '__main__':` and maybe even just in-line all the logic or separate it into two functions with easy to understand names `train_model` and `test_model`.

surendra1472:
Addressed the pr comments a and c. I do not see weights related info in learn interface: https://github.com/mindsdb/lightwood/blob/master/lightwood/api/predictor.py#L143
@George3d6 please review, thanks.

George3d6:
Ok, my bad @surendra1472 , the weights are in the column config... forgot that.

Another thing we can probably do is provide links to the default on credit data (pandas should be able to read the csv from a link) as to no add this much code. I'll checkout locally and make those changes.

Thanks for the rest of the modification, everything else looks fine now :)

PR #134, opened 2020-03-16
Range/Quantile loss

George3d6:
Added a loss for predicting ranges for numbers, it works via lowering loss to zero if a prediction is within a given range from the target (currently set to +/-5%).

Added QuantileLoss and changed the numerical encoder to decode the quantiles and encode "copies" of the targets to predict the quantiles (yes, the copies are theoretically not needed, but in practice this is by far the easiest way to get this to work and doesn't cost us much).

The `confidence_range` now returned by lightwood is sourced from the 0.95 and 0.05 quantiles, the actual `predictions` vectors is sourced from the 0.5 quantile (which is basically the mean, so should be basically just an attempt at predicting the number exactly, which is what we were doing before).

Currently the 0.05 and 0.95 quantiles are hard-coded, in the future we could modify them dynamically.

`quantile_confidences` are also getting returned, this is simply the lower (0.95) quantile minus the upper (0.05) quantile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ docs/index.rst
 docs/make.bat
 docs/Makefile
 docs/source
+examples/lightwood_model.dill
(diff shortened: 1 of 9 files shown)


George3d6:

PR #141, opened 2020-03-25
Qunatile loss only

George3d6:
Only for comparative purposes, do not merge.


George3d6:

PR #151, opened 2020-04-09
Quantile loss for the neural network mixer (3rd time's a charm)

George3d6:


lightwood/api/predictor.py
```diff
@@ -65,11 +63,11 @@ def __init__(self, config=None, output=None, load_from_path=None):
         self.overall_certainty = None
 
     @staticmethod
-    def evaluate_mixer(mixer_class, mixer_params, from_data_ds, test_data_ds, dynamic_parameters,
+    def evaluate_mixer(config, mixer_class, mixer_params, from_data_ds, test_data_ds, dynamic_parameters,
```
torrmal, line 66: if we are passing the config, there is probably no need to pass the mixer class and mixer params part, right?

lightwood/api/predictor.py
```diff
@@ -362,7 +358,7 @@ def type_map(col_name):
 
                     # Once we are past the priming/warmup period, start training the selfaware network
 
-                    if subset_iteration > 1 and not mixer.is_selfaware and CONFIG.SELFAWARE and not mixer.stop_selfaware_training and training_error < 0.35:
+                    if subset_iteration > 1 and not mixer.is_selfaware and self.config['mixer']['selfaware'] and not mixer.stop_selfaware_training and training_error < 0.35:
```
torrmal, line 361: I think the idea is that the predictor API, does not know about self-aware at all, so we should abstract this a little better, we can talk about it in the call, we can make this change in a later PR

lightwood/data_schemas/predictor_config.py
```diff
@@ -20,11 +20,15 @@
 })
 
 mixer_schema = Schema({
-    'class': object,
+    Optional('class'): object,
     Optional('attrs'): dict,
-    Optional('mixer_graph'): [mixer_graph_schema]
+    Optional('deterministic', default=True): bool,
+    Optional('selfaware', default=True): bool
```
torrmal, line 26: should not this be in the attrs?

lightwood/data_schemas/predictor_config.py
```diff
@@ -20,11 +20,15 @@
 })
 
 mixer_schema = Schema({
-    'class': object,
+    Optional('class'): object,
     Optional('attrs'): dict,
-    Optional('mixer_graph'): [mixer_graph_schema]
+    Optional('deterministic', default=True): bool,
```
torrmal, line 25: should not this be in the attrs?

lightwood/data_schemas/predictor_config.py
```diff
@@ -33,6 +37,7 @@
     'output_features': [
         feature_schema
     ],
-    Optional('mixer'): mixer_schema,
+    Optional('data_source', default=data_source_schema.validate({})): data_source_schema,
```
torrmal, line 40: lets discuss this, not sure how this is in the config, but would like to understand better your reasoning here

lightwood/encoders/categorical/autoencoder.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, is_target=False):
             self.max_encoded_length = None
         else:
             self.max_encoded_length = 100
-        self.max_training_time = CONFIG.MAX_ENCODER_TRAINING_TIME
+        self.max_training_time = 7200
```
torrmal, line 33: this should be in the params for the encoder

lightwood/encoders/text/distilbert.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, is_target=False, aim=ENCODER_AIM.BALANCE):
         self._prepared = False
         self._model_type = None
         self.desired_error = 0.01
-        self.max_training_time = CONFIG.MAX_ENCODER_TRAINING_TIME
+        self.max_training_time = 7200
```
torrmal, line 27: encoder params

lightwood/mixers/boost/boost.py
```diff
@@ -10,7 +10,7 @@ def __init__(self, quantiles=None):
         self.targets = None
         self.quantiles = quantiles
 
-    def train(self, data_source):
+    def fit(self, ds=None, callback=None):
```
torrmal, line 13: need to add a method for iter_fit

lightwood/mixers/helpers/default_net.py
```diff
@@ -92,7 +53,7 @@ def __init__(self, ds, dynamic_parameters, shape=None, selfaware=False, size_par
 
             layers = []
             for ind in range(len(shape) - 1):
-                linear_function = PLinear  if CONFIG.USE_PROBABILISTIC_LINEAR else torch.nn.Linear
+                linear_function = PLinear if CONFIG.USE_PROBABILISTIC_LINEAR else torch.nn.Linear
```
torrmal, line 56: let's move the probabilistic linear to the params

lightwood/api/predictor.py
George3d6: There's no guarantee the mixer class will be in the config at the moment, and there could be mixer_params which we set by default but aren't in the config.

The mixer class gets assigned as: `mixer_class = NnMixer` by default and then we check to see if we check the config and override it if it's there.

Though I guess we could have a "default" mixer class in the config instead of setting the default in code, would be easier to understand for the user what the defaults are that way (and even for us).

lightwood/api/predictor.py
George3d6: I mean... I agree that the whole loop in Predictor.py was ugly... I wanted to get rid of it and do it similarly to how I do it in the trainable-encoders (using the gym.py implementation).

But it ends up being complicated so until we have a second "main" mixer I don't think it's worth bothering, or at least until we get time to refactor. So I agree, not ideal, but also agree that we should leave it for another PR.

lightwood/data_schemas/predictor_config.py
George3d6: Hmh, weren't the attributes mainly a placeholder ?

As in, if we know the value of some of these attributes, as long as we leave them as optional (aka some mixers might not have them) is it no better to make them explicit like this ?

lightwood/data_schemas/predictor_config.py
George3d6: Same as my reply to the  comment above

lightwood/encoders/categorical/autoencoder.py
George3d6: Hmh, I'll take a look and see if there's a way to add it as a default to the config... but I see no reason why it should be there more so than any of the other encoder attributes. I think we'll end up setting this dynamically based on the total allowed training time anyway.

lightwood/mixers/helpers/default_net.py
George3d6: I think that, as it stands, Plinear is still a "dev" feature only, so I'd rather leave it in the CONFIG (See my comment about wanting to leave dev flags in the CONFIG because it's easy for us to switch them around when we play with stuff)

George3d6: But if you'd prefer I can move it, I have no strong opinion either way.

lightwood/data_schemas/predictor_config.py
torrmal: cache_encoded_data flag?

PR #157, opened 2020-04-20
Fix tokenizers error

ZoranPandovski:
This PR changes the transformers to be fixed to the 2.4.1 version, to avoid the issue reported with installing tokenizers ``ERROR: Could not build wheels for tokenizers which use PEP 517 and cannot be installed directly``

### DIFF
--- a/lightwood/__about__.py
+++ b/lightwood/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'lightwood'
 __package_name__ = 'mindsdb'
-__version__ = '0.22.2'
+__version__ = '0.22.3'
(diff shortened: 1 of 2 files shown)


George3d6:
Hmh, couldn't we specify a stricter version for `tokenizers` instead of the `transformers` library as a whole?

This seems like it could potentially introduce a whole other host of issues since we are being so strict with the library version.

ZoranPandovski:
Maybe, what we can do is to use transformers >= 2.4.1, < 2.5.0. It would be better to let transformers to install whatever version of tokenizers they are depending on, and skip the older versions below 2.4.x.

George3d6:
Hmh, well, in that case, I'll merge it as is assuming the tests all pass.

PR #160, opened 2020-04-28
Improve CPU/GPU switching

### LINKED ISSUE #139: Training certain encoder might fail when the user has multiple GPUs
See: https://discuss.pytorch.org/t/is-there-a-way-to-set-a-custom-scatter-function-for-dataparallel/56751/3

It seems that DataParallel (part of the DefaultNet we use in a bunch of places, might not behave well when the mini-batches are python lists rather than tensors... which I believe we do in the case of the Categorical Auto Encoder (?).

It's annoying to test DataParallel, so not doing it right now, maybe I'd be wise to just disable it for the categorical autoencoder just in case ?

Alternatively, we could just wait and see if it's an issue, I'm not 100% sure it will be given the current implementation.

### LINKED ISSUE #155: Add .to(device) method to encoders
We need to add a function to the encoders that use pytroch to allow transferring them from GPU to CPU and vice versa.

George3d6:
* Revamped the tests as to take a shorter time to finish, mainly focused on making the autoencoder test stop at a lower accuracy, the `nn.py` test is actually more extensive in this iteration and the overall lightwood test is left to train for longer, so no meaningful thing should be lost by making them shorter, indeed, I think it's an overall improvement.

* Major improvements to switching from GPU to CPU and vice versa, abstracted it all away behind `.to` functions and added it to all encoders which might need it.

This approach should increase training times when migrating model from CPU to GPU, allow migration from GPU to CPU (when autoencoder, text encoder, image encoder or rnn encoder is used) and it also reduces code size.

* Also, made some improvements in the RNN Timeseries encoder, remove the need for transofrming into list and simplified the code a bit.

I attempted migrating the training to the `Gym` but... it's actually more complicated than I thought :/

Also, this should close #155 
Also, this should close #139


lightwood/encoders/categorical/autoencoder.py
```diff
@@ -43,6 +43,17 @@ def _encoder_targets(self, data):
...
         return labels
 
+    def to(self, device, available_devices):
+        if self.use_autoencoder:
+            self.net.to(device, available_devices)
+
+            modules = [module for module in self.net.modules() if type(
+                module) != torch.nn.Sequential and type(module) != DefaultNet]
+
+            self.encoder = torch.nn.Sequential(*modules[0:2])
```
torrmal, line 53: does this really belong in the to method? it seems like this should not be part of the method that defines what device to send it to, as its defining also the architecture of the encoder

lightwood/encoders/time_series/rnn.py
```diff
@@ -61,34 +59,32 @@ def prepare_encoder(self, priming_data, feedback_hoop_function = None):
...
+
             if average_loss < self._stop_on_error:
                 break
             if feedback_hoop_function is not None:
                 feedback_hoop_function("epoch [{epoch_n}/{total}] average_loss = {average_loss}".format(average_loss=average_loss, epoch_n=i+1, total=self._train_iters))
 
         self._prepared = True
 
-    def encode_one(self, data, initial_hidden = None, as_list = False, return_next_value = False):
+    def _encode_one(self, data, initial_hidden = None, return_next_value = False):
```
torrmal, line 79: why a hidden method?

lightwood/encoders/time_series/rnn.py
```diff
@@ -173,15 +160,14 @@ def encode(self, column_data, get_next_count = None):
...
-    encoder = RnnEncoder(encoded_vector_size=3,train_iters=500)
+    encoder = RnnEncoder(encoded_vector_size=3,train_iters=10)
     encoder.prepare_encoder(series, feedback_hoop_function=lambda x:print(x))
 
 
     # test de decoder
     init_vector = ['31 33 35 37', '1 2 3 4 5 6']
 
     print(encoder.encode(column_data=init_vector, get_next_count=2))
+    print(encoder.encode(column_data=init_vector))
```
torrmal, line 173: did you test running the encoder?

lightwood/mixers/helpers/default_net.py
```diff
@@ -111,6 +110,33 @@ def reset_layer_params(layer):
...
+        self.net = self.net.to(device)
+        if self.selfaware:
+            self.awareness_net = self.awareness_net.to(device)
+
+        available_devices = 1
+        if 'cuda' in str(device):
+            available_devices = torch.cuda.device_count()
+
+        if available_devices > 1:
+            self._foward_net = torch.nn.DataParallel(self.net)
```
torrmal, line 126: again, does this belong in the to method?

lightwood/mixers/nn/nn.py
```diff
@@ -777,19 +730,19 @@ def iter_fit(self, ds, initialize=True, subset_id=None):
...
     mixer = NnMixer({}, config)
 
     for i in mixer.iter_fit(ds):
-        if i < 0.01:
-            break
+        break
+    mixer.fit(ds,ds, stop_training_after_seconds=50)
 
+    predict_input_ds = DataSource(data_frame[['x', 'y']], config)
+    predict_input_ds.prepare_encoders()
```
torrmal, line 746: i dont get it, how are you going to prepare encoders after you fit the mixer?

lightwood/shared/helpers.py
```diff
@@ -0,0 +1,16 @@
+import torch
```
torrmal, line 1: i think the folder should be could lightwood/helpers/
rather than 'shared' and this file should be called device.py

lightwood/encoders/time_series/rnn.py
George3d6: because we only use it internally (in the encoder) and I'd rather keep the encoder's interface standard (e.g. only expose `to`, `prepare_encoder`, `encoder`, `decode` and the constructor).

Is there a situation where you see us using `encode_one` outside of the RNN ts encoder ?

lightwood/encoders/categorical/autoencoder.py
George3d6: Well, it's not defining the architecture per say.

As in `self.net` is the underlying network and then `self.encoder` and `self.decoder` are just Sequential wrappers over part of that.

But honestly, I think there *might* be a better way of doing this, in that I might be able to get rid of self.encoder and self.decoder all together, I'll take a look at doing this tomorrow.

lightwood/encoders/time_series/rnn.py
George3d6: Yes, did it not work for you locally?

lightwood/mixers/helpers/default_net.py
George3d6: Yes, it does since when we pass from one machine to another we might switch from a single GPU to multiple (and thus we have to use the `DataParallel` wrapper.

I don't really see a reason *not* to have this here. This is not the same case as with the autoencoder's `.to`.

lightwood/mixers/nn/nn.py
George3d6: That's just on the `predict_input_ds` (the one we use for predicting), the `ds` object (the one that's used when we call `.fit` is being prepared above).

lightwood/shared/helpers.py
George3d6: Hmh, yeah, I think that might be a more consistent naming schema, will change.

lightwood/encoders/categorical/autoencoder.py
torrmal: lets keep it very standard, network element definition in the init, network/graph stitching in the forward.

lightwood/encoders/time_series/rnn.py
torrmal: ok, I agree

lightwood/encoders/time_series/rnn.py
torrmal: you removed the get_next_count=2, why?

lightwood/mixers/nn/nn.py
torrmal: ok

lightwood/encoders/categorical/autoencoder.py
George3d6: The reason why I was re-defining `self.encoder` and `self.decoder` is that I thought they might essentially become pointers to some null memory or cause a copy of the layers they are pointing to once we applied the `.to` method to the network above, but apparently I miss-understood how `torch.nn.Sequential` works and this is not the case, they keep pointing to the newly GPU/CPU-transfered layers.

Removed all the extra stuff and now this method is much simpler :)

PR #181, opened 2020-06-14
fix name-not-defined error

### LINKED ISSUE #178: Error training helper mixers
Error: `Failed to train helper mixers with error: name 'data_source' is not defined`... investigate more later, happens when running the mindsdb_server tests.

maximlopin:
I failed to reproduce the error, but it looks like this will fix #178


George3d6:
Not sure how I missed this, anyway, was resolved a while ago so no longer needed but would have saved us some error solving if I saw this, next time @maximlopin if me or @torrmal  fail to review a PR for 1+ days please ping us again about it on slack.

PR #189, opened 2020-07-02
Datasource refactor and tests for ds cache 

btseytlin:
* Datasource slightly refactored
* Tests added for Datasource `encoded_cache` and `transformed_cache`


tests/plugin.py
```diff
@@ -0,0 +1,31 @@
+"""This file allows to override pytest behavior,
```
ELanning, line 1: small grammar nit, `allows to override`.

tests/plugin.py
```diff
@@ -0,0 +1,31 @@
...
+add new command-line options and test markers."""
+
+import pytest
+
+
+def pytest_addoption(parser):
```
ELanning, line 7: `addoption` is a convention you might not want to let bleed through to the rest of the project.

For example, if we were using a python library that used `camelCase` instead of `snake_case`, if we're going to wrap it, we may as well use `snake_case` in the wrappers name, so that the entire project is more consistently `snake_case`.

so here -> consider `pytest_addoption` -> `pytest_add_option`

lightwood/api/data_source.py
```diff
@@ -202,22 +199,35 @@ def __getitem__(self, idx):
...
 
+    def lookup_encoder_class(self, column_type):
+        path = 'lightwood.encoders.{type}'.format(type=column_type)
+        module = importlib.import_module(path)
+        if hasattr(module, 'default'):
+            encoder_class = importlib.import_module(path).default
+        else:
+            raise ValueError(
+                'No default encoder for {type}'.format(type=column_type))
+        return encoder_class
```
George3d6, line 219: I know this was not something your introduced, but, I  feel like adding the `ValueError` raising here is superfluous, since this edge case is not even one we could encounter at the moment and the default exception thrown should be sufficient for the user to figure out what's happening (and again, I would expect most people to run into this anyway).

If so we could replace this whole block with `return importlib.import_module(path).default`

But your call here, I'm fine with it either way.

lightwood/api/data_source.py
```diff
@@ -202,22 +199,35 @@ def __getitem__(self, idx):
...
-        """
         if column_name not in self.data_frame:
             nr_rows = self.data_frame.shape[0]
             return [None] * nr_rows
 
         return self.data_frame[column_name].tolist()
 
+    def lookup_encoder_class(self, column_type):
+        path = 'lightwood.encoders.{type}'.format(type=column_type)
+        module = importlib.import_module(path)
```
George3d6, line 213: This is declared here but bellow `importlib.import_module(path)` is used again, use this one bellow instead (unless you change the code as suggested, in which case we could remove this whole function actually and just replace it with:

`importlib.import_module(f'lightwood.encoders.{column_type}').default`

lightwood/api/data_source.py
```diff
@@ -229,72 +239,42 @@ def prepare_encoders(self):
...
-        self.out_types = []
-
-        for feature_set in ['output_features', 'input_features']:
-            for feature in self.configuration[feature_set]:
-                column_name = feature['name']
-                config = self.get_column_config(column_name)
+        self.out_types = [config['type'] for config in self.configuration['output_features']]
 
-                args = [self.get_column_original_data(column_name)]
+        for config in self.configuration['output_features']:
```
George3d6, line 244: Since you've removed most of the boilerplate and separate this into two loops you might as well make this a separate function that returns `input_encoder_training_data` + stets `self.out_type`, I think it would make the whole thing make even more sense. 

But just a stylistic thing, go with whatever you think is best.

lightwood/api/data_source.py
```diff
@@ -325,7 +305,7 @@ def get_encoded_column_data(self, column_name, custom_data=None):
             return encoded_vals
         else:
             raise Exception(
```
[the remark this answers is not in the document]
btseytlin: Good idea!

tests/plugin.py
btseytlin: I 100% agree, however I think it's a pytest specific hook name, so if I change it then it's not going to do its thing (add options on test run), since pytest won't be able to discover it

lightwood/api/data_source.py
```diff
@@ -118,36 +118,33 @@ def __len__(self):
...
-
-        :param idx:
-        :return:
-        """
         sample = {}
 
         dropout_features = None
 
-        if self.training == True and random.randint(0,3) == 1 and self.enable_dropout and CONFIG.ENABLE_DROPOUT:
+        if self.training is True and random.randint(0,3) == 1 and self.enable_dropout and CONFIG.ENABLE_DROPOUT:
```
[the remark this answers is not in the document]
btseytlin: Usually when I see something like `if self.training == True` I assume that the person writing this wanted to explicitly check for `True` and not, say, `1` or anything that evaluates to `True`.

So I was afraid to break something. I actually thought that using `is` is the right way, but, thanks, TIL its wrong.

I think I will just accept the risk of breaking something and simplify this to `if self.training`

lightwood/api/data_source.py
btseytlin: Agreed. Great, less code

lightwood/encoders/categorical/autoencoder.py
```diff
@@ -19,11 +19,12 @@ class CategoricalAutoEncoder:
     def __init__(self, is_target=False):
         self._pytorch_wrapper = torch.FloatTensor
         self._prepared = False
+        self.is_target = is_target
         self.name = 'Categorical Autoencoder'
         self.net = None
         self.encoder = None
         self.decoder = None
-        self.onehot_encoder = OneHotEncoder()
+        self.onehot_encoder = OneHotEncoder(is_target=is_target)
```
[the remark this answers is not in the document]
btseytlin: This is crazy, but absolutely right

Issue #183, opened 2020-06-24
Simple text encoder


George3d6:
Allow for a simple text encoder for small text (the selection of this encoder should happen in `mindsdb_native` under `backend/lightwood.py`) this will handle text that has very few words and it will essentially:

1. For each word create a one hot encoder (actually, potentially better to use the AutoEncoder, such that, if there's a lot of potential values for the text, the encoding doesn't explode)
2. Prepare all said OHEs
3. When `encode` is called have two behaviors (based on a flag in init):
   * Run all one hot encoders and concat the vectors (e.g. `[1 0 0 0]` and `[0 0 1]` becomes `[1 0 0 0 0 0 1]`)
   * Run all the one hot encoder and add all the vectors, adding elements on the same position and padding them to be equal to the value of the longest vector (e.g. `[1 0 0 0]` and `[0 0 1]` becomes `[1 0 1 0]`)
4. `decode` should work when the first mode of operation is selected but not when the second mode of operation is selected. IF this encoder is selected to output test values we should always default to the first mode of operations (since otherwise decode doesn't work and thus it can't be used to decode the outputs from the mixer)

Also, maybe the selection criteria should be a bit "fancier" (I.e. select it if the text is less than 5 words and if there's < 5000 unique values for each word, or something like that)

George3d6:
Done

PR #198, opened 2020-07-06
Small text encoder

maximlopin:
https://github.com/mindsdb/lightwood/issues/183
TextAutoEncoder + test


tests/unit_tests/test_short_text_encoder.py
```diff
@@ -0,0 +1,67 @@
...
+    if onehot:
+        assert not enc.use_autoencoder
+    else:
+        assert enc.use_autoencoder
+
+    encoded_data = enc.encode(test_data)
+
+    assert len(test_data) == len(encoded_data)
+
+    try:
```
btseytlin, line 60: Good test. You can use this helper to shorten this construct:
```
with pytest.raises(ValueError):
        <your code>
```

maximlopin:
`test_datasource.py` currently fails

lightwood/encoders/text/short.py
```diff
@@ -0,0 +1,109 @@
+import torch
+from torch.nn.functional import pad
+from lightwood.encoders.categorical import CategoricalAutoEncoder
+
+
+def _get_tokens(text):
```
George3d6, line 6: In the future let's use a special tokenizer, for now let's add splitting at ` `, `,` `#`

lightwood/encoders/text/short.py
```diff
@@ -0,0 +1,109 @@
...
+
+        if combine not in ['mean', 'concat']:
+            self._unexpected_combine()
+        
+        self._combine = combine
+
+        # Defined in self.prepare_encoder()
+        self._combine_fn = None
+    
+    def _unexpected_combine(self):
```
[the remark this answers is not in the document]
maximlopin: the same exception is raised twice so I wanted to avoid repeating the string so it can be changed in only one place in future

lightwood/encoders/text/short.py
maximlopin: I made a temporaty tokenizer that will split by various punctuation-like symbols. If future, I guess we will use nltk for tokenization?

lightwood/encoders/text/short.py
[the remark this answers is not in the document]
George3d6: Ah, didn't see that, fair enough :)

lightwood/encoders/text/short.py
George3d6: If nltk is already required, yes, if not we should try to use Flair.

What was the issue with the flair tokenizers again ? I didn't quite get that during the call

maximlopin: It has two tokenizers and both are weird (space tokenizer just cares about spaces, segtok tokenizer converts "hello,world!" to tok_1: "hello,world", tok_2: "!"

https://github.com/flairNLP/flair/blob/master/flair/data.py#L560

Issue #202, opened 2020-07-13
Create a base encoder for all encoders and make all encoders use the same interface


btseytlin:

maximlopin:
resolved by https://github.com/mindsdb/lightwood/pull/204

PR #204, opened 2020-07-13
base encoder

maximlopin:
Resolves https://github.com/mindsdb/lightwood/issues/202

### DIFF
--- a/lightwood/encoders/encoder_base.py
+++ b/lightwood/encoders/encoder_base.py
@@ -1,0 +1,19 @@
+
+class EncoderBase:
+    """Base class for all encoders"""
+    def __init__(self, is_target=False):
+        self.is_target = is_target
+
+    def prepare_encoder(self, priming_data):
+        """Can be overridden"""
+        pass
+
+    def encode(self, column_data):
+        raise NotImplementedError
+
+    def decoder(self, encoded_data):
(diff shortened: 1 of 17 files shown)


lightwood/encoders/encoder_base.py
```diff
@@ -0,0 +1,19 @@
+
+class EncoderBase:
+    """Base class for all encoders"""
+    def __init__(self, is_target=False):
+        self.is_target = is_target
+
+    def prepare_encoder(self, priming_data):
+        """Can be overridden"""
```
George3d6, line 8: I think the comments for "Can be overdeen" aren't necessary, we can just leave the `pass` there and that's indication enough.

George3d6: Also I think `prepare_encoder` should probably raise `NotImplementedError` , since all encoders should be prepared.

maximlopin: I think any encoder defines for itself whether it has to be prepared or not, so it's a virtual method.
.encode() and .decode() **must** be implemented because they're expected to return specific type of data (e.g. list of vectors in case of encode).

George3d6: Ah, yes, you're correct there.

PR #210, opened 2020-07-15
Multi-label prediction

btseytlin:
* Added `MultiHotEncoder`
* Added data type `MULTIPLE_CATEGORIES`. It's expected to be a list of categories
* Use BCEWithLogits loss for new data type
* Added lowercasing and expanding contractions when tokenizing text, for example, `Dont` -> `do not`
* Tests

Refactoring changes:
* short text encoder test moved to match project structure
* common encoder properties moved to BaseEncoder
* text tokenization helpers moved to `helpers.text`
* vector concat/mean helpers moved to `helpers.torch`
* tokenization now uses `nltk.word_tokenize`


lightwood/encoders/categorical/__init__.py
```diff
@@ -1,6 +1,3 @@
 from lightwood.encoders.categorical.onehot import OneHotEncoder
+from lightwood.encoders.categorical.multihot import MultiHotEncoder
 from lightwood.encoders.categorical.autoencoder import CategoricalAutoEncoder
-
-default = CategoricalAutoEncoder
-onehot = OneHotEncoder
-autoencoder = CategoricalAutoEncoder
```
btseytlin, line 6: Are these (except the `default` which is replaced by this PR) used anywhere, guys? Tell me if they are, then I will bring these back

lightwood/encoders/text/short.py
```diff
@@ -1,74 +1,16 @@
 import torch
 from torch.nn.functional import pad
-from lightwood.encoders.categorical import CategoricalAutoEncoder
-
-
-def _get_tokens(text):
```
btseytlin, line 6: Moved to helpers

lightwood/helpers/text.py
```diff
@@ -0,0 +1,27 @@
+import nltk
```
btseytlin, line 1: I checked this, nltk is a dependency of lightwood

btseytlin: Though its best to use this:
https://github.com/huggingface/tokenizers

And also apply it to the whole array of strings at once, not to individual phrases like now

lightwood/api/data_source.py
```diff
@@ -209,8 +213,18 @@ def get_column_original_data(self, column_name):
         return self.data_frame[column_name].tolist()
 
     def lookup_encoder_class(self, column_type):
-        encoder_class = importlib.import_module(f'lightwood.encoders.{column_type}').default
-        return encoder_class
+        default_encoder_classes = {
```
maximlopin, line 216: according to the guidelines that we've agreed on, I think it should be an `if/elif` statement rather than a `dict` lookup

lightwood/encoders/image/img_2_vec.py
```diff
@@ -5,17 +5,16 @@
...
+from lightwood.encoders.encoder_base import BaseEncoder
 
 
-class Img2VecEncoder(EncoderBase):
+class Img2VecEncoder(BaseEncoder):
 
     def __init__(self, is_target=False, aim=ENCODER_AIM.BALANCE):
+        super().__init__(is_target)
         self._model = None
         # I think we should make this an enum, something like: speed, balance, accuracy
```
maximlopin, line 16: already is an enum

lightwood/encoders/numeric/numeric.py
```diff
@@ -4,15 +4,15 @@
...
 
-class NumericEncoder(EncoderBase):
+class NumericEncoder(BaseEncoder):
 
     def __init__(self, data_type=None, is_target=False):
+        super().__init__(is_target)
         self._type = data_type
         self._mean = None
-        self._pytorch_wrapper = torch.FloatTensor
         self._prepared = False
```
maximlopin, line 16: `self._prepared `is already set in `BaseEncoder`

tests/unit_tests/test_multi_label_classification.py
```diff
@@ -0,0 +1,134 @@
...
+        train_tags = df_train.tags
+        predicted_tags = predictions['tags']['predictions']
+        train_tags_encoded = predictor._mixer.encoders['tags'].encode(train_tags)
+        pred_labels_encoded = predictor._mixer.encoders['tags'].encode(predicted_tags)
+        score = f1_score(train_tags_encoded, pred_labels_encoded, average='weighted')
+        print('Train f1 score', score)
+        self.assertGreaterEqual(score, 0.3)
+
+        # Why does it try to encode the missing column tags?
+        # predictions = predictor.predict(when_data=df_test.drop(['tags'], axis=1))
```
paxcema, line 124: Leftover?

lightwood/encoders/categorical/multihot.py
```diff
@@ -0,0 +1,25 @@
...
+import numpy as np
+from lightwood.encoders import BaseEncoder
+from sklearn.preprocessing import MultiLabelBinarizer
+
+
+class MultiHotEncoder(BaseEncoder):
+    def __init__(self, is_target=False):
+        super().__init__(is_target)
+        self.max_words_per_sent = None
+        self.binarizer = MultiLabelBinarizer()
```
George3d6, line 11: should be private I guess

lightwood/encoders/categorical/multihot.py
```diff
@@ -0,0 +1,25 @@
...
+        self.binarizer.fit(column_data)
+        self._prepared = True
+
+    def encode(self, column_data):
+        data_array = self.binarizer.transform(column_data)
+        return self._pytorch_wrapper(data_array)
+
+    def decode(self, vectors):
+        # It these are logits output by the neural network, we need to treshold them to binary vectors
+        vectors = np.where(vectors > 0, 1, 0)
```
George3d6, line 23: Just one question: does it not make more sense to set the thershold at e.g. 0.5, since the networks use ReLU up until the last layer => no negative numbers until then. But maybe the intuition I have here is just bad, I think it would be worth running the experiment in the future, but unless you can do it quickly and notice some improvements I'm 100% fine with keeping it this way for now.

btseytlin: I think our networks use SeLU. Either way the output vectors of our current nn mixer have negative numbers. Changing the threshold to 0.5 seems to provide no improvenment

Issue #206, opened 2020-07-14
Auto Encoder normalization


George3d6:
Currently out auto encoders (by which I mean the RNN autoencoder and the Categorical autoencoder) don't give us a guarantee that they output values within a certain range.

Ideally the autoencoders (and all encoders really) should output values within a fairly tiny range (say 90% of values between -2 and 2 and the rest with an abs value < 100... but these are just arbitrary numbers to illustrate the point).

This would ideally help the mixer train much faster, even though in theory neural nets can use unormalized inputs, in practice it seems that normalizing them helps (as opposed to e.g. having some in the 0-1 range and other in the 10^3-10^4 range)... but @paxcema @torrmal if you've got some citations for/against this point I'd be glad to hear them, since it might help us figure out what our normalization aim is.

At any rate, the alternative to normalizing in the autoencoders themselves is adding a normalization layer, I'm against that because:

1. log representations wouldn't be normalized properly if it's a linear norm (and vice versa, and generalized to a lot of functions which are incompatible with whatever normalization function we chose in terms of preserving the original meaning)
2. Depending on the actual function used there's no guarantee that the embedding layer's values can be normalized without loosing meaning (i.e. I don't think we have a guarantee the embeddings from auto encoders are linear if we'd decide to go for a linear norm)
3. Seems like something that would make connecting the encoding networks and training them together with the mixer even harder

***

Thus I think it's safe to say we want to normalize in the autoencoders themselves.

My proposed solution is to just have a "normalizing" activation function for the embedding layer. But @paxcema whenever you pick a specific solution for the RNN autoencoder you can leave the details here and we can implement it for the categorical auto encoder.

btseytlin:
A good idea might be to just use a small l1/l2 regularization when training these encoders (`weight_decay=1e6` argument for the optimizer). This does not guarantee a range, but the embeddings will usually be somewhere around -1 and 1

paxcema:
With respect to the time series encoder-decoder, the current embedding is the hidden state that the GRU encoder generates. As such, and given the GRU cell inner operations, we have a guarantee that each embedding component will be in the [-1, 1] range, which is what we probably want. Having said that, @btseytlin idea of weight decay is a good practice as well, so I've implemented it in the PR. 
I will check the categorical autoencoder and comment back on ideas for it.

George3d6:
Done.

Issue #169, opened 2020-05-13
Mutli-dimensional timeseries and decode support for RNN autoencoder


George3d6:
`Warning: This issue is intended as a test to be given to candidates for a job at Mindsdb, please do not start working on it, unless we've message you telling you to do so.`

## Current state

We presently have an RNN-based encoder for timeseries data, which will take a sequence of arbitrary length, e.g. `[1, 2, 3, 4, 5, 6, .... , 2000]` and turn it into an embedding of a fixed size (I believe it currently defaults to 4).'

It can be found here:

https://github.com/mindsdb/lightwood/blob/stable/lightwood/encoders/time_series/rnn.py

And the actual network is defined here:

https://github.com/mindsdb/lightwood/blob/stable/lightwood/encoders/time_series/helpers/rnn_helpers.py
 
(Note, there is an `EncoderRNNNumerical` and a `DecoderRNNNumerical`, but only the former is currently in use)

The unit tests for it can be found here:

https://github.com/mindsdb/lightwood/blob/stable/lightwood/encoders/time_series/rnn.py#L154

They can be executed by cloning lightwood, installing the requirements (`pip3 install -r requirements.txt`), adding lightwood to your python path (e.g. `export PYTHONPATH=$PYTHONPATH:/home/<your user>/lightwood`) and then cd-ing into the cloned directory and running `lightwood/encoders/time_series/rnn.py`.

## The task

You need to add a working `decode` function to the encoder, as well as extend the `encode` function to work with multi-dimensional timeseries data. You also need to write appropriate unit tests to validate all this functionality.

### Decoding

The interface for decode should be: `def decode(self, encoded_data)`.

Example usage & testing:
```
data = [[1,2,3,4,5,6],[2,4,8,16,32]]

encoder = RnnEncoder()
encoder.prepare_encoder(self, priming_data=data)
encoded_data = encoder.encode(data)
decoded_data = encoder.decode(encoded_data)

for i in range(len(data)):
    for j in range(len(data[i])):
        assert(decoded_data[i][j] == data[i][j])
```

### Encode Multi-Dimensional time series

A multidimensional time series cell will look something like this: `[(x11,x12... x1n), (x21, x22.... x2n), (xm1, xm2... xmn)]`.

So, for example, assume we have a time samples:

t1, t2, t3,... .tn at uneven spacing form each other (e.g. t1 might be 1000, t2 3000, t3 3200)

We could have the multidimensional time-series:

[...]

PR #212, opened 2020-07-16
RNN time series decoder

paxcema:
Based on submission for #169:
- Added decoder in ```RnnEncoder``` to reconstruct time series from its embedding
- Added multi-dimensional time series support
- Added batch training
- Simple overfitting unit test to confirm functionality. Additional benchmarking is needed.
- Extended ```tensor_from_series``` to pad over dimension axis and timestep axis for dynamic length handling.
- Added weight decay normalization, based on @btseytlin #206 suggestion.
- Changed optimizer to AdamW, considering #206 and this [paper](https://openreview.net/pdf?id=rk6qdGgCZ) as argument.

I would like feedback on the ```test_ts.py``` module, maybe it is better to simply add that logic in the ```__name__ == '__main__'``` block instead.


lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
+    :param series: list of lists, corresponds to time series: [[x1_1, ..., x1_n], [x2_1, ..., x2_n], ...]
+                   the series is zero-padded on each axis so that all dimensions have equal length
+    :param device: computing device that PyTorch backend uses
+    :param n_dims: if not None, will zero-pad dimensions until series_dimensions = n_dims
+    :param pad_value: value to pad each dimension in the time series, if needed
+    :param max_len: length to pad or truncate each time_series
+    :return: series as a tensor ready for model consumption, shape (1, ts_length, n_dims)
+    """
+    if isinstance(series[0], torch.Tensor):
+        return series[0]
```
George3d6, line 18: I feel like this is a decision chain that should be inlined where it happens, unless there's a situation where we call this method there we are unsure if the internal elements are tesnors or lists.

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
 import torch.nn as nn
-import torch.nn.functional as F
 
-import numpy as np
 
-
-def tensor_from_series(series, device):
-    if type(series) != type([]):
-        series = str(series).split(' ')
+def tensor_from_series(series, device, n_dims=None, pad_value=0.0, max_len=64):
```
George3d6, line 7: Are the optional arguments ever not set explicitly ? If not, they shouldn't be optional.

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
+                logging.warning(f'Weird element encountered in timeseries: {ele} !')
+                dimn_series.append(0)
+        float_series.append(dimn_series)
+
+    # timestep padding and truncating
+    timesteps = max([len(dimn) for dimn in float_series])
+    for i in range(len(float_series)):
+        # for _ in range(max(0, timesteps - len(float_series[i]))):
+        for _ in range(max(0, max_len - timesteps)):
+            float_series[i].append(pad_value)
```
George3d6, line 38: Those this padding algorithm work ?

It seem incorrect to me, since `max_len` and `timesteps` will always be the same, but float_series[0] and float_series[1] could be of different lengths.

Also, in the previous itteration of `for _ in range(max(0, timesteps - len(float_series[i])))` the `max` is still incorrect, since when the subtraction is < 0 we should be removing extra elements instead of padding.

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
+        float_series[i] = float_series[i][:max_len]
+
+    # dimension padding
+    if n_dims:
+        for _ in range(max(0, n_dims - len(float_series))):
+            float_series.append([pad_value] * timesteps)
+    else:
+        n_dims = len(float_series)
+
+    tensor = torch.tensor(float_series, dtype=torch.float, device=device).T.view(-1, max_len, n_dims).float()
```
George3d6, line 48: Could you use `transpose` explicitly here. I feel like `T` is kind of "arcane" in this situation, as in, I'm actually not sure what the resulting tensor would be. But, optional nitpick, if you think it makes more sense  or if it covers a case `transpose` wouldn'tit's up to you.

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
+    return tensor
 
 
 class DecoderRNNNumerical(nn.Module):
-    def __init__(self, hidden_size, output_size):
+    def __init__(self, hidden_size, output_size, dropout=0.2):
```
George3d6, line 53: Ideally, if dropout is hardcoded anyway, don't have it as an arg for both the "networks" of the encoder and for the encoder itself.

If you think it makes sense to have different dropouts for the two networks, that's fine, but then it should either be set internally in the main encoder (probably the best idea) or the main encoder should expose an interface like `dropout_decoder=0.2, dropout_decoder=0.2`.

George3d6: Alternatively, if you don't think this will be changed often anyway, there's no need to set it as an argument, just set the magic number `0.2` inside the `__init__`

[But, again, this is an optional nitpick]

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
 
     def initHidden(self, device):
         return torch.zeros(1, 1, self.hidden_size, device=device)
 
 
-
 class EncoderRNNNumerical(nn.Module):
-    def __init__(self,  hidden_size):
-        input_size = 1
+    def __init__(self,  input_size, hidden_size, dropout=0.2):
```
George3d6, line 73: Same comment of the `dropout` as with the `DecoderRNNNumerical`

lightwood/encoders/time_series/rnn.py
```diff
@@ -5,122 +5,171 @@
...
+import numpy as np
 import torch
 import torch.nn as nn
 from torch import optim
-import numpy as np
 
 
 class RnnEncoder(EncoderBase):
+    def __init__(self, encoded_vector_size=4, train_iters=75000, stop_on_error=0.8, learning_rate=0.01,
+                 is_target=False, ts_n_dims=1, dropout=0.2, weight_decay=1e-4, max_timesteps=64):
```
George3d6, line 16: The interface here is *probably* too complex, in that, it will probably never be used in full, it might make more sense to just set these hyperparameters internally and move them to the interface once an external module needs to actually modify them.

lightwood/encoders/time_series/rnn.py
```diff
@@ -149,26 +197,76 @@ def encode(self, column_data, get_next_count = None):
...
-
+        return self._pytorch_wrapper(ret)
 
-    # test de decoder
-    init_vector = ['31 33 35 37', '1 2 3 4 5 6']
 
-    print(encoder.encode(column_data=init_vector, get_next_count=2))
-    print(encoder.encode(column_data=init_vector))
+# only run the test if this file is called from debugger
+if __name__ == "__main__":
```
George3d6, line 234: The tests should be inside the `tests`  dir, specifically here: https://github.com/mindsdb/lightwood/tree/6c1d250e04d2978e09af854175d8fd8e4c78156a/tests/unit_tests

under a new dir `encoders/time_series`

... Sorry, new change, still not a thing for the other encoders, but we're trying to separate the tests from code atm, should be done for the old ones, but for now we can do is to start doing it for new code.

lightwood/encoders/time_series/rnn.py
```diff
@@ -149,26 +197,76 @@ def encode(self, column_data, get_next_count = None):
...
+    from lightwood.encoders.time_series.helpers.test_ts import test_ts
+
+    # set log, fix random seed for reproducibility
+    logging.basicConfig(level=logging.DEBUG)
+    seed = 42
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    if 'cuda' in str(get_devices()[0]):
+        torch.backends.cudnn.deterministic = True
+        torch.backends.cudnn.benchmark = False
```
George3d6, line 244: Hmh, I guess this code is needed here but it shouldn't be.

Currently this already happens in `default_net.py`, it should instead happen in `__init__.py` so that we don't have to manually do it in every test. Can you remove it from here and add it as a snipet in `__init__.py` (I guess keep the seeding the `default_net.py` as well, since otherwise we'd be having the same feed for each run of lightwood, and that might bias us in a weird direction (?) maybe not (?)

lightwood/encoders/time_series/rnn.py
```diff
@@ -149,26 +197,76 @@ def encode(self, column_data, get_next_count = None):
...
+        :return: decoded time series list
+        """
+        self._decoder.eval()
+        with torch.no_grad():
+            ret = []
+            next_tensor = torch.full((1, 1, self._n_dims), self._sos, dtype=torch.float32).to(self.device)
+            for _ in range(self._max_dec_length):
+                next_tensor, hidden = self._decoder.forward(next_tensor, hidden)
+                ret.append(next_tensor)
+            return torch.stack(ret)
```
George3d6, line 213: Not really a not on this PR specifically, but I only now realize this method exist and I think this is done manually in a few places by adding a dimension + `concat` , if you spot this anywhere else feel free to replace with `stack`, it makes much more sense.

lightwood/encoders/time_series/rnn.py
```diff
@@ -5,122 +5,171 @@
...
         self._train_iters = train_iters
         self._pytorch_wrapper = torch.FloatTensor
-        self._encoder = None
+        self._encoder = EncoderRNNNumerical(input_size=ts_n_dims, hidden_size=self._encoded_vector_size,
+                                            dropout=dropout).to(self.device)
+        self._decoder = DecoderRNNNumerical(output_size=ts_n_dims, hidden_size=self._encoded_vector_size,
+                                            dropout=dropout).to(self.device)
+        self.parameters = list(self._encoder.parameters()) + list(self._decoder.parameters())
+        self.optimizer = optim.AdamW(self.parameters, lr=self._learning_rate, weight_decay=weight_decay)
+        self.criterion = nn.MSELoss()
```
George3d6, line 30: Internal members, use `_` to declare unless I'm mistaken on them being internal only

lightwood/encoders/time_series/rnn.py
```diff
@@ -5,122 +5,171 @@
...
-
+            data_tensor = tensor_from_series(data, self.device, n_dims=self._n_dims,
+                                             pad_value=self._eos, max_len=self._max_ts_length)
+            steps = data_tensor.shape[1]
             encoder_hidden = self._encoder.initHidden(self.device)
             encoder_hidden = encoder_hidden if initial_hidden is None else initial_hidden
-            if type(encoder_hidden) is list:
-                encoder_hidden = torch.Tensor([[encoder_hidden]], device=device)
+
             next_tensor = None
```
George3d6, line 147: Unless `steps` can ever be == 0 this needn't be declared here.

lightwood/encoders/time_series/rnn.py
```diff
@@ -5,122 +5,171 @@
...
+            average_loss = average_loss / len(priming_data)
 
             if average_loss < self._stop_on_error:
                 break
             if feedback_hoop_function is not None:
-                feedback_hoop_function("epoch [{epoch_n}/{total}] average_loss = {average_loss}".format(average_loss=average_loss, epoch_n=i+1, total=self._train_iters))
+                feedback_hoop_function("epoch [{epoch_n}/{total}] average_loss = {average_loss}".format(
+                    epoch_n=i + 1,
+                    total=self._train_iters,
+                    average_loss=round(average_loss, 2)))
```
George3d6, line 125: Why the rounding here ? If it's just in order to alter the stopping condition here `if average_loss < self._stop_on_error:
` then it makes sense to have it as part of the stopping condition.

lightwood/encoders/time_series/rnn.py
```diff
@@ -5,122 +5,171 @@
...
                         next_tensor, encoder_hidden = self._encoder.forward(next_tensor.detach(), encoder_hidden)
-                    loss += criterion(next_tensor, data_tensor[tensor_i+1])
 
+                    loss += self.criterion(next_tensor, data_tensor[:, tensor_i + 1, :].unsqueeze(dim=1))
+
+                # decode
+                decoder_hidden = encoder_hidden
+                sos_vector = torch.full((batch.shape[0], 1, batch.shape[2]), self._sos,
+                                        dtype=torch.float32).to(self.device)
+                next_tensor = torch.clone(sos_vector)
```
George3d6, line 98: Does this need to be cloned here ? since `next_tensor.detach()` is called in the case where this representation is used and in the next iteration the variable will be reassigned anyway.

If it's just here because you might want to remove the `.detach` in the future and that would cause issues in case this tensor isn't cloned (would it ?) or if it's some other reason, feel free to leave it in.

George3d6: Actaully 99% this isn't needed, since one step above `torch.full` is used and that will create a new tensor anyway.

lightwood/encoders/time_series/rnn.py
```diff
@@ -5,122 +5,171 @@
...
+                    loss += self.criterion(next_tensor, tensor_target[:, tensor_i + 1, :].unsqueeze(dim=1))
 
-                loss = loss
                 average_loss += int(loss)
                 loss.backward()
-                optimizer.step()
+                self.optimizer.step()
 
-            average_loss = average_loss/len(priming_data)
+            average_loss = average_loss / len(priming_data)
```
George3d6, line 117: This should probably be `average_loss = average_loss / len(priming_data*batch_size)`

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
+                dimn_series.append(float(ele))
+            except Exception as _:
+                logging.warning(f'Weird element encountered in timeseries: {ele} !')
+                dimn_series.append(0)
+        float_series.append(dimn_series)
+
+    # timestep padding and truncating
+    timesteps = max([len(dimn) for dimn in float_series])
+    for i in range(len(float_series)):
+        # for _ in range(max(0, timesteps - len(float_series[i]))):
```
btseytlin, line 36: Probably leftover

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -2,58 +2,84 @@
...
-            float_series.append(0)
-
-    return torch.tensor(float_series, dtype=torch.float, device=device).view(-1, 1, 1, 1).float()
+    for dimn in series:
+        dimn_series = []
+        if isinstance(dimn[0], str):
+            dimn = [i for i in dimn[0].split(" ")]
+        for ele in dimn:
+            try:
+                dimn_series.append(float(ele))
```
btseytlin, line 27: Maybe we can offload the preprocessing work to mindsdb_native or some other part of lightwood, so that here we can assume that the received values are floats?

lightwood/encoders/time_series/helpers/test_ts.py
```diff
@@ -0,0 +1,59 @@
+import logging
+import torch
+
+
+def test_ts(encoder, queries, answers, params):
```
btseytlin, line 5: Please make it a unit test under tests/unit_tests, or maybe even multiple unit tests

George3d6:
@paxcema merge conflicts here are caused by the switch from `EncoderBase` to `BaseEncoder`, keep everything from your branch and just switch `EncoderBase` to `BaseEncoder` to resolve and I think everything should be fine.

lightwood/encoders/time_series/rnn.py
paxcema: I am not sure if I understand, can you elaborate?

lightwood/encoders/time_series/helpers/rnn_helpers.py
paxcema: Yes, this would be very useful, probably a separate PR.

lightwood/encoders/time_series/helpers/rnn_helpers.py
paxcema: Fixed.

Issue #218, opened 2020-07-20
Short text output


George3d6:
We now support unordered/bag-of-words style output from lightwood, however it might be useful to implement something like this fort ordered yet very short text as well.

Basically, allowing columns encoded via the short-text encoder (ShortText type) to be targets. The mixer would just be predicting ordered one-hot encoded representations in this case. With an all 0s vector indicating a stopping point.

Given that we have something like:

Cat = [0 0 1]
Dog = [0 1 0]
The = [1 0 0]

The mixer predicting `[1 0 0 0 1 0 0 0 0]` would be == `The Dog`
The mixer predicting `[0 0 1 0 0 0 0 1 0]` would be == `Cat`
The mixer predicting `[1 0 0 0 0 1 0 1 0]` would be == `The Cat Dog`

The current sort text encoder already supports encoding/decoding these representations in `concat` mode, we just need to make sure `concat` mode is always selected when `target == True` in the encoder's args.

The other problem here is finding a dataset to test this on and finding an appropriate loss function.

I think @btseytlin you had some ideas as to the ideal loss to be used here and maybe some of the datasets you provided could work as well, if so, please do tell.

maximlopin:
my thoughts on loss function:
If the target is `The cat dog -> [1, 0, 0, 0, 1, 0, 0, 0, 1]`, and the prediction is `The Dog Cat -> [1, 0, 0, 0, 0, 1, 0, 1, 0]`,
the loss would be `CrossEntropy([1, 0, 0], [1, 0, 0]) + CrossEntropy([0, 1, 0], [0, 0, 1]) + CrossEntropy([0, 0, 1], [0, 1, 0])`.
Loss function should ignore vectors following the "stopping point" vector.

maximlopin:
partially resolved by https://github.com/mindsdb/lightwood/pull/219
now need to find a good dataset

PR #219, opened 2020-07-20
use concat mode if is_target

maximlopin:
related to https://github.com/mindsdb/lightwood/issues/218
- fix typo: `BaseEncoder.decoder()` in renamed to `BaseEncoder.decode()`
- `ShortTextEncoder` uses `"concat"` mode if `is_target is True` and tests for this are added too

### DIFF
--- a/tests/unit_tests/encoders/text/test_short_text_encoder.py
+++ b/tests/unit_tests/encoders/text/test_short_text_encoder.py
@@ -48,9 +48,9 @@ class TestShortTextEncoder(unittest.TestCase):
         assert enc.is_target == is_target
 
         if is_target:
-            enc._combine == 'concat'
+            assert enc._combine == 'concat'
         else:
-            enc._combine == 'mean'
+            assert enc._combine == 'mean'
 
         encoded_data = enc.encode(test_data)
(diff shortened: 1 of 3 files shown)


tests/unit_tests/encoders/text/test_short_text_encoder.py
```diff
@@ -20,27 +20,38 @@ def test_get_tokens(self):
...
         vocab_size = 99 if onehot else 800
         
         priming_data = generate_sentences(2, 6, vocab_size)
         test_data = random.sample(priming_data, len(priming_data) // 5)
 
-        enc = ShortTextEncoder(combine='concat')
+        enc = ShortTextEncoder(is_target=is_target)
         enc.prepare_encoder(priming_data)
 
         if onehot:
```
btseytlin, line 43: I think at this point, with 4 possible states for this function, it might be better to make separate tests. Yes, this will make the code of tests repetitive, but it will be much more readable. It will also make it easier to debug when a test fails: if there is an exception, you will not only get the info that `test_mean` failed, but you will get the info that `test_info_onehot_not_target` failed.

PR #221, opened 2020-07-20
Release 0.31

George3d6:
* Encoder refactor to respect new interface
* A bit of cleaning up of the DataSource
* Multicategorical output support
* Multicategorical encoder
* Fix for the way loss functions are weighted
* Short text encoder
* Moving some tests to unittest
* A few other minor non critical fixe/imporvements here and there


tests/unit_tests/encoders/time_series/test_ts.py
```diff
@@ -0,0 +1,99 @@
+import logging
+import torch
+from lightwood.helpers.device import get_devices
+from lightwood.encoders.time_series import RnnEncoder
+from lightwood.encoders.time_series.helpers.rnn_helpers import tensor_from_series
+
+
+def test_ts(encoder, queries, answers, params):
```
btseytlin, line 8: This should be a unit test

George3d6:
@btseytlin there's a few other changes needed to the RNN for the test PR to work, we're going to implement them there, but for now I will release this as-is in order to be within schedule. We can fix them in the next release, for now the RNN Encoder is not used as the default, so everything should be fine (even if it is, it will still work, the tests only revelaed there's a potential issue for timeseries output, for input it's alright).

PR #225, opened 2020-07-21
Use F1 score to evaluate multiple categories score

btseytlin:
To be merged after: multihot-handle-none


George3d6:
Ideally we'd want to move all scores to % accuracy scores, easier for users to read and for the selfaware network to interpret to give a probability of the prediction being true (instead of predicting the loss). For numbers we can do this by using the quanitles and thus just predicting % chance that a number is within the quantiles.

Do you think there's any way of implementing an accuracy score for this problem that goes from 0 to 1 ? Even if it's not "perfect".

Issue #217, opened 2020-07-19
Separate selfaware ?


George3d6:
The gist of this pondering boils down to these 2 things:

a) Currently we can't start training the selfaware when we start training the model, otherwise the model never converges unless our lr is unreasonably slow (think 1e-6 kinda values)

b) Currently the loss from the selfaware seems to often be much greater than that of the actual model and the whole setup makes training less stable (gradients go all over the place in different epochs instead of constantly going down)

c) Currently the selfaware network is part of the `DefaultNet` making that whole setup much more complicated... but I see no reason why it ought to be

d) Currently when the selfaware kicks in we have to lower the main networks learning rate, thus making it slower.

There seems to be an obvious solution to all of this problem:

Make the selfaware a separate network, with it's own optimizer.

This means that we can start training it right away and it won't hamper training the main model in the beginning. This means the selfaware can be trained for longer

It also means that we can just adjust the selfaware's optimizer (or weight down it's loss, though that seems like a much worst idea) to a lower learning rate and keep the main networks optimizer to the highest lr possible.

The code basically becomes something like:

```
Yh = net(X)
loss = criterion(Yh,Y)
loss.backwards()
opt.step()

awareness_in = torch.cat(Yh,X)
if not train_together:
   awareness_in = awareness_in.detach()
awareness = selfaware(awareness_in)
# Remember, in the current setup the selfaware is trying to predict the unreduced loss of the main model's predictions
unreduced_loss = criterion(Yh,Y)
selfaware_loss = selfaware_criterion(awareness, unreduced_loss)
selfaware_loss.backwards()
selfaware_opt.step()
```

Suddenly enabling the selfware and disabling it also becomes much simple, no need to reinitialize the whole main network, just a single if is needed to stop backpropagatin it's results through the main model and another one to stop it from making predictions and/or being trained.

paxcema:
Implemented in #228.

PR #228, opened 2020-07-24
[WIP] Separate Self-Aware

paxcema:
Very early implementation that separates the self-aware network from the main NN into its own module to have a training loop where each one has its separate optimizer. See #217. Testing and benchmarking is still pending.

### DIFF
--- a/lightwood/mixers/helpers/selfaware.py
+++ b/lightwood/mixers/helpers/selfaware.py
@@ -1,0 +1,66 @@
+import torch
+import math
+
+from lightwood.config.config import CONFIG
+from lightwood.mixers.helpers.ranger import Ranger
+
+
+# Todo: conform to style guidelines
+class SelfAware(torch.nn.Module):
+    def __init__(self, input_size, output_size, nr_outputs):
+        super(SelfAware, self).__init__()
+
+        self.available_devices = None
+        self.device = None
+
+        self.input_size = input_size
+        self.output_size = output_size
+        self.nr_outputs = nr_outputs
+
+        awareness_layers = []
+        awareness_net_shape = [(self.input_size + self.output_size),
+                               max([int((self.input_size + self.output_size) * 1.5), 300]),
+                               self.nr_outputs]
+
+        for ind in range(len(awareness_net_shape) - 1):
+            rectifier = torch.nn.SELU
+            awareness_layers.append(torch.nn.Linear(awareness_net_shape[ind], awareness_net_shape[ind + 1]))
+            if ind < len(awareness_net_shape) - 2:
+                awareness_layers.append(rectifier())
+
+        self.awareness_net = torch.nn.Sequential(*awareness_layers)
+
+        for layer in self.awareness_net:
+            if hasattr(layer, 'weight'):
+                torch.nn.init.normal_(layer.weight, mean=0., std=1 / math.sqrt(layer.out_features))
+            if hasattr(layer, 'bias'):
+                torch.nn.init.normal_(layer.bias, mean=0., std=0.1)
+
+        self.opt = Ranger(self.awareness_net.parameters())
+
+    def to(self, device, available_devices):
+        self.awareness_net = self.awareness_net.to(device)
+
+        if available_devices > 1:
+                self.awareness_net = torch.nn.DataParallel(self.awareness_net)
+
+        self.device = device
+        self.available_devices = available_devices
+
+        return self
+
+    def forward(self, input):
+        """
+        :param input: tensor with (true_input, main_net_output) pairs to estimate loss magnitude
+        :return: predicted loss value over the tensor samples
+        """
+        output = self.awareness_net(input)
+        return output
+
+
+
+if __name__ == '__main__':
+    print("Various tests that need to be aggregated in tests/ before PRing")
+    sa = SelfAware(100, 50, 1)
+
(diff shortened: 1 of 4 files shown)


lightwood/mixers/helpers/selfaware.py
```diff
@@ -0,0 +1,61 @@
...
+
+        for ind in range(len(awareness_net_shape) - 1):
+            rectifier = torch.nn.SELU
+            awareness_layers.append(torch.nn.Linear(awareness_net_shape[ind], awareness_net_shape[ind + 1]))
+            if ind < len(awareness_net_shape) - 2:
+                awareness_layers.append(rectifier())
+
+        self.net = torch.nn.Sequential(*awareness_layers)
+
+        for layer in self.net:
```
paxcema, line 30: Move to a helper that is shared

lightwood/mixers/nn/nn.py
```diff
@@ -83,7 +85,9 @@ def build_confidence_normalization_data(self, ds, subset_id=None):
 
             with torch.no_grad():
                 if self.is_selfaware:
-                    outputs, awareness = self.net(inputs)
+                    outputs = self.net(inputs)
+                    aware_in = torch.cat((inputs, outputs), 1)
```
paxcema, line 89: Add cat into selfaware forward

lightwood/mixers/nn/nn.py
```diff
@@ -320,7 +324,9 @@ def predict(self, when_data_source, include_extra_data=False):
 
             with torch.no_grad():
                 if self.is_selfaware:
-                    output, awareness = self.net(inputs)
+                    output = self.net(inputs)
```
paxcema, line 327: set eval mode

lightwood/mixers/nn/nn.py
paxcema: add eval

lightwood/mixers/nn/nn.py
```diff
@@ -619,13 +621,16 @@ def iter_fit(self, ds, initialize=True, subset_id=None):
...
+                self.selfaware_optimizer.zero_grad()
 
                 # forward + backward + optimize
-                # outputs = self.net(inputs)
+                outputs = self.net(inputs)
                 if self.is_selfaware:
-                    outputs, awareness = self.net(inputs)
-                else:
-                    outputs = self.net(inputs)
+                    if self.start_selfaware_training:
```
paxcema, line 629: Flags: selfaware_training_standalone, _with_net.

lightwood/mixers/nn/nn.py
```diff
@@ -658,14 +662,11 @@ def iter_fit(self, ds, initialize=True, subset_id=None):
...
 
-
                 if CONFIG.MONITORING['batch_loss']:
                     self.monitor.plot_loss(loss.item(), self.total_iterations, 'Targets Batch Loss')
 
-
-
                 if awareness_loss is not None:
-                    awareness_loss.backward(retain_graph=True)
+                    awareness_loss.backward()
```
paxcema, line 669: if training_encoder: retain_graph

lightwood/mixers/helpers/selfaware.py
George3d6: Ideally keep them separate actually, since we might want to differentiate between the deafultNet and the selfaware net, we can move boilerplate to helper functions as we go.

lightwood/mixers/nn/nn.py
```diff
@@ -320,7 +325,9 @@ def predict(self, when_data_source, include_extra_data=False):
 
             with torch.no_grad():
                 if self.is_selfaware:
-                    output, awareness = self.net(inputs)
+                    self.selfaware_net.eval()
```
George3d6, line 328: it seems like the main network is not set to `eval`, if so we should also do this here. If it is set, but somewhere else, we should set the selfaware to eval there as well.

lightwood/mixers/nn/nn.py
```diff
@@ -83,7 +86,9 @@ def build_confidence_normalization_data(self, ds, subset_id=None):
 
             with torch.no_grad():
                 if self.is_selfaware:
-                    outputs, awareness = self.net(inputs)
+                    self.selfaware_net.eval()
```
George3d6, line 89: it seems like the main network is not set to `eval`, if so we should also do this here. If it is set, but somewhere else, we should set the selfaware to eval there as well.

lightwood/mixers/helpers/selfaware.py
```diff
@@ -0,0 +1,63 @@
...
+        output = self.net(aware_in)
+        return output
+
+
+
+if __name__ == '__main__':
+    print("Various tests that need to be aggregated in tests/ before PRing")
+    sa = SelfAware(100, 50, 1)
+
+    print(sa)
```
George3d6, line 63: Remove or move under the current test structure.

lightwood/mixers/nn/nn.py
paxcema: It is set at the start of this function, so as it stands it seems fine to call .eval() for the self-aware here.

lightwood/mixers/nn/nn.py
paxcema: Same as in the previous comment

PR #233, opened 2020-08-01
use list.extend() and list.append() instead of list.__add__()

maximlopin:
`list.__add__(a, b)` is `O(len(a) + len(b))`
`list.extend(a, b)` is `O(len(b))`
`list.append()` is `O(1)`

using `list.__add__` might slow down lightwood a lot, especially when there is a for loop like this:

```
ret = []
for x in L:
    ret += [x]
```
because this is `O(n^2)`


maximlopin:
OSX CI fails: `The command pip3 install --no-cache-dir --no-use-pep517 -e . exited with 1.`

PR #234, opened 2020-08-03
Mock rich text output

George3d6:
I call this a "mock" rich text output since it doesn't really generate coherent text outputs, it's basically just a work tokenizer head using the hugging-face pretrained BERT tokenizer. In the future we should:

* Figure out how to `train` the tokenizer on our training body of text but still keep the pretrained vocabulrary.
* Figure out how to mix the mixers output with a search algorithm that infers the probability of the next word based on previous word[s] (or some such simplistic mechanism to get coherent sentences).

As it stands it *might* be better than having nothing for rich text and we might be able to build on this, at least that's part of my thinking, though it might make more sense to disable rich text output for now until there's a usecase for it. Since this won't do and any good implementation seems to require way to much work that @paxcema is more qualified to do then myself.


tests/unit_tests/encoders/text/test_vocab.py
```diff
@@ -0,0 +1,26 @@
+import unittest
+
+from lightwood.encoders.text.vocab import VocabularyEncoder
+
+
+class TestVocabularyEncoder(unittest.TestCase):
+    def test_encode_decode(self):
+        self.maxDiff = None
```
paxcema, line 8: Currently useless

George3d6: useful since the text diffs can be large, and `maxDiff` set to the magic value allows us to see them.

George3d6:
OSX is failing because of travis, the rest are passing, so will go ahead and merge

PR #236, opened 2020-08-04
Release 0.33

George3d6:


lightwood/encoders/text/vocab.py
```diff
@@ -0,0 +1,34 @@
...
+
+from lightwood.encoders.encoder_base import BaseEncoder
+
+
+class VocabularyEncoder(BaseEncoder):
+    def __init__(self, is_target=False):
+        super().__init__(is_target)
+        self._tokenizer_class = DistilBertTokenizer
+        self._pretrained_model_name = 'distilbert-base-uncased'
+        self._max_len = None
```
maximlopin, line 13: To avoid searching for member variables definitions they can be initialized to None in `__init__()`:
```
# These are initialized in self.prepare_encoder()
self._max_len = None
self._tokenizer = None
self._pad_id = None
```

lightwood/encoders/text/vocab.py
```diff
@@ -0,0 +1,34 @@
...
+    def prepare_encoder(self, priming_data):
+        self._max_len = max([len(x) for x in priming_data])
+        self._tokenizer = self._tokenizer_class.from_pretrained(self._pretrained_model_name)
+        self._pad_id = self._tokenizer.convert_tokens_to_ids([self._tokenizer.pad_token])[0]
+
+    def encode(self, column_data):
+        vec = []
+        for text in column_data:
+            encoded = self._tokenizer.encode(text[:self._max_len], add_special_tokens=True)
+            encoded = torch.tensor(encoded + [self._pad_id] * (self._max_len - len(encoded)))
```
maximlopin, line 24: I guess `len(encoded)` can be greater than `self._max_len` if special tokens are added

George3d6: True, but the resulting vectors would not be of the same size, which would be an issue. We always want the encoded representation from our encoders to have the same size.

lightwood/encoders/text/vocab.py
George3d6: Solved

PR #246, opened 2020-08-14
Release 0.36

paxcema:
- Separated self-aware net from default net, simplified mixer logic
- CI for macOS is now only done for Python 3.7


maximlopin:
we might want a test for self aware network

PR #247, opened 2020-08-17
Time series RNN optimizations

paxcema:
- Added time series maximum length detection, replacing default value
- Increased default batch_size

### DIFF
--- a/lightwood/encoders/time_series/rnn.py
+++ b/lightwood/encoders/time_series/rnn.py
@@ -12,7 +12,7 @@ from torch import optim
 
 class RnnEncoder(BaseEncoder):
 
-    def __init__(self, encoded_vector_size=4, train_iters=100, stop_on_error=0.8, learning_rate=0.01,
+    def __init__(self, encoded_vector_size=4, train_iters=100, stop_on_error=0.01, learning_rate=0.01,
                  is_target=False, ts_n_dims=1):
         super().__init__(is_target)
         self.device, _ = get_devices()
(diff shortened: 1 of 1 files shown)


lightwood/encoders/time_series/rnn.py
```diff
@@ -12,14 +12,14 @@
 
 class RnnEncoder(BaseEncoder):
 
-    def __init__(self, encoded_vector_size=4, train_iters=75000, stop_on_error=0.8, learning_rate=0.01,
-                 is_target=False, ts_n_dims=1, max_timesteps=64):
+    def __init__(self, encoded_vector_size=4, train_iters=100, stop_on_error=0.8, learning_rate=0.01,
```
George3d6, line 15: Unrelated to the PR, but, is `stop_on_error` supposed to be this high ? Shouldn't it be closer to ~ 0.08 or thereabouts ?

Ideally we'd want to add "stop when not improving" logic to this encoder anyway (see Gym for how that's done). But for now a quick fix might involve lowering this number together with lowering `train_iters`.

lightwood/encoders/time_series/rnn.py
```diff
@@ -48,6 +48,15 @@ def prepare_encoder(self, priming_data, feedback_hoop_function=None, batch_size=
         if self._prepared:
             raise Exception('You can only call "prepare_encoder" once for a given encoder.')
 
+        # determine time_series length
+        for str_row in priming_data:
+            l = len(str_row.split(" "))
+            self._max_ts_length = l if l > self._max_ts_length else self._max_ts_length
```
maximlopin, line 54: `max(l, self._max_ts_length)`

lightwood/encoders/time_series/rnn.py
paxcema: You are right, changing this to 0.01 as a quick fix.

Issue #255, opened 2020-08-25
Make helper mixers normal mixers


George3d6:
Remove the helper mixer logic from lightwood, instead make the helper mixers (only one of them exists, the boosting mixer) behave like a normal mixer and allow for them to be selected in place of the NN mixer from the config passed to lightwood.

Only implement this into stable after is https://github.com/mindsdb/mindsdb_native/issues/131 merged

PR #259, opened 2020-09-03
Make common interface for lightwood mixers

maximlopin:
Resolves https://github.com/mindsdb/lightwood/issues/255

- Move all `NnMixer` logic from `Predictor` class to `NnMixer` class
- Fix `NnMixer` test
- Remove `'attrs'` from config and add `'kwargs'` instead and pass it to encoder's constructor
- Make `BaseMixer` class
- Move encoder preparation logic to `DataSource`
- Add test for `BoostMixer`
- Allow passing `DataSource` to `Predictor.learn`

### DIFF
--- a/tests/unit_tests/api/test_datasource.py
+++ b/tests/unit_tests/api/test_datasource.py
@@ -34,7 +34,9 @@ class TestDataSource(TestCase):
                     'type': 'categorical',
 
                 }
-            ]
+            ],
+
+            'data_source':{'cache_transformed_data': True}
         }
         config = predictor_config_schema.validate(config)
         n_points = 100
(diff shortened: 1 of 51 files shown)


maximlopin:
Currently there is a very strange test that fails when you run all unit tests (`py -m unittest /unit_tests`), but it passes when you run it standalone (`py -m unittest intest_multiple_categories_as_output.py`).
```
FAIL: test_multiple_categories_as_output (test_multi_label_classification.TestMultiLabelPrediction)
----------------------------------------------------------------------
  File test_multi_label_classification.py", line 122, intest_multiple_categories_as_output
    self.assertGreaterEqual(score, 0.15)
AssertionError: 0.08328666909070302 not greater than or equal to 0.15
```

maximlopin:
BoostMixer doesn't work and its test is commented out, I'll be working on it in a separate PR

George3d6 (approved):

tests/unit_tests/api/test_datasource.py
```diff
@@ -49,11 +51,10 @@ def setUp(self):
         self.config = config
         self.df = df
 
-    def test_prepare_encoders(self):
+    def testprepare_encoders(self):
```
paxcema, line 54: is this a typo?

paxcema (approved):

PR #260, opened 2020-09-04
Hotfix for DistilBert call to DefaultNet

paxcema:
Should fix [#671](https://github.com/mindsdb/mindsdb/issues/671)

- Removes `selfaware` argument passing in call to DefaultNet
- Restores DistilBert unit test to catch issues like this next time (for now, without asserting an accuracy threshold)


George3d6 (approved):
Good catch, please merge right away :+1:

maximlopin (approved):

PR #266, opened 2020-09-15
Quicker ts-string to array transformation for the RNN encoder

George3d6:
The RNN encoder transforms strings like `[1 2 3 4]` into `[1, 2, 3, 4]` sooner in the encoder and priming functions.

This also means it can be more easily removed once we make mindsdb_native only pass arrays to lightwood for timeseries columns.

### DIFF
--- a/lightwood/encoders/time_series/rnn.py
+++ b/lightwood/encoders/time_series/rnn.py
@@ -171,6 +171,11 @@ class RnnEncoder(BaseEncoder):
         if not self._prepared:
             raise Exception('You need to call "prepare_encoder" before calling "encode" or "decode".')
 
+        for i in len(column_data):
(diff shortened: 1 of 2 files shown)


paxcema (approved):

lightwood/encoders/time_series/rnn.py
```diff
@@ -48,10 +48,12 @@ def prepare_encoder(self, priming_data, feedback_hoop_function=None, batch_size=
         if self._prepared:
             raise Exception('You can only call "prepare_encoder" once for a given encoder.')
 
-        # determine time_series length
-        for str_row in priming_data:
-            l = len(str_row.split(" "))
-            self._max_ts_length = max(l, self._max_ts_length)
+        # Convert to array and determine max length:
+        for i in len(priming_data):
```
paxcema, line 52: `range(len(priming_data)):`

PR #269, opened 2020-09-15
Time series MinMax normalization

paxcema:
This PR introduces a MinMaxScaler into the Lightwood RNN time series encoder. For now it operates over any and all data, but soon it will only be applied for numerical columns (timestamps will get a different treatment). 

It doesn't say that much, but the unit test synthetic dataset now converges an order of magnitude quicker with better accuracy when decoding. We are currently adding timeseries datasets to the benchmarking suite to confirm that this is an improvement across the board.

NOTE: this PR assumes the data will be provided as numeric arrays. It will stay as a draft until this is accomplished in MindsDB Native.

Changes:
- Add MinMaxNormalizer
- Fixed couple of encoding bugs
- Unified test that considers both single-dimensional and multi-dimensional inputs

### DIFF
--- a/lightwood/encoders/time_series/helpers/rnn_helpers.py
+++ b/lightwood/encoders/time_series/helpers/rnn_helpers.py
@@ -6,6 +6,7 @@ import torch.nn as nn
 import numpy as np
 from sklearn.preprocessing import MinMaxScaler
 
+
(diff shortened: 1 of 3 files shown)


George3d6:
closing as it is merged into the hotifx.

Issue #265, opened 2020-09-12
Encode timeseries dates & datetimes


George3d6:
Before the timeseries encoding we should have a transformation that extracts certain features from date and datetime typed column, potentially in a similar manner to what the `DateTime` encoder does currently.

paxcema:
#273 closes this.

PR #273, opened 2020-09-18
Datetime sinusoidal normalizer

paxcema:
~~Added a class to transform a datetime() object into a tensor of sinusoidal descriptors. Integration to the actual TS RNN encoder will be done after the joint column encoding feature is deployed.~~

Adds capability of encoding each date/time stamp using sinusoidal functions to the `DatetimeEncoder` class.

This PR also fixes a small bug in the text RNN decoding step and a `batch_size` argument for the hidden state initializer.


lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -93,3 +97,45 @@ def fit_transform(self, x):
 
     def inverse_transform(self, y):
         return self.scaler.inverse_transform(y)
+
+
+class DateNormalizer:
```
George3d6, line 102: Would it not be possible to share this with the current date encoder ?

paxcema: Do you mean as in adding these functionalities to the encoder class instead?

paxcema: Okay, functionality is migrated into the date encoder, looks better now.

George3d6 (approved):

George3d6:
By the way, this closes #265

Issue #261, opened 2020-09-04
Add sklearn selection mixer


George3d6:
Add an sklearn mixer that is able to:

1. Pick a list of "potential" models for the given dataset based on some heuristics (e.g. nr of columns, dataset size, target type)
2. Fit all of those models on the training datasets
3. Evaluate all models on the test dataset
4. Fit the "winner" on the test+train dataset and export it as the final mixer

PR #277, opened 2020-09-23
Sklearn mixer

maximlopin:
https://github.com/mindsdb/lightwood/issues/261
- Add `SklearnMixer` + test


lightwood/api/data_source.py
```diff
@@ -55,7 +56,7 @@ def __setattr__(self, name, value):
 
 
 class DataSource(Dataset):
-    def __init__(self, data_frame, config, prepare_encoders=True):
+    def __init__(self, data_frame, config, prepare_encoders=True, initialize_transformer=True):
```
George3d6, line 59: Is the addition of `initialize_transformer` purely a measure for optimization ? I.e. so that no new ones get created when we create a SubSet.

Or is it somehow necessary for this PR ? If not, I think it's a very good catch that should have been spotted a while ago, but please consider making separate PRs for these in the future.

lightwood/api/data_source.py
```diff
@@ -305,6 +316,22 @@ def _prepare_column_encoder(self, config, is_target=False, training_data=None):
...
+    def input_feature_names(self):
+        return [feature['name'] for feature in self.config['input_features']]
+
+    @property
+    def output_features(self):
+        return self.config['output_features']
+
+    @property
+    def input_features(self):
+        return self.config['input_features']
```
George3d6, line 333: Please just define these as getters, i.e. `input_features`.

Usually I want to avoid using @property for anything *but* external interfaces (where I use it a few times for backwards compatibility).

If the values are determined at construction time, set them in the constructor as `self.output_feature_names`... etc.

If they aren't, these should be normal getter methods.

Unless there's a reason why these have to be the way they are.

George3d6: I see now that in `fit_datasource` we were previously creating these properties, so doing this change (replacing them with getters) might require changing things in a few more places. Still, my original comment stands, but if you think this can be useful then please make a case.

Usually I'd want to introduce as few concepts in the codebase as possible, and `@property` is something I've managed to avoid until now outside of cases where it was really necessary.

lightwood/mixers/nn.py
```diff
@@ -150,22 +150,29 @@ def fit(self, train_ds, test_ds, stop_training_after_seconds=None):
             stop_training_after_seconds = self.stop_training_after_seconds
 
         self.fit_data_source(train_ds)
+        self.fit_data_source(test_ds)
+
+        assert train_ds.transformer is test_ds.transformer
+        assert train_ds.encoders is test_ds.encoders
```
George3d6, line 156: Why are these asserts needed here ?

Seems to me like these should be part of the tests for nn.py, not part of the logic here.

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+CLASSIFICATION_MODELS = [
+    (SGDClassifier, {}),
+    (SVC, {}),
+    (GaussianNB, {}),
+]
+
+REGRESSION_MODELS = [
+    (LinearRegression, {}),
+    (SGDRegressor, {}),
+]
```
George3d6, line 21: Any particular reason for the choice of models ? 

If so, please make some comments as to why.

It will fall to the "research" side of things to optimize this mixer anyway, so don't worry too much about picking "the best" models, but it's worth noting down the intutions/experiments that made you pick these one since it will ease @paxcema 's job.

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+        self.fit_data_source(train_ds)
+
+        self.targets = {}
+
+        if test_ds is None:
+            test_ds = train_ds.subset(0.1)
+        else:
+            self.fit_data_source(test_ds)
+            assert test_ds.transformer is train_ds.transformer
+            assert test_ds.encoders is train_ds.encoders
```
George3d6, line 44: Again, why assert here and not as part of the tests ?

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        :param train_ds: DataSource
+        :param test_ds: DataSource
+        """
+        self.fit_data_source(train_ds)
+
+        self.targets = {}
+
+        if test_ds is None:
+            test_ds = train_ds.subset(0.1)
```
George3d6, line 40: This should happen in `predictor.py` since multiple mixers will require a test_ds.

George3d6: Also, probably not an ideal way to do this at all, since `subset` will not "remove" the rows from the `train_ds`. And the idea with the test_ds is that it should contain different rows from train_ds.

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+        self.encoders = train_ds.encoders
+        
+        for output_feature in train_ds.output_features:
+            self.targets[output_feature['name']] = {
+                'type': output_feature['type']
+            }
+            if 'weights' in output_feature:
+                self.targets[output_feature['name']]['weights'] = output_feature['weights']
+            else:
+                self.targets[output_feature['name']]['weights'] = None
```
George3d6, line 56: Shouldn't this be part of `fit_data_source` ?

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+
+        X_train = []
+        for row in train_ds:
+            X_train.append(np.array(row[0]))
```
George3d6, line 60: Why use `row[0]` here ? is the row shaped in a batch friendly way ? Where the 1-st dimension, if itteration 1 by 1, is just an empty dimension made to allow for easier batch concatentation ?

I assume this is the case but I forget to be perfectly honest.

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+
+        X_test = []
+        for row in test_ds:
+            X_test.append(np.array(row[0]))
+
+        for target_col_name in self.targets:
+            Y_train = train_ds.get_column_original_data(target_col_name)
+            Y_test = test_ds.get_column_original_data(target_col_name)
```
George3d6, line 68: Is it ideal to do this regardless of what type our target is ? I.e.  numeric I'd assume it's easier to predict the min-maxed numbers.

For categorical... aren't the sklearn models able to predict OHE targets ?

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+
+                if self.quantiles is not None:
+                    self.targets[target_col_name]['quantile_models'] = {}
+                    for i, quantile in enumerate(self.quantiles):
+                        self.targets[target_col_name]['quantile_models'][i] = best_model_class(**best_model_kwargs)
+
+            else:
+                self.targets[target_col_name]['model'] = None
+        
+        # Fit best model of each column on [train_ds + test_ds]
```
George3d6, line 137: I was expecting I'd have to make a comment reminding you to do this bit, I am oh so very happy this is already here ^^

lightwood/mixers/sklearn_mixer.py
```diff
@@ -0,0 +1,185 @@
...
+        for target_col_name in self.targets:
+            Y_train = train_ds.get_column_original_data(target_col_name)
+            Y_test = test_ds.get_column_original_data(target_col_name)
+            X, Y = (X_train + X_test), (Y_train + Y_test)
+
+            if self.targets[target_col_name]['model'] is not None:
+                self.targets[target_col_name]['model'].fit(X, Y)
+
+            if 'quantile_models' in self.targets[target_col_name]:
```
George3d6, line 146: Is Mindsdb_native currently setting `quantile_models` ? Or is this a feature to be implemented later ?

George3d6 (changes requested):
Few small nitpicks but overall an amazing job :+1:

lightwood/mixers/sklearn_mixer.py
maximlopin: `DataSource.subset()`  documentation:
```
Removes :percentage: of data randomly from itself and returns a new
datasource made from that data
```

lightwood/mixers/sklearn_mixer.py
maximlopin: it's set here: https://github.com/mindsdb/lightwood/blob/ff4e9edb3df097ce2247359d2a4d10586a5d356f/lightwood/mixers/sklearn_mixer.py#L132

lightwood/api/data_source.py
maximlopin: it's used for creating instances of `DataSource` in methods `DataSource.make_child`, `DataSource.subset`

lightwood/mixers/nn.py
maximlopin: because it's assumed that test_ds is the same as train_ds (same columns, encoders, transformer) but with different rows, it will raise assertion error if someone attempts to use different datasources for train and test

lightwood/mixers/sklearn_mixer.py
maximlopin: my focus was just on making this work, so I quickly chose a few popular regression and classification models

lightwood/mixers/sklearn_mixer.py
maximlopin: https://github.com/mindsdb/lightwood/pull/277#discussion_r495425909

lightwood/mixers/sklearn_mixer.py
maximlopin: test_ds is now required

lightwood/mixers/sklearn_mixer.py
maximlopin: it returns a tuple: `(input_vector, output_vector)`

lightwood/mixers/sklearn_mixer.py
maximlopin: boost mixer uses get_column_original_data so i used it for sklearn mixer too
we can make an issue and change it for both mixers in a separate PR

lightwood/mixers/sklearn_mixer.py
George3d6: Ah yes, of course, since this is a pytorch datasource -_-

Nevermind, that was a silly brain-freeze on my end.

lightwood/mixers/sklearn_mixer.py
George3d6: Agree

George3d6 (approved):

PR #279, opened 2020-09-23
replace dill with pickle

maximlopin:


George3d6:
Ok, so, I remembered why dill was a thing, passing the pickled objects over the netwrok, easier with dills.
The plan was to replace mindsdb pickles with dills, but we needed versioned storage (which we now have), so keep that in please

Issue #263, opened 2020-09-11
Joint encoding of time series columns


paxcema:
The time series encoder should be able to jointly operate over a set of `{order-by, target}` columns as a multi-dimensional array of shape `[batch_size, window_size, (encoded_target_sample_size + encoded_order_by_sample_size)]` for autoregressive time series tasks.

PR #281, opened 2020-09-24
Joint encoding in time series

paxcema:
This PR adds autoregressiveness capability to the time series encoder. This is done by training and encoding using data from `previous_` columns as given by MindsDB Native, in addition to the encoder's own column data.

Related: #263


lightwood/api/data_source.py
```diff
@@ -181,11 +181,13 @@ def __getitem__(self, idx):
                 col_config = self.get_column_config(col_name)
 
                 if col_name not in self.encoded_cache:  # if data is not encoded yet, encode values
-                    if not ((dropout_features is not None and col_name in dropout_features) or not self.enable_cache):
+                    if not ((dropout_features is not None and col_name in dropout_features) or not self.enable_cache or
+                            'previous_' in col_name):  # TODO: fragile check for name
```
George3d6, line 185: check should be:

`col_name.startswith('previous_') and col_name in self.outpout_features`

lightwood/api/data_source.py
```diff
@@ -198,6 +200,8 @@ def __getitem__(self, idx):
                         custom_data = {col_name: [None]}
 
                     sample[feature_set][col_name] = self.get_encoded_column_data(col_name, custom_data=custom_data)[0]
+                elif 'previous_' in col_name:   # TODO: fragile check for name
```
George3d6, line 203: See comments above on this check (note, if that one is too hard checking with `.startswith` is still better.

lightwood/api/data_source.py
```diff
@@ -331,14 +340,27 @@ def _prepare_encoders(self):
 
         for config in self.config['input_features']:
             column_name = config['name']
-            encoder_instance = self._prepare_column_encoder(
-                config,
-                is_target=False,
-                training_data=input_encoder_training_data
-            )
+            if 'previous_' in column_name:  # TODO: fragile check, should be by position
```
George3d6, line 343: Same as above.

lightwood/api/data_source.py
```diff
@@ -331,14 +340,27 @@ def _prepare_encoders(self):
...
+                column_data = self.get_column_original_data(column_name)
+
+                input_encoder_training_data['previous'].append({'data': column_data,
+                                                                'name': column_name,
+                                                                'output_type': config['type']
+                                                                })
+
+        for config in self.config['input_features']:
+            column_name = config['name']
+            if 'previous_' not in column_name:  # TODO: fragile check, should be by position
```
George3d6, line 353: can just check if the column name is in `input_encoder_training_data['previous']`

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -7,17 +6,14 @@
 
 def tensor_from_series(series, device, n_dims, pad_value, max_len=None, normalizer=None):
     """
-    :param series: list of lists, corresponds to time series: [[x1_1, ..., x1_n], [x2_1, ..., x2_n], ...]
+    :param series: list of lists, corresponds to a time series: [[x1_1, ..., x1_n], [x2_1, ..., x2_n], ...]
                    the series is zero-padded on each axis so that all dimensions have equal length
```
George3d6, line 10: Is the series zero-padded as part of this method, or should the series being passed already be zero-padded ?

It seems to be padded in here, so maybe mention that (or rather, make it clearer that it's the case as part of the comment)

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -28,17 +24,24 @@ def tensor_from_series(series, device, n_dims, pad_value, max_len=None, normaliz
...
-    # normalize and transpose
     if normalizer:
-        tensor = torch.Tensor([normalizer.encode(s) for s in series][0])
+        if isinstance(normalizer, MinMaxNormalizer):
+            tensor = torch.Tensor([normalizer.encode(s) for s in series]).transpose(0, 1)
+        else:
+            tensor = torch.Tensor([normalizer.encode(s) for s in series][0])
+
+        if len(tensor.shape) > 2:
+            tensor = tensor.squeeze(2)
```
George3d6, line 34: When can this be the case ? Is it not only when a normalizer other then MinMax is used (i.e. the date one) ? If so, just make it part of that `else`.

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -28,17 +24,24 @@ def tensor_from_series(series, device, n_dims, pad_value, max_len=None, normaliz
...
     else:
-        tensor = torch.transpose(torch.Tensor(series), 0, 1)
+        tensor = torch.Tensor(series).transpose(0, 1)
```
George3d6, line 36: `.transpose(0, 1)` seems to be called once here, in some situation ,and once above for the MinMaxNormalizer.

Again, I'm not sure what the exact logic is but it seems to me like this second else/if could be made part of the one above ? Maybe not, just a thought.

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -28,17 +24,24 @@ def tensor_from_series(series, device, n_dims, pad_value, max_len=None, normaliz
...
+
+    # dimension padding
+    for _ in range(max(0, n_dims - tensor.shape[-1])):
+        pad = torch.full((tensor.shape[0], 1), pad_value)
+        tensor = torch.cat([tensor, pad], dim=-1)
```
George3d6, line 41: One thing that might be worth noting on here is if the padding is being done "forward" or "backwards".

It should be done "backwards" I think, i.e. [1,2,3] where the max length is `5` then we have [0,0,1,2,3] because the bigger and index is the more recent the value, and usually (always?) when padding is needed it's older data that's missing.

lightwood/encoders/time_series/helpers/rnn_helpers.py
```diff
@@ -28,17 +24,24 @@ def tensor_from_series(series, device, n_dims, pad_value, max_len=None, normaliz
...
 
-    # add batch dimension
+    # remove nans and add batch dimension
+    tensor[torch.isnan(tensor)] = 0.0
```
George3d6, line 44: Any chance we can rename `tensor` or `tensorized_series` or just `series` or `series_t` ?

I was actually a bit confused here thinking `tensor` is a torch-related construct rather than a variable. But maybe it's only me, in which case keep it as is. If you feel the same way however, make rename it.

George3d6 (changes requested):
This is only a partial review, will get back to it when I have time.

lightwood/api/data_source.py
paxcema: Second clause seems off, `previous_` columns are input features.

How about this check: `col_name.startswith('previous_') and feature_set == 'input_features'`?

lightwood/encoders/time_series/helpers/rnn_helpers.py
paxcema: You are right, it should be backwards

lightwood/encoders/time_series/helpers/rnn_helpers.py
paxcema: Agreed, will change

lightwood/api/data_source.py
George3d6: Yes, but it's `previous_{name}` where `name` is always the name of an output feature.

lightwood/mixers/helpers/transformer.py
```diff
@@ -14,10 +14,11 @@ def transform(self, sample):
...
         for input_feature in self.input_features:
-            sub_vector = sample['input_features'][input_feature].tolist()
-            input_vector.extend(sub_vector)
-            if input_feature not in self.feature_len_map:
-                self.feature_len_map[input_feature] = len(sub_vector)
+            if 'previous_' not in input_feature:  # TODO: fragile check for name
+                sub_vector = sample['input_features'][input_feature].tolist()
+                input_vector.extend(sub_vector)
+                if input_feature not in self.feature_len_map:
+                    self.feature_len_map[input_feature] = len(sub_vector)
```
George3d6, line 21: Why do we need to ignore the `previous_` features here ?

This seem like an unreliable way to do this, since we might want to jointly encode many more columns in the future.

Is it not easier to simply support "empty" encoders in the datasource, where a particular column is encoded to an empty list ?

lightwood/encoders/time_series/rnn.py
```diff
@@ -28,18 +28,24 @@ def __init__(self, encoded_vector_size=4, train_iters=100, stop_on_error=0.01, l
         self._eos = 0.0  # end of input sequence -- padding value for batches
         self._n_dims = ts_n_dims
         self._normalizer = None
+        self._target_ar_normalizer = None
         self._criterion = nn.MSELoss()
 
-    def setup_nn(self):
+    def setup_nn(self, additional_targets=None):
```
George3d6, line 34: Why do we need to pass the `additiona_targets` here ? As in, since the timeseries encoder can handle arrays of arbitrary sizes would it not make more sense not to have any "special" logic around previous target values ?

Think, for example, of a case where we not only want to encoder the order column + the previous targets, but rather jointly encode 3x or 4x different columns with the same RNN encoder, this approach wouldn't scale to that.

lightwood/encoders/time_series/rnn.py
```diff
@@ -48,29 +54,39 @@ def to(self, device, available_devices):
...
+            self._max_ts_length = max(len(priming_data[i][0]), self._max_ts_length)  # TODO: this is set at Native... should we still check?
 
+        # normalize data
         if self._normalizer:
             self._normalizer.prepare(priming_data)
 
-        # decrease for small datasets
+        if previous_target_data is not None and len(previous_target_data) > 0:
+            target_dict = previous_target_data[0]
+            normalizer = MinMaxNormalizer()  # TODO: normalizer determined by subtype (for datetime previous_ columns support)
```
George3d6, line 84: I think your native PR is almost ready for merging, maybe merge that into `staging` and then add the subtype check here.

lightwood/encoders/time_series/rnn.py
```diff
@@ -81,23 +97,31 @@ def prepare(self, priming_data, feedback_hoop_function=None, batch_size=256):
 
             while data_idx < len(priming_data):
```
George3d6, line 98: Ok, a bit off-topic, but isn't it easy enough to use a DataLoader for this instead of manually doing batching :?

George3d6 (commented):
Overall looks fine, can only find relatively minor issues and comments (e.g. see the handling arbitrary amount of columns one) that would be hard to impl, if this version works we can start with this one and move from here.

lightwood/encoders/time_series/helpers/rnn_helpers.py
paxcema: Yup, padded here, will update docstring.

lightwood/encoders/time_series/helpers/rnn_helpers.py
paxcema: Never, it's a leftover that I forgot about, good catch

paxcema: Scratch that, it is used and only for the MinMax, point taken.

lightwood/encoders/time_series/rnn.py
paxcema: Yeah, I agree. We should aim for that in a new PR.

lightwood/api/data_source.py
paxcema: Ah, I get what you meant now

lightwood/encoders/time_series/rnn.py
paxcema: An alternative would be to pass the `Transformer.feature_len_map` instead

George3d6 (approved):
I think merging this before the lightwood release today is ideal.

Issue #239, opened 2020-08-05
Challenge, TS encoders


torrmal:
`Warning: This issue is intended as a test to be given to candidates for a job at Mindsdb, please do not start working on it, unless we've message you telling you to do so.`

## Current state

We presently have an RNN-based encoder for timeseries data, which will take a sequence of arbitrary length, e.g. `[1, 2, 3, 4, 5, 6, .... , 2000]` and turn it into an embedding of a fixed size.'

It can be found here:

https://github.com/mindsdb/lightwood/blob/stable/lightwood/encoders/time_series/rnn.py

The unit tests for it can be found here:

https://github.com/mindsdb/lightwood/blob/stable/tests/unit_tests/encoders/time_series/test_rnn.py

They can be executed by cloning lightwood, installing the requirements (`pip3 install -r requirements.txt`), adding lightwood to your python path (e.g. `export PYTHONPATH=$PYTHONPATH:/home/<your user>/lightwood`) and then cd-ing into the cloned directory and running `tests/unit_tests/encoders/time_series/test_rnn.py`.

## The task

You need to create ONE of two new proposed time-series encoders:

- *Transformer encoder for TS*: The issue with RNNs is that they are sequential and therefore you can only parallelize on batches with padding. Transformers solve this, therefore the task at hand is to build a time series encoder based the transformer architecture. Please find inspiration here: https://towardsdatascience.com/how-to-code-the-transformer-in-pytorch-24db27c8f9ec

- *CNN encoder for TS*:  Implement a way to encode time-series data using CNNs, we invite you to use Restnet as the encoder (see inspiration here: https://github.com/hsinyilin19/ResNetVAE). *Bonus points if you use, Residual Attention.*

## Notes

* The specific unit tests to be written are left as an exercise for the implementer, the scope of the unit tests should be to assess performance as well as to see if the autoencoder is doing roughly what would be expected of it. You needn't validate anything about how the resulting embeddings behave when feeding further down the line into the mixer.

* Please ask any questions you might have here, unless they need to be private, since that way they can remain here as an FAQ

* Message us when you are done with this project, don't make a PR (since it would be visible to other participants that might not have finished yet)

* Ideally your encoder should run on both GPU and CPU, but if you only had time to make it work on one of them, tell us which

* Feel free to rewrite as much of the existing code as you wish, as long as the underlying functionality stays the same whatever changes you make should be fine.

walexand3r:
Hi, I have a question: 
[...]

PR #285, opened 2020-09-25
Transformer. Solves #239

Lezcano:
Implements the transformer encoder to solve #239 
In particular, this PR impelements:

 - Transfoermer Encoder
 - Parallel Batching
 - Truncated Back-Propagation Through Time (TBPTT)
 - CPU and GPU
 - Supports different length sequences
 - Any mix of the above

With a few more lines, it would be direct to implement the encoding of multidimensional sequences, but I have not done so.

This PR also deletes the `__init__.py` files from the test folders, as suggested by  @ZoranPandovski and renames a test file, as it was creating a conflict with a different file which had the same name.


lightwood/encoders/time_series/helpers/transformer_helpers.py
```diff
@@ -0,0 +1,78 @@
...
+        self.init_weights()
+
+    def _generate_square_subsequent_mask(self, sz):
+        mask = (torch.triu(torch.ones(sz, sz)) == 1).transpose(0, 1)
+        # Take the logarithm
+        mask = (
+            mask.float()
+            .masked_fill(mask == 0, float("-inf"))
+            .masked_fill(mask == 1, float(0.0))
+        )
```
George3d6, line 54: I'm not quite sure I understand the purpose of this code.

To me it seems like it generates a matrix of `sz`x`sz` where the "lower" triangle + the diagonal are `0` and the "upper" triangle is `-inf` but I fail to see how this mask is helpful/relevant.

Maybe add a comment explaining how it works here or above the forward call to the `transformer_encoder` where it is used.

But @paxcema is the expert here, so maybe this is super obvious and I'm missing something

lightwood/encoders/time_series/helpers/transformer_helpers.py
```diff
@@ -0,0 +1,78 @@
...
+import torch
+import torch.nn as nn
+
+
+def len_to_mask(lengths, zeros):
+    # Clean trick from:
+    # https://stackoverflow.com/questions/53403306/how-to-batch-convert-sentence-lengths-to-masks-in-pytorch
+    mask = torch.arange(lengths.max())[None, :] < lengths[:, None]
+    if zeros:
+        mask = ~mask  # Logical not
```
George3d6, line 11: Might be worth being explicit here (in a comment and/or `dtype` cast if `mask` is intended to have type bool or int since bitwise not doesn't operate on `0` and `1` the same way it would operate on `False` and `True` (I assume the above code is meant for dtype==bool, but maybe I'm wrong. hence why a comment would help)

lightwood/encoders/time_series/helpers/transformer_helpers.py
Lezcano: This is part of the transformer architecture for processing time series. The attention in the transformer, a priori, does not take in account position (think of attention in images). This triangular mask is for the attention, so that words in a sentence can just attend to previous words. This, in turn, enables decoding, as the model that you train has just learnt to attend to previous positions.

lightwood/encoders/time_series/helpers/transformer_helpers.py
Lezcano: The mask has a `bool` type, as it is created by a comparison operator. I will add a comment and proper documentation to this and to other functions (I forgot to write the docs before the PR)

lightwood/encoders/time_series/transformer.py
```diff
@@ -0,0 +1,196 @@
...
+                loss = sum(losses)
+                loss.backward()
+                torch.nn.utils.clip_grad_norm_(self._encoder.parameters(), 0.5)
+                self._optimizer.step()
+
+                total_loss += loss.item() * len(train_batch)
+
+            average_loss = total_loss / npoints
+            if average_loss < self._stop_on_error:
+                break
```
George3d6, line 162: Would you agree that a better approach here is:

1. Have a small validation subset
2. Look at the accuracy on that validation subset
3. If that accuracy drops for a given number of epochs, stop training and take the model that had the best accuracy on the validation subset
4. Or stop training if a given time limit is reached (rather than nr of epochs) and, again, pick the best model on the valdiation data
5. Run another epoch or two where the model is also trained on the validation  data

That is the approach we currently use for our mixer, I'm curios if you think it's preferable ? If so, no worries, no need to adopt it right now, was working on a generic way of implementing this for all encoders in `gym.py ` which I have to revisit at some poinrt.

Lezcano: The approach I implemented is that I saw in `rnn_encoder.py`. I believe that it is a fairly rough one, but it is not clear to me how large are usually the sets sent to this `prepare` method. If they are very very small, then maybe it would be enough as it is, although the model would probably heavily overfit the data for a sufficiently large number of steps.

I think that, in genera, it would be better the approach that you have sketched above, i.e. that of a training with early stopping, to avoid overfitting to the training set when the datasets are quite small

George3d6: The `prepare` method is currently given the full training dataset (although not the validation dataset that is used for the "mixer", i.e. the component that comes after the encoders), so currently we design the encoders to work with the full dataset, it would be fairly trivial to also give them a validation datasets.

At any rate, the approach here is good, I was just curious what you thought about the above since (hopefully) it's some version of that which I'll be moving all trainable encoders too soon(tm).

lightwood/encoders/time_series/helpers/transformer_helpers.py
```diff
@@ -0,0 +1,85 @@
...
+                device
+            )
+        src = self.encoder(src)
+        src = self.pos_encoder(src)
+        # The lengths_mask has to be of size [batch, lengths]
+        lengths_mask = len_to_mask(lengths, zeros=True).to(device)
+        output = self.transformer_encoder(
+            src, mask=self.src_mask, src_key_padding_mask=lengths_mask
+        )
+        output = self.decoder(output)
```
paxcema, line 84: I guess the only thing missing here is to return the encoded tensor in addition to the decoded output

lightwood/encoders/time_series/transformer.py
```diff
@@ -0,0 +1,194 @@
...
+        # It would be worth to consider the use of a scheduler with a warm-up in some datasets
+        self._optimizer = optim.AdamW(
+            self._encoder.parameters(), lr=self._learning_rate, weight_decay=1e-4
+        )
+        self._criterion = nn.MSELoss(reduction="none")
+
+        self._prepared = False
+        self.bptt = 35
+
+        # Lezcano: These should be global constants of the library
```
paxcema, line 44: Agreed

lightwood/encoders/time_series/transformer.py
```diff
@@ -0,0 +1,194 @@
...
+                    )
+                    data = data.unsqueeze(-1)
+                    output = self._encoder(data, lengths_chunk)
+                    output = output.squeeze(-1)
+                    losses.append(
+                        self._masked_criterion(output, targets, lengths_chunk)
+                    )
+                loss = sum(losses)
+                loss.backward()
+                torch.nn.utils.clip_grad_norm_(self._encoder.parameters(), 0.5)
```
paxcema, line 155: The `0.5` value should probably be a class parameter

tests/unit_tests/encoders/time_series/test_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # Test data
+        example = torch.tensor(example)
+        correct_answer = example[:, 1:]
+        # Decoder overfit, discard last element as it doesn't correspond to answer
+        answer = torch.tensor(encoder.encode(example))[:, :-1]
+        # Round answer
+        answer = answer.round()
+        n = correct_answer.numel()
+        correct = (correct_answer == answer).sum()
```
paxcema, line 64: Why the sum instead of something like `torch.allclose()`?

lightwood/encoders/time_series/transformer.py
```diff
@@ -0,0 +1,194 @@
...
+                self._optimizer.step()
+
+                total_loss += loss.item() * len(train_batch)
```
paxcema, line 158: I'm not sure I get the `*len(train_batch)` operation here. `loss` holds the sum of the average loss in every chunk, right?

paxcema:
I like this PR, code is very well structured and can help us simplify the current approach in the RNN encoder as well. Truncated BPTT is a good idea to tackle big datasets, too.

tests/unit_tests/encoders/time_series/test_transformer.py
Lezcano: This counts the number of elements that are equal in `answer` and `correct_answer`, as there might be errors (and there are errors in practice.

lightwood/encoders/time_series/transformer.py
Lezcano: At the moment we do stochastic gradient descent, and as such, we compute the batch loss as 1/B \sum_{i=1}^B \grad f(x_i, \theta). As such, we multiply by the batch size to recover the sum of the losses, to then compute the average loss per element of the whole dataset in the outside loop.

lightwood/encoders/time_series/helpers/transformer_helpers.py
Lezcano: As discussed in the first issue, I am not sure that the encoded sequence should be returned in an seq2seq model, but I can add it if you think it is necessary.

PR #288, opened 2020-09-30
fix mixers for tags

maximlopin:
- Add support for tags data type in `SklearnMixer`
- FIx schema of `depends_on_column` key in column config

### DIFF
--- a/lightwood/mixers/base_mixer.py
+++ b/lightwood/mixers/base_mixer.py
@@ -162,11 +162,9 @@ class BaseMixer:
     def _apply_accuracy_function(col_type, reals, preds, weight_map=None, encoder=None):
         if col_type == COLUMN_DATA_TYPES.CATEGORICAL:
             if weight_map is None:
-                sample_weight = [1 for x in reals]
+                sample_weight = [1] * len(reals)
             else:
-                sample_weight = []
-                for val in reals:
-                    sample_weight.append(weight_map[val])
+                sample_weight = [weight_map[val] for val in reals]
 
             accuracy = {
                 'function': 'accuracy_score',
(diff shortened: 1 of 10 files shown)


George3d6 (changes requested):
Please keep the booster mixer, since it's a good example mixer and we might want to use it in favor of sklearn for some tasks because training is quicker and in a lot of problems gradient boosters seems to be a very good speed/accuracy intersection

lightwood/mixers/base_mixer.py
```diff
@@ -162,23 +162,22 @@ def calculate_accuracy(self, ds):
...
-                sample_weight = [1 for x in reals]
+                sample_weight = [1] * len(reals)
             else:
-                sample_weight = []
-                for val in reals:
-                    sample_weight.append(weight_map[val])
+                sample_weight = [weight_map[val] for val in reals]
+
+            if encoder is None:
+                raise ValueError('you must provide encoder to compue accuracy for tags data type')
```
paxcema, line 180: typo: `compue` -> `compute`

lightwood/mixers/boost.py
```diff
@@ -36,6 +42,22 @@ def _fit(self, train_ds, test_ds=None):
                 self.targets[target_col_name]['model'] = GradientBoostingClassifier(n_estimators=600)
                 self.targets[target_col_name]['model'].fit(X, Y, sample_weight=sample_weight)
 
+            elif self.targets[target_col_name]['type'] == COLUMN_DATA_TYPES.MULTIPLE_CATEGORICAL:
+                weight_map = self.targets[target_col_name]['weights']
+                if weight_map is None:
+                    sample_weight = [1] * len(Y)
+                else:
+                    sample_weight = [weight_map[val] for val in Y_train]
```
paxcema, line 50: Is `Y_train` defined in this method?

lightwood/mixers/sklearn_mixer.py
```diff
@@ -92,28 +139,29 @@ def _fit(self, train_ds, test_ds):
...
-
                 if self.quantiles is not None:
                     self.targets[target_col_name]['quantile_models'] = {}
                     for i, quantile in enumerate(self.quantiles):
                         self.targets[target_col_name]['quantile_models'][i] = best_model_class(**best_model_kwargs)
 
             else:
                 self.targets[target_col_name]['model'] = None
-        
+                continue
```
paxcema, line 149: Seems to me like this `continue` is not needed?

paxcema (approved):

PR #295, opened 2020-10-09
Categorical target support in time series

paxcema:
- Adds `CatNormalizer()` to one-hot encode categorical target column when fed into the RNN TS network.
- Improved / simplified logic in `setup_nn()` method

### DIFF
--- a/lightwood/encoders/time_series/rnn.py
+++ b/lightwood/encoders/time_series/rnn.py
@@ -42,8 +42,15 @@ class RnnEncoder(BaseEncoder):
 
         total_dims = self._n_dims
         if additional_targets:
-            for _ in additional_targets:
-                total_dims += 1  # We are assuming target dim == 1
+            for t in additional_targets:
+                if t['original_type'] == 'categorical':
+                    t['normalizer'] = CatNormalizer()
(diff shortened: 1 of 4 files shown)


George3d6 (commented):
Needs a test :p

lightwood/encoders/time_series/rnn.py
```diff
@@ -42,8 +42,15 @@ def setup_nn(self, additional_targets=None):
 
         total_dims = self._n_dims
         if additional_targets:
-            for _ in additional_targets:
-                total_dims += 1  # We are assuming target dim == 1
+            for t in additional_targets:
+                if t['original_type'] == 'categorical':
+                    t['normalizer'] = CatNormalizer()
+                    t['normalizer'].prepare(t['data'])
+                    total_dims += len(t['normalizer'].scaler.categories_[0])
```
George3d6, line 49: Is this adding 1 dimension per category basically ? If so I feel like it could get way too compute heavy very quickly, but for now I guess it works.

paxcema: This extends the input tensor dimension by the # of categories observed at train time. The encoder should stay at manageable computing requirements.

PR #301, opened 2020-10-15
Fix for https://github.com/mindsdb/mindsdb/issues/798

neil007m:
logs what device is being used in `learn` and `predict` methods [here](https://github.com/mindsdb/lightwood/blob/stable/lightwood/api/predictor.py)

### DIFF
--- a/lightwood/api/predictor.py
+++ b/lightwood/api/predictor.py
@@ -92,6 +92,8 @@ class Predictor:
         :param test_data: DataFrame or DataSource
             The data to test accuracy and learn_error from
         """
+        device, _available_devices = get_devices()
+        log.info(f'Computing device used: {device}')
         # generate the configuration and set the order for the input and output columns
         if self._generate_config is True:
             self._input_columns = [col for col in from_data if col not in self._output_columns]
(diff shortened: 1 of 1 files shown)


paxcema:
Hello Neil! Thanks for contributing.

This looks good. I'm curious if you tested it though. I'm trying it out and there seems to be an issue with the logger, because the information is not shown.

neil007m:
@paxcema Hi Patricio, 

Actually, no, I did not when I submitted the PR. Sorry about that. 

I tried the following approach now to test:
- created a virtual environment (venv) 
- installed via pip -e (for my forked branch)
- ran the sample code in documentation with a simple csv file

It worked as expected, but no logs. I also manually overrode what [logger](https://github.com/mindsdb/lightwood/blob/stable/lightwood/logger/logger.py) does here and set it to various other levels like DEBUG and INFO, but didn't see anything logged on command line.  

Not sure if this is the correct approach, would appreciate any input.

paxcema:
I'm not seeing the specific ones to this PR. Maybe @StpMax has an idea of what might be the issue.

George3d6:
@paxcema @neil007m  this is probably a local logging config issue, since the proper logger is being used here. Will merge this for now. Thanks for this contribution @neil007m

PR #312, opened 2020-11-10
Convolutional Autoencoder for time-series. Solves #239

walexand3r:
Causal 1D CAE for time-series, solves #239 .

Simple initial model (full ResNet-18 (and 34) implementation on the way). 
Makes use of dilated causal 'temporal' convolution blocks with residual connections.
 - Encodes and decodes multidimensional time-series of variable sequence length
 - Works on CPU & GPU

Please see notes in cnn.py for full description and discussion of unit test results.


paxcema (changes requested):
For this encoder-decoder pair to be compatible with the rest of the time series codebase, we need the intermediate representation to be a tensor shaped: `[batch_size, length, embedding_dim]`. As per the paper and some experimenting of my own, this encoder modifies the time series length; which means we cannot use it yet, so I'll be closing the PR for now.

I merged most of the enc-dec usage into the common loop we have for time series, so that the only thing pending is the shape constraint; @walexand3r if you can find the time to modify this behavior, feel free to reopen the PR.

PR #315, opened 2020-11-18
Remove transformers==2.4.1 constraint for macOS

paxcema:
- Removed custom logic in `setup.py`, as error in #157 seems to not be happening anymore


ZoranPandovski (approved):
Looks good. One thing we can confirm is if this will work on 3.6 or 3.8 Python since the action is running only MacOS Python 3.7

paxcema:
Confirmed locally it works for 3.8, merging.

PR #318, opened 2020-11-19
Positive domain flag for numerical encoder

paxcema:
This is a (long overdue) quick fix for the negative value prediction issue in positive numerical tasks. It forces the numerical encoder to return the absolute magnitude of the prediction in case the corresponding flag is set to `True` (passed through `encoder_attrs` from MindsDB Native).

Once we move away from using the QuantileLoss, I'll experiment with loss functions that heavily penalize wrongly signed predictions. There's also the idea of mini-networks to cope with missing data, which should help in getting us to a point where this shortcut is not needed. For now though, it might be better than nothing.

Also, fixes a small bug in `NnMixer._fit()`


maximlopin (approved):
would be nice to have a test for this

Issue #235, opened 2020-08-03
"ValueError: month must be in 1..12" when having output column of type Date


maximlopin:
described here:
https://github.com/mindsdb/mindsdb_native/issues/92

PR #323, opened 2020-11-24
DateTime encoder improvements

paxcema:
Addresses #235 and [N#92](https://github.com/mindsdb/mindsdb_native/issues/92).

- Determine monthly number of days given the year and month, instead of using a fixed value as before.
- Normalizes intermediate representation components in .decode() to avoid errors when NnMixer encodes values that are negative and/or too large.
- Introduces test for this.


George3d6 (commented):
All looks fine, in hindsight the `decode` for this encoder was a horrible mess, and I didn't realize `day` (of month) was hardcoded. This is a much welcome improvement to a subpar encoder.

PR #325, opened 2020-11-26
Dropout for DefaultNet

paxcema:
- Adds dropout (`p = 0.3`) to the DefaultNet when used in the NnMixer, based on the results of our parameter search.
- Also simplifies logic in DefaultNet, deleting `deterministic` flag as most of that logic is already implemented in `__init__.py`

### DIFF
--- a/lightwood/mixers/helpers/default_net.py
+++ b/lightwood/mixers/helpers/default_net.py
@@ -59,7 +59,7 @@ class DefaultNet(torch.nn.Module):
             layers = []
             for ind in range(len(shape) - 1):
                 if 0 < ind < len(shape):
-                    layers.append(torch.nn.Dropout(p=0.5))
+                    layers.append(torch.nn.Dropout(p=CONFIG.DEFNET_DROPOUT))
                 linear_function = PLinear if CONFIG.USE_PROBABILISTIC_LINEAR else torch.nn.Linear
                 layers.append(linear_function(shape[ind],shape[ind+1]))
                 if ind < len(shape) - 2:
--- a/lightwood/mixers/helpers/default_net.py
+++ b/lightwood/mixers/helpers/default_net.py
@@ -58,6 +58,8 @@ class DefaultNet(torch.nn.Module):
 
             layers = []
             for ind in range(len(shape) - 1):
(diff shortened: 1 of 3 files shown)


lightwood/mixers/helpers/default_net.py
```diff
@@ -58,6 +59,8 @@ def __init__(self, dynamic_parameters,
 
             layers = []
             for ind in range(len(shape) - 1):
+                if (dropout is not None) and (not deterministic) and (0 < ind < len(shape)):
```
George3d6, line 62: Ok... two questions:

1. Why not apply dropout when `deterministic==True` ?

2. Why was `deterministic` set to `False` as the default ? (might have been me, granted, but if so it was a mistake)

We should aim for lightwood's results to be replicable.
Granted, `deterministic` might not be needed at all, since we set torch to be deterministic in `__init__.py`

So I guess we should solve `1.` and then just remove `deterministic` and move all seeding logic into `__init__.py`, since most is already replicated there (no pun intended).

George3d6: Also this should check `if dropout > 0` and set the default value to `0`, probably make more sense. but maybe not ? It just seems "better" to me that way style/logic wise, but if you disagree @paxcema then leave it as is.

lightwood/mixers/nn.py
```diff
@@ -155,7 +155,8 @@ def _fit(self, train_ds, test_ds, stop_training_after_seconds=None):
             input_size=len(input_sample),
             output_size=len(output_sample),
             nr_outputs=len(train_ds.out_types),
-            deterministic=self.deterministic
+            deterministic=self.deterministic,
+            dropout=CONFIG.DEFNET_DROPOUT if not self.deterministic else None
```
George3d6, line 159: Any reason for this to be in the `CONFIG` now that you're done experimenting ?

If so, it should be in the mixer config (part of the larger config that is passed when creating a lightwood predictor)

George3d6 (changes requested):

lightwood/mixers/nn.py
paxcema: Nope, no reason, good catch.

PR #326, opened 2020-11-26
Offset for non-categorical features in DataSource.output_weights

paxcema:
The call to `DataSource.output_weights` in the `NnMixer` did not account for offsets in this array when some of the output features were non categorical, which meant an empty tensor was passed to the CrossEntropyLoss, resulting in this [issue](https://github.com/mindsdb/mindsdb_native/issues/291). This PR fixes the bug.

### DIFF
--- a/lightwood/api/data_source.py
+++ b/lightwood/api/data_source.py
@@ -265,7 +265,7 @@ class DataSource(Dataset):
             self.output_weights_offset = {}
             for idx, (otype, oidxs) in enumerate(zip(self.out_types, self.out_indexes)):
                 self.output_weights_offset[idx] = self.output_weights_offset.get(idx-1, 0)
-                if otype != ColumnDataTypes.CATEGORICAL:
+                if otype not in (ColumnDataTypes.CATEGORICAL, ColumnDataTypes.MULTIPLE_CATEGORICAL):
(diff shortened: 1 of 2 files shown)


lightwood/api/data_source.py
```diff
@@ -260,6 +260,12 @@ def __getitem__(self, idx):
         if self.enable_cache:
             self.transformed_cache[idx] = sample
 
+        # Add offset in output_weights (output_features might be categorical _and_ numerical)
+        self.output_weights_offset = 0
+        for otype, oidxs in zip(self.out_types, self.out_indexes):
+            if otype != ColumnDataTypes.CATEGORICAL:
+                self.output_weights_offset += (oidxs[1] - oidxs[0])
```
paxcema, line 267: Actually, I now realize this does not work in the general case, so I'll be modifying it to comply with any ordering of any set of feature types

George3d6 (changes requested):
Doesn't seem to account for the multi-categorical case (`output_type == COLUMN_DATA_TYPES.MULTIPLE_CATEGORICAL`)

George3d6:
Other than that... I can't believe we had this bug for so long, nice job spotting it.

Issue #320, opened 2020-11-23
Slow training loop with QuantileLoss


paxcema:
* Lightwood version: last staging

Training on `us_health_insurance` is very slow for a dataset of that size. Changing the loss function from `QuantileLoss()` to `MSELoss()` makes the issue dissappear. We should understand why this happens.

paxcema:
Closed by #331

PR #331, opened 2020-12-03
Remove quantile ranges

George3d6:
## Why

* Non-conformist is being used in the model analysis phase of mindsdb native to generate confidence ranges
* This overrides the quantile confidences of lightwood
* Non-conformist should work for most if not all cases when lightwood is able to generate these ranges, so the work being done to generate them in lightwood is wasted time and complexity

## How

* Remove quantile loss from nn mixer
* Remove quantile models from sklearn and boost mixer
* Remove quatile loss helper class
* Remove extra_outputs from the numeric encoder (used for training with mixer when it had a quantile loss)
* Remove quantile range selection from the nn mixer


lightwood/mixers/nn.py
```diff
@@ -193,11 +191,11 @@ def _fit(self, train_ds, test_ds, stop_training_after_seconds=None):
...
                         self.unreduced_criterion_arr.append(torch.nn.BCEWithLogitsLoss(weight=weights_slice, reduce=False))
                 elif output_type == COLUMN_DATA_TYPES.NUMERIC:
-                    self.criterion_arr.append(QuantileLoss(quantiles=self.quantiles))
-                    self.unreduced_criterion_arr.append(QuantileLoss(quantiles=self.quantiles, reduce=False))
-                else:
                     self.criterion_arr.append(torch.nn.MSELoss())
                     self.unreduced_criterion_arr.append(torch.nn.MSELoss(reduce=False))
+                else:
+                    self.criterion_arr.append(torch.nn.torch.nn.MSELoss())
+                    self.unreduced_criterion_arr.append(torch.nn.torch.nn.MSELoss(reduce=False))
```
paxcema, line 198: Shouldn't this be a single `else:` clause? Also, typo: `torch.nn.torch.nn`

lightwood/mixers/nn.py
```diff
@@ -489,29 +440,12 @@ def _predict(self, when_data_source, include_extra_data=False):
...
                 predictions[output_column] = {'predictions': decoded_predictions}
 
             if awareness_arr is not None:
                 predictions[output_column]['selfaware_confidences'] = [1/abs(x[k]) if x[k] != 0 else 1/0.000001 for x in awareness_arr]
 
-            if when_data_source.out_types[k] in (COLUMN_DATA_TYPES.NUMERIC):
-                predictions[output_column]['confidence_range'] = []
-                predictions[output_column]['quantile_confidences'] = []
-
-                for i, pred in enumerate(decoded_predictions):
```
paxcema, line 505: This bit (`selfaware_confidences` key) is still used by the nonconformist normalizer

lightwood/encoders/numeric/numeric.py
```diff
@@ -100,9 +99,9 @@ def decode(self, encoded_values, decode_log=None):
                 else:
                     if decode_log:
                         sign = -1 if vector[0] > 0.5 else 1
-                        real_value = [math.exp(vector[i*2 + 1]) * sign for i in range(1 + self.extra_outputs)]
+                        real_value = math.exp[1] * sign
```
paxcema, line 102: `math.exp(vector[1])`

paxcema (changes requested):
An error is failing tests. Also, there's a bit that nonconformist still uses.

lightwood/mixers/nn.py
George3d6: I prefer to keep it explicit but... should be a comment about why no and else

lightwood/mixers/nn.py
George3d6: The `selfaware_confidences` confidence is not removed, only the `sc` which is a reference to it used to select the quantile

lightwood/encoders/numeric/numeric.py
George3d6: fixed

lightwood/mixers/nn.py
paxcema: Ah, I missed that, you're right

paxcema (approved):

paxcema:
This also closes #320

PR #333, opened 2020-12-04
Torch 1.7, Transformers 4.0

paxcema:
## Why
To keep Lightwood compatible with the latest version of PyTorch and Transformers.

Small caveat: PyTorch 1.7 will support Python 3.9 starting next version (1.7.1), which should be released this month.


ZoranPandovski (changes requested):
We should do the same changes for windows also https://github.com/mindsdb/lightwood/blob/stable/setup.py#L43

ZoranPandovski (approved):

PR #340, opened 2020-12-09
Automatic mixed precision

paxcema:
## Why
Automatic Mixed Precision (AMP) is a technique that enables using either full or half precision tensors at training time in a dynamic way, without compromising on accuracy while improving training runtime.

## How
As of PyTorch 1.6, it's rather easy to enable AMP while following recommended guidelines. This PR adds an AMP context manager for forward and loss operations, based on Torch's original context manager. We use a custom manager to check the device computing capability: if the GPU has no tensor cores, then enabling AMP yields a slower runtime. 

Another change is that now we cast the output to a generic tensor instead of a specific wrapper per each encoder, because doing so triggers errors given the dynamic change in type that can happen while using AMP.

Finally, there are some changes in the time series transformer encoder (+ tests, so that they can pass with this change).

~~The PR will stay as a draft until benchmarks confirm that we are at least on par with stable.~~ Done.

### DIFF
--- a/lightwood/helpers/torch.py
+++ b/lightwood/helpers/torch.py
@@ -31,8 +31,9 @@ class LightwoodAutocast:
     """
     def __init__(self, enabled=True):
         self.major = 0  # GPU major version
+        torch_version = [int(i) for i in torch.__version__.split('.')[:-1]]
 
-        if enabled and not torch.cuda.is_available():
+        if not enabled or not torch.cuda.is_available() or torch_version[0] < 1 or torch_version[1] < 6:
             self._enabled = False
         else:
             device, _ = get_devices()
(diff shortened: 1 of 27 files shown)


lightwood/helpers/torch.py
```diff
@@ -19,3 +22,49 @@ def concat_vectors_and_pad(vec_list, max_):
...
+            else:
+                self._enabled = False  # gpu is available but cpu is forced
+
+        self.prev = self._enabled
+
+    def __enter__(self):
+        # tensor cores only exist from 7 onwards
+        # if this is not the case, then AMP is unnecessary overhead
+        if self.major > 6:
+            self.prev = torch.is_autocast_enabled()
```
George3d6, line 52: Why does this need to be checked with every `__enter__` ?
Why not check it once in the `__init__` and then just use `self._enabled` here, and never even use the `self.prev` variable.

George3d6 (changes requested):
I assume `_pytorch_wrapper` was replaced with `Tensor` because things of type `Tensor` can be auto-casted later down the line ?

If so, I guess that makes sense, only other thing I have left is my comment regrading the auto-cast impl.

lightwood/helpers/torch.py
paxcema: Good point, will fix

PR #349, opened 2020-12-23
Multiple column dependence for time series encoders

paxcema:
## Why
Our time series encoders currently support dependence only on the previous target values. It is perfectly possible to have a time series task where another column, given as additional _historical_ context (that is, as a sequence), could prove to be very useful for the RNN or Transformer encoder to generate a more refined encoder that would yield better forecasts downstream.

## How

This is a rather complex PR where the flow in the `NnMixer` is somewhat changed. The time series encoders are now prepared at the very end, after all other encoders have been trained. This is because their encoded outputs are passed as input to the time series encoder, which will be created to handle tensors with context-rich content.

As additional points:
- some restructuring was done in the time series encoders and helpers, resulting in (hopefully) clearer code
- a fix for #350 is introduced


lightwood/api/data_source.py
```diff
@@ -310,11 +316,16 @@ def _make_column_encoder(self, encoder_class, encoder_attrs=None, is_target=Fals
         return encoder_instance
 
     def _prepare_column_encoder(self, config, is_target=False, training_data=None):
-        column_data = self.get_column_original_data(config['name'])
-        encoder_class = config.get(
-            'encoder_class',
-            self._lookup_encoder_class(config['type'], is_target)
-        )
+        column_data = self.get_column_original_data(config['name'], is_array=config.get('historical') or config.get('previous'))
```
George3d6, line 319: Ok, so, the big problem here is that this will *remove* some behavior that we previously had, name encoding the `historical*` and `previous*` columns with a timeseries encoder *and* at the same time still waste the time of having them reshaped to arrays in mindsdb_native.

While I agree side by side encoding might be good, if we have something like:

```
order_col -> [1,2,3,4,5]
prev_target -> [0,1,4,9,16]
```

It's unclear to me why we couldn't just use the array representation of `prev_target` when encoding it along side `order_col`, i.e. apply the timeseries encoder to:

```
[[1,2,3,4,5]
 [0,1,4,9,16]] 
````

It might be that we want 3 different concepts here that are being confused:

1. A `array/timeseries auto-encoder context column` that contains a single value for each row and gets feed into the RNNEncoder used for another column
2. A `historical` column (like the current `historical*` and `previous*` columns), this is a column that gets encoded as an array/timeseries, using the RNNEncoder
3. An ability for the RNNEncoder and other ts/array encoders to auto encode two arrays combined (e.g. a `historical*` and an order by column combined)

It seems to me that this PR is implementing `1.` but in an inefficient way that breaks `2.`, though maybe I am wrong.

George3d6 (changes requested):

PR #354, opened 2021-01-14
Class distribution for classification tasks 

paxcema:
## Why
For classification tasks, we need Lightwood to return the complete class distribution probabilities for certain scenarios downstream, like using XAI techniques like LIME, or running MindsDB in the OpenML AutoML benchmark suite.

## How
Added support for returning class probabilities in all mixers (SkLearn and Boost by calling `.predict_proba(X)`, NnMixer by adding non-calibrated softmax to the one-hot encoder). Predictions are now sent to native as a dictionary.

### DIFF
--- a/lightwood/api/data_source.py
+++ b/lightwood/api/data_source.py
@@ -459,8 +459,8 @@ class DataSource(Dataset):
             decoder_instance = self.encoders[column_name]
 
         decoded_data = {}
-        if hasattr(decoder_instance, 'predict_proba'):
-            # used for encoders with onehot decoding
+        if hasattr(decoder_instance, 'predict_proba') and decoder_instance.predict_proba:
+            # return complete belief distribution
             preds, pred_probs, labels = decoder_instance.decode(encoded_data)
             decoded_data['predictions'] = preds
             decoded_data['predict_proba'] = pred_probs
--- a/lightwood/api/data_source.py
+++ b/lightwood/api/data_source.py
@@ -459,11 +459,11 @@ class DataSource(Dataset):
             decoder_instance = self.encoders[column_name]
 
         decoded_data = {}
-        if hasattr(decoder_instance, 'predict_proba') and decoder_instance.predict_proba:
+        if getattr(decoder_instance, 'predict_proba', False):
             # return complete belief distribution
             preds, pred_probs, labels = decoder_instance.decode(encoded_data)
             decoded_data['predictions'] = preds
-            decoded_data['predict_proba'] = pred_probs
+            decoded_data['class_distribution'] = pred_probs
             decoded_data['class_labels'] = labels
         else:
             decoded_data['predictions'] = decoder_instance.decode(encoded_data)
--- a/lightwood/api/data_source.py
+++ b/lightwood/api/data_source.py
@@ -457,7 +457,16 @@ class DataSource(Dataset):
                     You should not decode before having encoding at least once
                     """)
             decoder_instance = self.encoders[column_name]
-        decoded_data = decoder_instance.decode(encoded_data)
+
+        decoded_data = {}
+        if hasattr(decoder_instance, 'predict_proba'):
+            # used for encoders with onehot decoding
+            preds, pred_probs, labels = decoder_instance.decode(encoded_data)
+            decoded_data['predictions'] = preds
+            decoded_data['predict_proba'] = pred_probs
+            decoded_data['class_labels'] = labels
+        else:
+            decoded_data['predictions'] = decoder_instance.decode(encoded_data)
 
         return decoded_data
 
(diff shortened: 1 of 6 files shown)


lightwood/encoders/categorical/onehot.py
```diff
@@ -60,9 +63,22 @@ def encode(self, column_data):
...
+                del(vector[UNCOMMON_TOKEN])
+                probs.append(softmax(vector).tolist())
+
+        if self.predict_proba:
+            # UNK not included in class_map nor belief distribution
+            if UNCOMMON_TOKEN != 0:
+                raise Exception("Uncommon token should be the first assigned token in the vocabulary, aborting.")
+            class_map = deepcopy(self._lang.index2word)
+            del(class_map[UNCOMMON_TOKEN])
+            return ret, probs, {k-1: v for k, v in class_map.items()}
```
George3d6, line 82: `{k-1: v for k, v in class_map.items() if k != UNCOMMON_TOKEN}`

should solve the need for a potentially expensive deepcopy (since done very often) and make the code less confuing (since that `del` is no longer needed)

lightwood/encoders/categorical/onehot.py
```diff
@@ -1,6 +1,8 @@
 import torch
-from lightwood.encoders.text.helpers.rnn_helpers import Lang
 import numpy as np
+from copy import deepcopy
+from scipy.special import softmax
+from lightwood.encoders.text.helpers.rnn_helpers import Lang
```
George3d6, line 5: Should we really reuse this here ? Isn't it better to create a helper that is inherited by `Lang` and get rid of things like the `UNCOMMON_TOKEN` deletion (which seems to be uncecessary for the purposes of this encoder) ?

lightwood/mixers/boost.py
```diff
@@ -91,6 +91,11 @@ def _predict(self, when_data_source, include_extra_data=False):
                 else:
                     predictions[target_col_name]['predictions'] = list(self.targets[target_col_name]['model'].predict(X))
 
+                    # add distribution belief if the flag was set in the target encoder
+                    if hasattr(self.encoders[target_col_name], 'predict_proba') and self.encoders[target_col_name].predict_proba:
```
George3d6, line 95: `if getattr(self.encoders[target_col_name], 'predict_proba', False) `

lightwood/mixers/sklearn_mixer.py
```diff
@@ -183,6 +183,12 @@ def _predict(self, when_data_source, include_extra_data=False):
                         self.targets[target_col_name]['model'].predict(X)
                     )
 
+
+                    # add distribution belief if the flag was set in the target encoder
+                    if hasattr(self.encoders[target_col_name], 'predict_proba') and self.encoders[target_col_name].predict_proba:
```
George3d6, line 188: if getattr(self.encoders[target_col_name], 'predict_proba', False)

lightwood/api/data_source.py
```diff
@@ -457,7 +457,16 @@ def get_decoded_column_data(self, column_name, encoded_data, decoder_instance=No
                     You should not decode before having encoding at least once
                     """)
             decoder_instance = self.encoders[column_name]
-        decoded_data = decoder_instance.decode(encoded_data)
+
+        decoded_data = {}
+        if hasattr(decoder_instance, 'predict_proba') and decoder_instance.predict_proba:
```
George3d6, line 462: if getattr(self.encoders[target_col_name], 'predict_proba', False)

lightwood/mixers/nn.py
```diff
@@ -447,6 +447,10 @@ def _predict(self, when_data_source, include_extra_data=False):
             if include_extra_data:
                 predictions[output_column]['encoded_predictions'] = output_trasnformed_vectors[output_column]
 
+            if 'predict_proba' in decoded_predictions:
+                predictions[output_column]['class_distribution'] = decoded_predictions['predict_proba']
+                predictions[output_column]['class_labels'] = decoded_predictions['class_labels']
```
George3d6, line 452: Why not keep consistent naming ? i.e call `class_distribution` here `predict_proba` or vice versa, replace `predict_proba` key with `class_distribution`

George3d6 (changes requested):

lightwood/encoders/categorical/onehot.py
paxcema: Notice that `Lang` usage isn't introduced here. I agree we could probably do this, however I'm not sure if it should be part of this PR as it may impact accuracy in the categorical AE.

George3d6: Ah, ok, my bad, will make an issue about this.

Issue #378, opened 2021-02-03
Improve NnMixer DefaultNet shape


paxcema:
Whenever a dataset has a huge amount of columns, our NnMixer should not inflate the feature size in the intermediate layers. As it stands, this leads to an incredibly large NN that takes ages to train, and does not really perform better than smaller alternatives.

We should improve the policy for determining the DefaultNet object shape, depending on the size of the input tensor.

PR #379, opened 2021-02-03
Improved DefaultNet shape policy

paxcema:
## Why
As framed in #378, there were certain cases for which the concatenated input tensor to the DefaultNet was very big, and this led to prohibitively big NnMixers. 

As an example, let's consider the `micro_mass` [dataset](https://www.openml.org/d/1514), which has some 1.3k columns. After all respective encoders have been trained, the NnMixer has to deal with (or learn from) a `(B, 6072)` shaped input tensor. The current shape policy for the NnMixer's DefaultNet would then instantiate a sequential model with dimensions `[6072, 12144, 3]` with some 220 million trainable parameters. This, naturally, led to incredibly long training times with poor performance to boot. 

## How
The PR introduces a non-destructive additional size check, where we check if the candidate shape for the NnMixer has less than `target_params`. If not, we rescale accordingly. In extreme cases (whenever the input descriptor is enormous), we omit the hidden layer as well.

## Experiments
We experimented with the `micro_mass` and `stack_overflow_survey` datasets, forcing the predictor to use an NnMixer:

- `micro_mass`: `5590 [s]` runtime to `52 [s]` runtime, with a very small decrease in balanced accuracy from `0.159` to `0.1314`.
- `stack_overflow_survey`: `501 [s]` runtime to `470 [s]` runtime, increasing r2 score from `0.5995` to `0.6663`.

It is very likely that some OpenML datasets which timed out will now run normally. This will be tried out during the week.

### DIFF
--- a/lightwood/mixers/helpers/default_net.py
+++ b/lightwood/mixers/helpers/default_net.py
@@ -1,3 +1,4 @@
+import math
 import torch
 from functools import reduce
 
@@ -16,7 +17,7 @@ class DefaultNet(torch.nn.Module):
                      output_size=None,
                      nr_outputs=None,
                      shape=None,
-                     target_params=3e5,
+                     max_params=3e5,
                      dropout=None,
                      pretrained_net=None):
         self.input_size = input_size
@@ -39,9 +40,9 @@ class DefaultNet(torch.nn.Module):
             # default shape, inflated, yields 240k params if input has 200 components
             shape = [self.input_size, max([self.input_size*2, self.output_size*2, 400]), self.output_size]
 
-            # if NN is too big, we do not inflate and aim for max target_params
-            if reduce(lambda x, y: x*y, shape) > target_params:
-                hidden_size = int(target_params//(self.input_size*self.output_size))
+            # if NN is too big, we do not inflate and aim for max_params instead
+            if reduce(lambda x, y: x*y, shape) > max_params:
+                hidden_size = math.floor(max_params/(self.input_size*self.output_size))
 
                 if hidden_size > self.output_size:
                     shape = [self.input_size, hidden_size, self.output_size]
--- a/lightwood/mixers/helpers/default_net.py
+++ b/lightwood/mixers/helpers/default_net.py
@@ -1,4 +1,5 @@
 import torch
+from functools import reduce
 
 from lightwood.config.config import CONFIG
 from lightwood.mixers.helpers.shapes import *
@@ -15,6 +16,7 @@ class DefaultNet(torch.nn.Module):
                      output_size=None,
                      nr_outputs=None,
                      shape=None,
+                     target_params=3e5,
                      dropout=None,
                      pretrained_net=None):
         self.input_size = input_size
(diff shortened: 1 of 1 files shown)


lightwood/mixers/helpers/default_net.py
```diff
@@ -15,6 +16,7 @@ def __init__(self, dynamic_parameters,
                      output_size=None,
                      nr_outputs=None,
                      shape=None,
+                     target_params=3e5,
```
George3d6, line 19: Should probably measure GPU free memory size and use that, also rename to `max params` since we aren't "targeting this" we are "targeting being bellow this"

lightwood/mixers/helpers/default_net.py
```diff
@@ -34,11 +36,21 @@ def __init__(self, dynamic_parameters,
         super(DefaultNet, self).__init__()
 
         if shape is None and pretrained_net is None:
-            shape = [self.input_size, max([self.input_size*2,self.output_size*2,400]), self.output_size]
+            # default shape, inflated, yields 240k params if input has 200 components
+            shape = [self.input_size, max([self.input_size*2, self.output_size*2, 400]), self.output_size]
+
+            # if NN is too big, we do not inflate and aim for max target_params
+            if reduce(lambda x, y: x*y, shape) > target_params:
```
George3d6, line 43: Every time I can't think of an imperative equivalent simpler than the usage of a HOF the functional programming old gods gather a bit more strength.

lightwood/mixers/helpers/default_net.py
```diff
@@ -34,11 +36,21 @@ def __init__(self, dynamic_parameters,
...
+                hidden_size = int(target_params//(self.input_size*self.output_size))
```
George3d6, line 44: I had actually forgotten this was a thing, can you replace with `math.floor` after the division? Unless you think this operator is common knowledge, might trip up people.

lightwood/mixers/helpers/default_net.py
```diff
@@ -34,11 +36,21 @@ def __init__(self, dynamic_parameters,
...
+
+                if hidden_size > self.output_size:
+                    shape = [self.input_size, hidden_size, self.output_size]
+                else:
+                    shape = [self.input_size, self.output_size]
```
George3d6, line 49: The way this scales down is too aggressive.
Instead have  `while reduce...` and every loop:

`if hidden_size > self.output_size` => `hidden_size=hidden_size*0.99`

else: `[self.input_size, self.output_size]` and break (though if the data is so big we might as well just "quit" and rely on the lightGBM mixer maybe?, to think about for later though)

George3d6 (changes requested):

lightwood/mixers/helpers/default_net.py
paxcema: sure thing

lightwood/mixers/helpers/default_net.py
paxcema: Hm, if we do it based on free GPU memory, doesn't it get tricky when we have more than one concurrent predictor? Will rename to `max_params`.

PR #380, opened 2021-02-04
Add GradScaler for AMP usage

paxcema:
## Why
With Automated Mixed Precision, sometimes gradients would underflow, resulting in NaNs being propagated.

## How
Implementing `torch.amp.GradScaler` recommended usage.

### DIFF
--- a/lightwood/mixers/nn.py
+++ b/lightwood/mixers/nn.py
@@ -634,7 +634,7 @@ class NnMixer(BaseMixer):
                     self.optimizer.step()
                     if self.is_selfaware and self.start_selfaware_training and awareness_loss is not None:
                         awareness_loss.backward(retain_graph=True)
-                        self.selfaware_optimizer.st ep()
+                        self.selfaware_optimizer.step()
 
                 error = running_loss / (i + 1)
 
--- a/lightwood/mixers/nn.py
+++ b/lightwood/mixers/nn.py
@@ -1,21 +1,22 @@
+import time
 import copy
 import random
-import time
-
-import torch
-from torch.utils.data import DataLoader
-import numpy as np
 import operator
 
+import torch
+import numpy as np
+from torch.utils.data import DataLoader
+from torch.cuda.amp import GradScaler
+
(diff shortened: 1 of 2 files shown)


lightwood/mixers/nn.py
```diff
@@ -600,10 +603,10 @@ def _iter_fit(self, ds, subset_id=None, max_epochs=120000):
                         target_loss = target_loss.tolist()
                         if type(target_loss[0]) == type([]):
                             target_loss = [np.mean(x) for x in target_loss]
-                        for i, value in enumerate(target_loss):
-                            if len(unreduced_losses) <= i:
+                        for j, value in enumerate(target_loss):
```
hakunanatasha, line 607: why the change from i to j? Is it cosmetic?

lightwood/mixers/nn.py
```diff
@@ -634,7 +634,7 @@ def _iter_fit(self, ds, subset_id=None, max_epochs=120000):
                     self.optimizer.step()
                     if self.is_selfaware and self.start_selfaware_training and awareness_loss is not None:
                         awareness_loss.backward(retain_graph=True)
-                        self.selfaware_optimizer.st ep()
+                        self.selfaware_optimizer.step()
```
hakunanatasha, line 637: shouldn't a unittest catch this? it was self.selfaware_optmizer.st ep() which seems like it should have broke something (as it looks like a syntax error) ...

hakunanatasha (approved):

lightwood/mixers/nn.py
paxcema: Just to be sure. There is an `i` variable already being used a little farther up. The closure in the for loop should mean it's not an issue, but it's probably safer to change it so that one doesn't make mistakes later on.

lightwood/mixers/nn.py
paxcema: Yup, it should fail (and it did, locally). The diff GitHub shows here is between the last two commits in the PR, so this error never made it into `stable`.

PR #382, opened 2021-02-05
Time series: normalize by group combination

paxcema:
## Why
For time series tasks where the target is grouped by certain columns/dimensions, these new sub-series can have markedly different scales of magnitude. Rescaling all series using a single normalizer can lead to a lot of signal that's lost in our encoders and mixers, meaning forecasts that are biased to the sub-series with largest absolute values.

This PR introduces a dynamic normalizing schema, where we rescale each sub-series based only on its own values, both at training and inference time.

## How
The PR involves quite a few changes:
- New `common` time series helper module, where we move all normalizers, and add custom logic to discover all possible group combinations and train each respective normalizer.
- Changes to the RNN time series encoder so that it MinMax normalizes each sub-series individually before training and encoding.
- Passing information (various flags,  normalizers themselves, etc.) between time series and numerical encoders.
- New `ts_numeric` numerical encoder, which will have any and all specialized logic for encoding and decoding numerical time series targets. For now, this considers dynamic rescaling through the use of each sub-series corresponding absolute mean.


lightwood/encoders/numeric/ts_numeric.py
```diff
@@ -0,0 +1,128 @@
...
+class TsNumericEncoder(NumericEncoder):
+    """
+    Variant of vanilla numerical encoder, supports dynamic mean re-scaling
+    """
+    def __init__(self, is_target=False):
+        super(TsNumericEncoder, self).__init__(is_target=is_target)
+        # time series normalization params
+        self.normalize_by_group = False
+        self.normalizers = None
+        self.group_combinations = None
```
hakunanatasha, line 19: what does group_combinations refer to

lightwood/mixers/nn.py
```diff
@@ -9,6 +9,7 @@
 from torch.cuda.amp import GradScaler
 
 from lightwood.logger import log
+from lightwood.encoders import *
```
hakunanatasha, line 12: might want to specify which modules/things are you importing for consistency's sake and clarity

hakunanatasha: to also add, do you need to import all of it?

hakunanatasha (commented):
It's worth @George3d6 taking a look, because I am still ignorant with how many modules interact together BUT I don't think I see anything that could break something.

Most of my comments are much more from ignorance than performance oriented. One of the things I think about is what attributes are allowed to be passed and what design plan you had (i.e. 'normalizer' or 'group_combination' examples). It gives me a better idea of what you intend a 'base case' to be, and a future functionality might be.

Would you be wiling to share your jupyter notebook, when the time is ready? I think I can piece together the types of the input/output arguments more effectively when I can see what kind of things are being passed, as some functions referenced aren't in the set of scripts here for review. 

Also  - this is a ridiculously hard problem and im very impressed at the results!

lightwood/api/data_source.py
```diff
@@ -294,11 +296,17 @@ def _lookup_encoder_class(self, column_type, is_target):
...
         }
 
+        column_type = column_config['type']
         if is_target and column_type in target_encoder_classes:
             encoder_class = target_encoder_classes[column_type]
         else:
             encoder_class = default_encoder_classes[column_type]
 
+        # specialized dispatches
+        if is_target and column_config.get('additional_info', {}).get('time_series_target', False) and \
```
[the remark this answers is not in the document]
paxcema: Indeed, we select the default encoder. In this particular dispatch though, the default would be `NumericEncoder`, for whenever your numerical target is _not_ time series related.

lightwood/api/data_source.py
```diff
@@ -345,35 +353,25 @@ def _prepare_encoders(self):
         with all available data for that column.
         """
         encoders = {}
-    
         input_encoder_training_data = {'targets': [], 'previous': []}
 
-        for config in self.config['output_features']:
```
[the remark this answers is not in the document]
paxcema: It's not deleted, I just changed the order (see line 394 in the incoming changes). I had to opt for this because we need to setup the group normalizers before preparing output encoders (if they turn out to be `TsNumericEncoders`, that info needs to exist first).

lightwood/encoders/numeric/ts_numeric.py
paxcema: Good point, I'll add a comment. `group_combinations` are all sets of grouped by columns observed at training time.

Imagine we group by 2 columns: `country` and `size`. If the possible values for these in the training corpus are, respectively, `USA, UK` and `Big, Medium, Small`, then `group_combinations` will hold all 6 possible combinations between these two sets. 

We use a `frozenset()` of the respective strings to represent these combinations, so that we can retrieve normalizer objects at inference time once we know what combination we are looking at.

lightwood/mixers/nn.py
paxcema: Fair enough, I do not. Only `NumericEncoder` is needed right now. Will fix.

lightwood/api/data_source.py
```diff
@@ -393,6 +391,40 @@ def _prepare_encoders(self):
...
+        for config in self.config['output_features']:
+            column_name = config['name']
+            column_data = self.get_column_original_data(column_name)
+
+            prev_col = [c for c in self.config['input_features'] if c['name'] == f'__mdb_ts_previous_{column_name}']
+            if prev_col:
+                # 0-indexed as we know it can only be one
+                config['depends_on_column'] = [prev_col[0]['name']]
+                config['encoder_attrs']['normalizers'] = prev_col[0]['additional_info']['normalizers']
+                config['encoder_attrs']['group_combinations'] = prev_col[0]['additional_info']['group_combinations']
```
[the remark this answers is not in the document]
paxcema: Yup, `encoder_attrs` is specified in `lightwood/data_schemas/prediction_config.py` as a `dict()`, and it should include any attribute that we 1) know a particular encoder might have and 2) want to set with some specific value. These attributes can come from Native or from within Lightwood, like in this case.

lightwood/api/data_source.py
```diff
@@ -345,35 +353,25 @@ def _prepare_encoders(self):
...
             column_name = config['name']
             if column_name.startswith('__mdb_ts_previous_'):
                 column_data = self.get_column_original_data(column_name)
                 previous_cols.append(column_name)
-                input_encoder_training_data['previous'].append({'data': column_data,
-                                                                'name': column_name,
-                                                                'original_type': config['original_type'],
-                                                                'output_type': config['type']
-                                                                })
+                col_info = {'data': column_data,
```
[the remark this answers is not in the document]
paxcema: We might, not sure though as this is the only place it can be setup from. Thoughts @George3d6?

lightwood/api/data_source.py
```diff
@@ -393,6 +391,40 @@ def _prepare_encoders(self):
...
+            column_name = config['name']
+            column_data = self.get_column_original_data(column_name)
+
+            prev_col = [c for c in self.config['input_features'] if c['name'] == f'__mdb_ts_previous_{column_name}']
+            if prev_col:
+                # 0-indexed as we know it can only be one
+                config['depends_on_column'] = [prev_col[0]['name']]
+                config['encoder_attrs']['normalizers'] = prev_col[0]['additional_info']['normalizers']
+                config['encoder_attrs']['group_combinations'] = prev_col[0]['additional_info']['group_combinations']
+                config['encoder_attrs']['normalize_by_group'] = True
```
George3d6, line 404: Why not default this to `False`, as it stands right now it's only turned on for autorgressive timeseries it seems... which doesn't really make sense to me, seems it would work just as well for all timeseries.

lightwood/api/data_source.py
```diff
@@ -393,6 +391,40 @@ def _prepare_encoders(self):
...
+                group_info = {col['name']: self.get_column_original_data(col['name'])
+                              for col in self.config['input_features'] if col['grouped_by']}
```
George3d6, line 406: Generate the group_info once at datasource creation (I think) but minor improvement, just makes it more redable.

lightwood/api/data_source.py
```diff
@@ -468,6 +504,10 @@ def get_decoded_column_data(self, column_name, encoded_data, decoder_instance=No
             decoded_data['predictions'] = preds
             decoded_data['class_distribution'] = pred_probs
             decoded_data['class_labels'] = labels
+        elif getattr(decoder_instance, 'normalize_by_group', False):
```
George3d6, line 507: Just check the class type and always pass it, the getattr thing makes it kinda too arcane. Even if we keep `normalize_by_group` we can still pass the group info all the time.

lightwood/encoders/time_series/helpers/common.py
```diff
@@ -0,0 +1,120 @@
...
+        for val, key in zip(combination, keys):
+            all_sets.append(set([i for i, elt in enumerate(data['group_info'][key]) if elt == val]))
+        if all_sets:
+            idxs = list(set.intersection(*all_sets))
+            return idxs, np.array(data['data'])[idxs, :]
+        else:
+            return [], np.array([])
+
+
+def generate_target_group_normalizers(data):
```
George3d6, line 85: very easy to just return the normalizer, also, it seems like `group_combinations` should be created outside of this function (though than you'd have to add some special logic here around handling the `__default` group so that might make it uglier.

lightwood/api/data_source.py
```diff
@@ -345,35 +353,25 @@ def _prepare_encoders(self):
...
-                                                                'output_type': config['type']
-                                                                })
+                col_info = {'data': column_data,
+                            'name': column_name,
+                            'original_type': config['original_type'],
+                            'group_info': {conf['name']: self.get_column_original_data(conf['name'])
+                                           for conf in self.config['input_features'] if conf['grouped_by']},
+                            'output_type': config['type']
+                            }
+                col_info = generate_target_group_normalizers(col_info)
```
George3d6, line 371: Just to be clear on this:
* It will still get triggered when no group bys exist
* It will just generate the `__default` normalizer which will be used for everything

Or is the no-group-by situation handled some other way ?

George3d6 (changes requested):

lightwood/api/data_source.py
paxcema: Turns out this flag is a leftover from when there was no `TsNumericEncoder`, removing

lightwood/api/data_source.py
paxcema: That's indeed what happens, correct

Issue #327, opened 2020-11-30
Wrong assumption in weight slices for multi-categorical tasks


paxcema:
The `weight_slice` construction code (in `api/data_source.py`) assumes one-hot encoding for the encoded value:

```
...
encoded_val = self.get_encoded_column_data(col_config['name'], custom_data={col_config['name']: [val]})

# @Note: This assumes one-hot encoding for the encoded_value
value_index = np.argmax(encoded_val[0])

if new_weights is None:
    new_weights = [np.mean(list(weights.values()))] * len(encoded_val[0])

new_weights[value_index] = weights[val]
...
```

In multi-categorical tasks this assumption is not valid, which means the weight_slice construction is incorrect. We should keep the current logic for the simple categorical case, and add custom logic for the multi-categorical case.

PR #385, opened 2021-02-11
Create weights for multiclass

maximlopin:
Closes https://github.com/mindsdb/lightwood/issues/327

### DIFF
--- a/lightwood/api/data_source.py
+++ b/lightwood/api/data_source.py
@@ -261,7 +261,7 @@ class DataSource(Dataset):
 
                         for value_index, hot in enumerate(encoded_val[0]):
                             if hot:
-                                new_weights[value_index] = weights[val] / sum(encoded_val[0])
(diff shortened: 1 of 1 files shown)


lightwood/api/data_source.py
```diff
@@ -238,14 +238,19 @@ def __getitem__(self, idx):
...
 
                         if new_weights is None:
-                            new_weights = [np.mean(list(weights.values()))] * len(encoded_val[0])
+                            mean = np.mean(list(weights.values()))
+                            new_weights = [mean] * len(encoded_val[0])
 
-                        new_weights[value_index] = weights[val]
+                        for value_index, hot in enumerate(encoded_val[0]):
+                            if hot:
+                                new_weights[value_index] = weights[val] / sum(encoded_val[0])
```
paxcema, line 253: `sum(encoded_val[0]).item()` so that all `new_weights` values are float, and not tensors

paxcema (changes requested):

PR #387, opened 2021-02-11
checking for and using gpu

George3d6:
Checking if the GPU version of lightgbm is installed, kind of a hacky way to do it, asked here for a better variant: https://github.com/microsoft/LightGBM/issues/3939 ... but it's good enough.

If it's installed, use that + tweak a parameter based on the official docs recommendation (though maybe designed for older GPUs ... ? Should we bin numbers at all on GPUs ? Can lightgbm not bin numbers and determine values to split at on a continous axis? Doesn't seem to be the case but idk)


George3d6:
Will add a way to install lightgbm with GPU in the new mindsdb dependency install endpoint.

paxcema (approved):

PR #391, opened 2021-02-12
Plain time series encoder

paxcema:
## Why
For improved time series forecasting performance.

## How
This simple encoder fits a normalizer for any given time series target column. When training the mixer, we now have access to a normalized array of `window` previous values to predict a data point. 

We could further improve this encoder using grouped-by normalizers, but this simple variant already shows improvements in benchmarks, so I'm PRing as it is and we can implement this further down the line.


lightwood/encoders/time_series/plain.py
```diff
@@ -0,0 +1,34 @@
...
+            else:
+                self._normalizer = MinMaxNormalizer()
+            self._normalizer.prepare(priming_data)
+        self._prepared = True
+
+    def encode(self, column_data):
+        if not self._prepared:
+            raise Exception('You need to call "prepare" before calling "encode" or "decode".')
+        data = torch.cat([self._normalizer.encode(column_data)], dim=-1)
+        data[torch.isnan(data)] = 0.0
```
George3d6, line 33: Probably not needed, but just in case you forgot about that bit, this will not handle `-inf/inf`

lightwood/encoders/time_series/plain.py
```diff
@@ -0,0 +1,34 @@
...
+    def prepare(self, priming_data):
+        if self._prepared:
+            raise Exception('You can only call "prepare" once for a given encoder.')
+        else:
+            if self.original_type == COLUMN_DATA_TYPES.CATEGORICAL:
+                self._normalizer = CatNormalizer()
+            else:
+                self._normalizer = MinMaxNormalizer()
+            self._normalizer.prepare(priming_data)
+        self._prepared = True
```
George3d6, line 27: just a coding style note (I thik it's in the guidelines)

when you have `if something -> return/raise` and the other branch is a long list of lines, don't but it in an `else` (I'm not inventing this, clang, pytorch and others this gudieline for readability and for python it's even more importnat due to indentation)

lightwood/mixers/helpers/transformer.py
```diff
@@ -14,10 +14,12 @@ def transform(self, sample):
...
-            sub_vector = sample['input_features'][input_feature].tolist()
-            input_vector.extend(sub_vector)
-            if input_feature not in self.feature_len_map:
-                self.feature_len_map[input_feature] = len(sub_vector)
+            # omit PlainTsEncoder, its information is encoded in the normal TsEncoder already
+            if '__mdb_ts_previous' not in input_feature:
+                sub_vector = sample['input_features'][input_feature].tolist()
+                input_vector.extend(sub_vector)
+                if input_feature not in self.feature_len_map:
+                    self.feature_len_map[input_feature] = len(sub_vector)
```
George3d6, line 22: Ok, I'm not sure I get this, will ping in slack

George3d6 (changes requested):

George3d6 (changes requested):
Also, please avoid renaming whenever possible, `TimeSeriesPlainEncoder` shouldn't become `TsPlainEncoder`

lightwood/encoders/time_series/plain.py
paxcema: good point, will consider this for good measure

lightwood/encoders/time_series/plain.py
paxcema: Duly noted

Issue #393, opened 2021-02-15
Datasource does not process text properly


hakunanatasha:
* Python version: 3.8.5
* Operating system: Ubuntu [20.04]
* Lightwood version: 0.60.0

Text no longer encodes properly; the issue arises in the creation of the DataSource. It seems like there is some information not passed along correctly to the distilbert encoder that isn't a problem for numeric/categorical. An easy dataset to check this out with is the wine_reviews dataset in the private benchmarks. 

Additionally - this was fine in 0.57.0

![Screenshot from 2021-02-15 15-05-27](https://user-images.githubusercontent.com/34254572/107989047-5d41ab00-6f9f-11eb-9127-075ad8e0be54.png)

![Screenshot from 2021-02-15 15-16-16](https://user-images.githubusercontent.com/34254572/107989850-1eacf000-6fa1-11eb-899f-0d1cde5641a0.png)

PR #394, opened 2021-02-15
Unittest text

hakunanatasha:
With regards to: [Issue 393](https://github.com/mindsdb/lightwood/issues/393)

The recent change to data_source.py influences the text encoder. This simple additional test checks to see if the text encoder is properly called by ensuring the config file tells lightwood that the input column is a text-type.

You can check this behavior between stable (0.60.0) and fix_l393
git checkout stable
python -m unittest test_predictor.TestPredictor.test_text_encoder

Versus
git checkout fix_l393
python -m unittest test_predictor.TestPredictor.test_text_encoder

The fix_l393 should be merged into stable in addition.

### DIFF
--- a/tests/unit_tests/test_predictor.py
+++ b/tests/unit_tests/test_predictor.py
@@ -97,7 +97,7 @@ class TestPredictor(unittest.TestCase):
             "output_features": [
                 {"name": "output", "type": "numeric"},
             ],
-            "mixer": {"class": BoostMixer},
+            "mixer": {"class": LightGBMMixer},
         }
 
         inpx = ["this is a test case of text", "lightwood is awesome!"] *100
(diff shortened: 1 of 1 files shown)


tests/unit_tests/test_predictor.py
```diff
@@ -82,3 +82,27 @@ def test_learn_and_predict_sklearn_mixer(self):
...
+        Config ensures "input" is treated as text.
+        """
+        config = {
+            "input_features": [
+                {"name": "input", "type": "text"},
+            ],
+            "output_features": [
+                {"name": "output", "type": "numeric"},
+            ],
+            "mixer": {"class": BoostMixer},
```
paxcema, line 100: Ideally use `LightGBM` or `NnMixer` here, as `BoostMixer` might be deprecated in the near future

PR #399, opened 2021-02-19
Round robin GPUs

George3d6:
Works via an env flag


lightwood/helpers/device.py
```diff
@@ -6,14 +7,27 @@
...
+            else:
+                round_robin = False
+
+            if round_robin:
+                current_gpu = int(os.environ.get('ROUND_ROBIN_GPU_CURRENT', '0'))
+                device_str = 'cuda:' + str(current_gpu % available_devices)
+                os.envrion['ROUND_ROBIN_GPU_CURRENT'] = (str(current_gpu+1))
     else:
         device_str = "cpu"
+
```
paxcema, line 25: missing an `available_devices` default value, probably should go inside this else

PR #402, opened 2021-02-25
Textenc

hakunanatasha:
Creates the 'PretrainedLang' class, that uses deploys hugging face models to directly tokenize, and encode text data into embeddings.

It uses distilbert from huggingface; for classification problems, it defaults to fine-tuning for 1 epoch with num_classes+1 categories in prediction (based off earlier work with distilbert.py). For regression (or multiple targets), it defaults to general embeddings from the base transformer. Fine-tuning is dependent on the dataset, but generally is as fast, if not faster than current implementation with distilbert.

The major changes, outside of the encoder itself, is calls in DataSource that actively deploy the text encoder. 

I omitted other language models for simplicity; I'm happy to explore those options, but for now, we will default to distilbert. GPT-2 might be a reasonable follow up. 

I have print statements included at various points in the encoder; I can omit these, but it's a nice flag for the training procedure (on my end). This can also be relegated to the logger-  just let me know.


lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,362 @@
...
+
+Pre-trained embedding model.
+
+This deploys a hugging face transformer
+and trains for a few epochs onto the target.
+
+Once this has been completed, it provides embedding
+using the updated transformer embedding.
+
+NOTE - GPT2 does NOT have a padding token!!
```
paxcema, line 22: Is this handled in the PR? If the `lightwood.config` specifies a GPT2 text encoder, what happens to sentences that need to be padded to construct the tensor?

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,362 @@
...
+    get_linear_schedule_with_warmup,
+)
+
+
+class PretrainedLang(BaseEncoder):
+    """
+    Pretrained language models.
+    Option to train on a target encoding of choice.
+
+    The "sent_embedder" parameter refers to a function to make
```
paxcema, line 63: Potentially outdated bit of info. here

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,362 @@
...
+    custom_train ::Bool; whether to train text on target or not.
+    frozen ::Bool; whether to freeze tranformer and train a linear layer head
+    epochs ::int; number of epochs to train model with
+    """
+
+    def __init__(
+        self,
+        is_target=False,
+        model_name="distilbert",
+        desired_error=0.01,
```
paxcema, line 83: `desired_error` seems unused

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,362 @@
...
+        frozen=True,
+        epochs=3,
+    ):
+        super().__init__(is_target)
+
+        self.name = model_name + " text encoder"
+        print(self.name)
+
+        # Token/sequence treatment
+        self._pad_id = None
```
paxcema, line 97: `self._pad_id` seems unused

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,362 @@
...
+            # Prepare the priming data inputs with attention masks etc.
+            text = self._tokenizer(priming_data, truncation=True, padding=True)
+
+            # To train in the space, use labels as argmax.
+            labels = training_data["targets"][0]["encoded_output"].argmax(
+                dim=1
+            )  # Nbatch x N_classes
+            xinp = TextEmbed(text, labels)
+
+            # Pad the text tokens on the left (if padding allowed)
```
paxcema, line 175: Is this comment accurate?

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,362 @@
...
+        n_epochs - number of epochs to train
+
+        """
+        self._model.train()
+
+        if optim is None:
+            print("Setting all model params to AdamW")
+            optim = AdamW(self._model.parameters(), lr=5e-5)
+
+        for epoch in range(n_epochs):
```
paxcema, line 273: It's not strictly needed, but we could try to use `LightwoodAutocast` (unless HuggingFace automatically handles it internally). You can find an example [here](https://github.com/mindsdb/lightwood/blob/a8e5f7678a6aef0ea233a35f30a58107a0913484/lightwood/encoders/time_series/rnn.py#L174).

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,362 @@
...
+
+                encoded_representation.append(output.detach())
+
+        return torch.stack(encoded_representation).squeeze(1)
+
+    def decode(self, encoded_values_tensor, max_length=100):
+        raise Exception("Decoder not implemented yet.")
+
+    # @staticmethod
+    # def _mean_norm(xinp, dim=1):
```
paxcema, line 342: I'm curious, is this `_mean_norm` method used in the benchmarks you've made so far? How has it performed compared to the disaggregated embeddings?

paxcema (changes requested):

lightwood/encoders/text/pretrained.py
hakunanatasha: I'm going to opt to stick with distilbert since its at least currently faster than the old implementation; I need to look into GPT-2 specifically

lightwood/encoders/text/pretrained.py
hakunanatasha: it is; fixed

lightwood/encoders/text/pretrained.py
hakunanatasha: removed - holdover from older code

lightwood/encoders/text/pretrained.py
hakunanatasha: also holdover from earlier code - removed

lightwood/encoders/text/pretrained.py
hakunanatasha: No - updated to reflect what this actually does; a comment like this should have been where the tokenizer call is. Will address.

lightwood/encoders/text/pretrained.py
hakunanatasha: I actually swapped to CLS. Clustering of the the embedded space and a quick inspection of the pretrained only with mean-norm + CLS suggest there is little difference. To follow the fine-tuning prescription, I used the CLS token as the input to the pre-classifier layer

lightwood/api/data_source.py
```diff
@@ -18,13 +18,14 @@
     TsNumericEncoder,
     CategoricalAutoEncoder,
     MultiHotEncoder,
-    DistilBertEncoder,
+    #DistilBertEncoder,
```
paxcema, line 21: Let's remove the `lightwood/encoders/text/distilbert.py` file

lightwood/api/data_source.py
```diff
@@ -297,7 +298,8 @@ def _lookup_encoder_class(self, column_config, is_target):
             ColumnDataTypes.MULTIPLE_CATEGORICAL: MultiHotEncoder,
             ColumnDataTypes.DATETIME: DatetimeEncoder,
             ColumnDataTypes.IMAGE: Img2VecEncoder,
-            ColumnDataTypes.TEXT: DistilBertEncoder,
+            # ColumnDataTypes.TEXT: DistilBertEncoder,
```
paxcema, line 301: Remove this comment after deleting `distilbert.py`

lightwood/encoders/__init__.py
```diff
@@ -35,6 +36,7 @@ class Text:
     ShortTextEncoder = ShortTextEncoder
```
paxcema, line 36: Also here

lightwood/encoders/text/__init__.py
```diff
@@ -1,3 +1,4 @@
+from lightwood.encoders.text.pretrained import PretrainedLang
 from lightwood.encoders.text.rnn import RnnEncoder
 from lightwood.encoders.text.distilbert import DistilBertEncoder
```
paxcema, line 3: Also here

lightwood/encoders/__init__.py
hakunanatasha: short text encoder stays; it's meant for very short-form text i thought

hakunanatasha: or rather it's used in "SHORT_TEXT" column data type

lightwood/encoders/text/helpers/pretrained_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+
+            total_loss += loss.item()
+            
+            loss.backward()
+            optim.step()
+
+            if scheduler is not None:
+                scheduler.step()
+
+        print("Epoch", epoch+1, "Loss", total_loss)
```
paxcema, line 67: I think it's better to do `log.info` calls instead of prints, considering we have easy access to the object within the method

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+        batch_size=10,
+        max_position_embeddings=None,
+        custom_train=True,
+        frozen=False,
+        epochs=1,
+    ):
+        super().__init__(is_target)
+
+        self.name = model_name + " text encoder"
+        print(self.name, flush=True)
```
paxcema, line 85: Same as above (`log.info`)

lightwood/encoders/__init__.py
paxcema: Ah, sorry, this is Github acting weird

paxcema: I meant removing DistilBert

paxcema: Or maybe I misread, idk

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+        column_data:: [list[str]] list of text data in str form
+
+        Returns:
+        encoded_representation:: [torch.Tensor] N_sentences x Nembed_dim
+        """
+        if self._prepared is False:
+            raise Exception("You need to first prepare the encoder.")
+
+        # Freeze training mode while encoding
+        self._model.eval()
```
George3d6, line 295: If this takes time it's better to have frozen as default and unfreeze after training, since `encode` could be called on a row-by-row basis.

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+            print("Embeddings Generator only", flush=True)
+
+            self.model_type = "embeddings_generator"
+            self._model = self._embeddings_model_class.from_pretrained(
+                self._pretrained_model_name
+            ).to(self.device)
+
+        self._prepared = True
+
+    def _train_model(self, dataset, optim=None, scheduler=None, n_epochs=1):
```
George3d6, line 226: Please avoid superfolous arguments. If they aren't ever used, don't put them, if they are, don't have a default value.
We should limit arguments with default values to interfaces which are used in a lot of places, this is only used once, so it makes on sense to have the defaults

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+
+            else:
+                print("\tFine-tuning model", flush=True)
+                """
+                Fine-tuning parameters with weight decay
+                """
+                no_decay = [
+                    "bias",
+                    "LayerNorm.weight",
+                ]  # decay on all terms EXCLUDING bias/layernorms
```
George3d6, line 184: Maybe a small comment explaining why? (Unless it's just heuristic for benchmarking)

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+        else:
+            print("Optimizer provided", flush=True)
+
+        if scheduler is None:
+            print("No scheduler provided.", flush=True)
+        else:
+            print("Scheduler provided.", flush=True)
+
+        print("Beginning training.")
+        for epoch in range(n_epochs):
```
George3d6, line 252: Maybe make it more obvious that this is a single-shot/few-shot approach, and that a validation dataset (or at least loss-based) stopping condition should be applied if people want to train for longer (e.g. via a comment or via the naming of the method)

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+        n_epochs - number of epochs to train
+
+        """
+        self._model.train()
+
+        if optim is None:
+            print("No opt. provided, setting all params with AdamW.", flush=True)
+            optim = AdamW(self._model.parameters(), lr=5e-5)
+        else:
+            print("Optimizer provided", flush=True)
```
George3d6, line 244: Don't use `print`, use `log`.

Also flush=True helps with debugging but please don't use it in prod code (but should be irrelevant, since again, this should be using `log.info` or `log.debug`.

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+                        "weight_decay": 0.0,
+                    },
+                ]
+
+            optimizer = AdamW(optimizer_grouped_parameters, lr=1e-5)
+            scheduler = get_linear_schedule_with_warmup(
+                optimizer,
+                num_warmup_steps=0,  # default value for GLUE
+                num_training_steps=len(dataset) * self._epochs,
+            )
```
George3d6, line 209: No point in having a schedule for a single/few-shot approach though, right?

lightwood/encoders/text/helpers/pretrained_helpers.py
George3d6, line 67: Is this function ever used? Seems to be a remnant from experimenting.

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+            raise Exception("You need to first prepare the encoder.")
+
+        # Freeze training mode while encoding
+        self._model.eval()
+
+        encoded_representation = []
+
+        with torch.no_grad():
+            # Set the weights; this is GPT-2
+            for text in column_data:
```
George3d6, line 301: Batching here, even if in pairs of 5 or 10 due to size, might actually be *very* useful, especially since with text datasets we might not afford to store the embeddings in full (due to size, i.e. we'll have to disable the cache or it will get auto-disabled) leading to this function being called very often.

But I'm like 40% sure this would be useful, if it's too annoying don't do it.

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+
+                # Omit NaNs
+                if text == None:
+                    text = ""
+
```
George3d6, line 306: Isn't there an "empty" token we should be using here instead?

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+                # Tokenize the text with the built-in tokenizer.
+                inp = self._tokenizer.encode(
+                    text, truncation=True, return_tensors="pt"
+                ).to(self.device)
+
+                output = self._model.base_model(inp).last_hidden_state[:, 0]
+
+                # If the model has a pre-classifier layer, use this embedding.
+                if hasattr(self._model, "pre_classifier"):
+                    output = self._model.pre_classifier(output)
```
George3d6, line 316: Wait... wasn't the idea to concat the classifier layer output with the embedding? This seems to be replacing it.

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+        # Replaces empty strings with ''
+        priming_data = [x if x is not None else "" for x in priming_data]
+
+        # Checks training data details
+        # TODO: The dict should be streamlined
+        output_avail = training_data is not None and len(training_data["targets"]) == 1
+
+        output_type = training_data["targets"][0]["output_type"]
+
+        type_flag = output_type == COLUMN_DATA_TYPES.CATEGORICAL
```
George3d6, line 131: Is adding training for regression (numeric output) a todo?

Also this checked can be added to the `if` directly, seems to be a remannt for debugging.

lightwood/encoders/text/pretrained.py
```diff
@@ -0,0 +1,323 @@
...
+            )
+
+        # Replaces empty strings with ''
+        priming_data = [x if x is not None else "" for x in priming_data]
+
+        # Checks training data details
+        # TODO: The dict should be streamlined
+        output_avail = training_data is not None and len(training_data["targets"]) == 1
+
+        output_type = training_data["targets"][0]["output_type"]
```
George3d6, line 129: I know that making this work with multiple-target is really hard, but at the moment this will break the whole thing and is very bad.

Please check if the output length is == 1 and then do the training, otherwise use the raw embeddings. As it stands this risks breaking all multi-target models in weird ways.

George3d6 (changes requested):
Overall looks good, some of the changes suggested seem critical to me but maybe I'm misreading the code, reviewed kind of in a hurry, sorry :(

lightwood/encoders/text/pretrained.py
hakunanatasha: The comment isn't exactly accurate; the frozen base transformer model underperforms the tuning approach, which is why the default is set to tuned

lightwood/encoders/text/pretrained.py
hakunanatasha: Adding it to log

lightwood/encoders/text/pretrained.py
hakunanatasha: Can do- tokenizing for sure can be done in batching.

lightwood/encoders/text/pretrained.py
hakunanatasha: I followed the distilbert defaults from the prior iteration - i can look into the 'empty' token instead

lightwood/encoders/text/pretrained.py
hakunanatasha: The benchmarks I've done used explicitly this layer's output; the pre-classifier layer is just a linear layer, hence the output is actually Nbatch x Nembedding in dim (it's a 1:1 from the transformer output embedding size but only uses the CLS token). You can think of the pre-classifier layer like a "head" that modifies the output of the transformer. In theory, one linear layer that directly maps into the output probably is ok, but the original idea is that this intermediate layer could be tweaked into a lower dimensional representation instead of being 1:1. 

If we concat the output of the embedding to the transformer, we would get a vector thats 768*2 (768 - embedding dim); given the tuning approach changes the output of transformer params, I'm not sure why the embedding itself is needed. Maybe you're thinking of the prior approach of which the output was just the target prediction?

lightwood/encoders/text/pretrained.py
hakunanatasha: Training for regression is a TODO as it's going to require a bit of finesse to be appropriate; will add to the `if` directly.

lightwood/encoders/text/pretrained.py
hakunanatasha: ok, can add

hakunanatasha: doesn't "len(training_data["targets"]) == 1" take care of it?

hakunanatasha: oh never mind - i understand your point

lightwood/encoders/text/pretrained.py
hakunanatasha: This is sort of the 'prescribed recipe' for fine-tuning; the goal though is to dampen the weight updates so they don't overfit dramatically to the new dataset while tuning. I'm not particularly sure why biases/layer-norms are omitted, though

hakunanatasha (commented):
The big change I added was the log.info statements instead of print; this works as intended.

Some asks:
(1) Scheduler dynamics  - I'm opting to leave it as is, as the current approach for fine-tuning recommends a scheduler. I can run some experiments over the weekend to see if removing the scheduler makes no difference, but I do think having some way to adjust the learning rate is important given how diverse the different data types we may have access too (as indicated by some of our internal benchmarks).

(2) Batching in the encode statement - internal testing shows me that batching the command actually takes longer to process and I run into weird CUDA issues (on AWS); this may be on my end and we can discuss this, but I'm electing to leave it as is, since the current implementation is fast *enough*

(3) empty token; as far as I can tel there is no empty token for these models. The current implementation encodes a blank CLS + SEP for every " " and then a nondescript embedding which would be the same for any missing text, functioning the same as an "empty" embedding.

hakunanatasha (commented):
Added changes

PR #414, opened 2021-03-31
Autoregressive stream in NnMixer for time series

paxcema:
Adds an `AR_Net`, used in time series tasks to get better forecasts. The idea behind this approach is to incorporate last `window` values directly into an AR linear layer that outputs a linear combination of them. 

~~Benchmarks suggest minor improvements in forecasting error and drift, but these are still ongoing.~~
EDIT: Benchmarks actually show a much improved capability of adjusting to new temporal dynamics


lightwood/encoders/numeric/ts_numeric.py
```diff
@@ -37,7 +37,7 @@ def encode(self, data, extra_data=None):
                 except:
                     real = None
             if self.is_target:
-                vector = [0] * 3
+                vector = [0] * 2
```
hakunanatasha, line 40: Why did we change the list size from 3 to 2?

lightwood/encoders/numeric/ts_numeric.py
```diff
@@ -37,7 +37,7 @@ def encode(self, data, extra_data=None):
...
                 if group is not None and self.normalizers is not None:
                     try:
```
hakunanatasha, line 42: this is probably unnecessary but why do we have no catch statement here? Do we have any weird edge cases? What happens if the try fails?

lightwood/encoders/numeric/ts_numeric.py
```diff
@@ -48,23 +48,18 @@ def encode(self, data, extra_data=None):
                     mean = self._abs_mean
                 if real is not None and mean > 0:
                     vector[0] = 1 if real < 0 and not self.positive_domain else 0
-                    vector[1] = math.log(abs(real)) if abs(real) > 0 else -20
```
hakunanatasha, line 51: why is this statement no longer necessary? it may be obvious later in the code but not atm

lightwood/encoders/numeric/ts_numeric.py
```diff
@@ -48,23 +48,18 @@ def encode(self, data, extra_data=None):
...
-                    vector[2] = real / mean
+                    vector[1] = real / mean
                 else:
                     log.debug(f'Can\'t encode target value: {real}')
 
             else:
-                vector = [0] * 4
```
hakunanatasha, line 57: im very confused why the dimensionality changes? My best guess is that numeric has 3 terms and time series may have 2?

lightwood/encoders/numeric/ts_numeric.py
```diff
@@ -48,23 +48,18 @@ def encode(self, data, extra_data=None):
...
+                vector = [0] * 3
                 try:
-                    if real is None:
-                        vector[0] = 0
-                    else:
+                    if real is not None:
```
hakunanatasha, line 58: does real never have an opportunity to be None in the new framework?

lightwood/mixers/helpers/ar_net.py
```diff
@@ -0,0 +1,62 @@
...
+                         pretrained_net=pretrained_net
+                         )
+        target = transformer.output_features[0]
+        self.transformer = transformer
+        self.ar_column = f'__mdb_ts_previous_{target}'
+        self.ar_idxs = list(*[transformer.input_idxs[col]
+                              for col in transformer.input_idxs
+                              if col == self.ar_column])
+
+        dims = [(len(self.ar_idxs), transformer.feature_len_map[target])]
```
hakunanatasha, line 43: what is the dimensionality of dims?

lightwood/mixers/helpers/ar_net.py
```diff
@@ -0,0 +1,62 @@
...
+        linears = [nn.Linear(in_features=inf, out_features=outf) for inf, outf in dims]
```
hakunanatasha, line 44: this is an aside to compliment you - i like this as opposed to the deepcopy approach where I think weird things happen like all the weights get connected in a dumb way if you're not careful

lightwood/mixers/helpers/ar_net.py
```diff
@@ -0,0 +1,62 @@
...
+        return super().to(device, available_devices)
+
+    def forward(self, input):
+        with LightwoodAutocast():
+            residual_output = self._foward_net(input)
+            ar_output = self.ar_net(input[:, self.ar_idxs])
+            # force unit root
+            if self.ar_net.training:
+                self.ar_net._modules['0'].weight = nn.Parameter(torch.clamp(self.ar_net._modules['0'].weight,
+                                                                            0.0,
```
hakunanatasha, line 59: do these necessarily need to be 0/0.99 or do you want to give more flexibility to the clamping dimensions?

lightwood/mixers/helpers/ar_net.py
```diff
@@ -0,0 +1,62 @@
...
+        self.ar_net.to(self.device)
+
+    def to(self, device=None, available_devices=None):
+        self.ar_net.to(device)
+        return super().to(device, available_devices)
+
+    def forward(self, input):
+        with LightwoodAutocast():
+            residual_output = self._foward_net(input)
+            ar_output = self.ar_net(input[:, self.ar_idxs])
```
hakunanatasha, line 55: This is a cool approach - how do the weight updates work here? what I'm seeing is currently:

output1 = net1(inp)
output2 = net2(inp)

but the loss would combine output1 + output2 (as the forward pass sums the two terms up). 

I'm curious to know what happens when dynamics of net1/net2 differ (i.e. output1 >>> output2 or vice versa).

lightwood/mixers/helpers/transformer.py
```diff
@@ -7,17 +7,21 @@ def __init__(self, input_features, output_features):
         self.output_features = output_features
 
         self.feature_len_map = {}
+        self.input_idxs = {}
```
hakunanatasha, line 10: why is input_idxs required now?

lightwood/mixers/nn.py
```diff
@@ -202,6 +219,9 @@ def _fit(self, train_ds, test_ds, stop_training_after_seconds=None):
         if self.optimizer_args is None:
             self.optimizer_args = {'lr': 0.0005}
 
+        if is_ts_target:
+            self.optimizer_args['weight_decay'] = 2e-2
```
hakunanatasha, line 223: I've seen weight decay at 2e-5 to 1e-5; I trust it works but any reason its 2e-2?

hakunanatasha (commented):
There is a handful of things that I don't understand but I don't see an explicit "game-breaking" bugs. The change in vector dimensionality is a potential cause for annoyance later and may be worth documenting what the terms of the input are.

Also maybe I'm dumb and the ARnet is a vanilla implementation that everyone does in this fashion BUT I see a lot of cool things we can experiment with!

lightwood/encoders/numeric/ts_numeric.py
paxcema: It's something that should have been done before: for TS numerical tasks the log target value (last element in the encoded vector) is not used, and leaving it as part of the vector unnecessarily complicates the optimization procedure.

lightwood/encoders/numeric/ts_numeric.py
paxcema: Same as above: it was not being used anyway, so I removed it

lightwood/mixers/helpers/ar_net.py
paxcema: They do need to be clamped between 0.0 and 0.99 if we want this stream of the ARNet to learn coefficients for a stationary AR process. This, from my limited experimenting, is much more stable than the unclamped version, though I didn't really focus on non-stationary series.

lightwood/mixers/helpers/transformer.py
paxcema: So that the ARNet knows which indexes of the input belong to the (normalized) previous values of the time series

lightwood/mixers/nn.py
paxcema: Purely empirical. I started _very_ aggresively and settled for this value, though I didn't test with smaller values, maybe I should?

lightwood/encoders/numeric/ts_numeric.py
paxcema: If the try fails we use the standard normalizer, which considers all training data irrespective of the group to which rows belong. Looking at this, I'm fairly certain there should be no weird uncovered edge cases.

lightwood/encoders/numeric/ts_numeric.py
paxcema: Same as above: the log term was not being used anyway, so I removed it. In this case (non-target column) the first element is there following the standard numerical encoder, as far as I can remember it's there as a sanity check (i.e. it should always decode close to 1, else something is wrong and we decode `None`).

lightwood/encoders/numeric/ts_numeric.py
paxcema: Good point. It can be `None`. In that case the encoded vector stays zero-filled and when decoding we see its first element is `< 0.5` and we decode `None` back.

lightwood/mixers/helpers/ar_net.py
paxcema: Should be `(window_size, hidden_size)`, where `window_size` is the # of previous rows the predictor can use to forecast and `hidden_size` the length of the IR vector (so, 2 if target, 3 if not)

lightwood/mixers/helpers/ar_net.py
hakunanatasha: They need to be clamped between 0/0.99 but what happens if you do something weird like 0.5/0.5?

lightwood/mixers/nn.py
hakunanatasha: I'm ok with leaving it as 2e-2 but it may be worth allowing the user to specify it if it's not declared. I wonder if different data sets would do better with different LR but these are the magic numbers of optimization that as far as I know are all empirical anyway

lightwood/mixers/helpers/ar_net.py
paxcema: I'll admit this is an analysis I did not make. In theory outputs should be in the same ballpark as they're both regularized (in part why I chose such a big decay). The original inspiration for the AR stream in this network is based off two papers:

1) https://arxiv.org/abs/1911.12436 ; for the actual layer definition
2) https://arxiv.org/pdf/1703.07015.pdf ; for the combined final output + loss propagation over two streams. See section 4.6 for some interesting analysis, though it doesn't exactly address your question. I guess we could further explore this at some point!

lightwood/mixers/helpers/ar_net.py
paxcema: Hmm, likely you are forcing a much reduced hypothesis space for the learned AR process. No idea how that would impact things, to be honest.

PR #415, opened 2021-04-01
Fixes for self-aware AMP usage

paxcema:
- Fairly minor fix for correctly using AMP with our self-aware neural network in the NnMixer


lightwood/helpers/torch.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, enabled=True):
                 self._enabled = False  # gpu is available but cpu is forced
 
         self.prev = self._enabled  # necessary reference to exit
-        active = self._enabled
```
paxcema, line 50: Removed, this method does not exist

PR #422, opened 2021-04-20
Increase autoencoder lr

George3d6:


paxcema:
Closing, as benchmarks do not show an improvement across the board

Issue #421, opened 2021-04-16
KeyError when training with metro_traffic_ts dataset


maximlopin:
mindsdb_native 2.40.0

reproduce:
`py -m benchmarks --datasets=metro_traffic_ts --frameworks=mindsdb --force_run`

error:
```
Traceback (most recent call last):
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\run.py", line 230, in run
    use_cache=False if args.force_run else True
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\frameworks\frameworks.py", line 40, in run
    result = self._run(ds, df_train, df_test)
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\frameworks\frameworks.py", line 208, in _run
    learn_kwargs=ds['learn_kwargs']
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\mdb\benchmark.py", line 22, in run
    predictor.learn(from_data=df_train, to_predict=target, **learn_kwargs)
  File "Q:\Desktop\mdb\mindsdb_native\mindsdb_native\libs\controllers\predictor.py", line 317, in learn
    self.transaction.run()
  File "Q:\Desktop\mdb\mindsdb_native\mindsdb_native\libs\controllers\transaction.py", line 277, in run
    self._run()
  File "Q:\Desktop\mdb\mindsdb_native\mindsdb_native\libs\controllers\transaction.py", line 273, in _run
    raise e
  File "Q:\Desktop\mdb\mindsdb_native\mindsdb_native\libs\controllers\transaction.py", line 246, in _run
    self._call_phase_module(module_name='ModelInterface', mode='train')
  File "Q:\Desktop\mdb\mindsdb_native\mindsdb_native\libs\controllers\transaction.py", line 178, in _call_phase_module
    ret = module(self.session, self)(**kwargs)
  File "q:\desktop\mdb\mindsdb_native\mindsdb_native\libs\phases\base_module.py", line 53, in __call__
    ret = self.run(**kwargs)
  File "q:\desktop\mdb\mindsdb_native\mindsdb_native\libs\phases\model_interface\model_interface.py", line 18, in run
    self.transaction.model_backend.train()
  File "q:\desktop\mdb\mindsdb_native\mindsdb_native\libs\phases\model_interface\lightwood_backend.py", line 359, in train
    config=lightwood_config
[...]

PR #424, opened 2021-04-28
Fix 421

paxcema:
## Why
To fix issue #421

## How
As the `grouped_by` key is optional, we retrieve its contents using the `.get()` accessor instead of `[]`.


requirements.txt
```diff
@@ -1,6 +1,6 @@
 NLTK >= 3, != 3.6
 python-dateutil <2.8.1,>=2.1
-numpy >= 1.16.2
+numpy >= 1.16.2,<=1.19.5
```
paxcema, line 3: Revert this

PR #431, opened 2021-05-25
Improved normalization in plain TS encoder

paxcema:
## Why
To improve MindsDB time series capabilities

## How
Instead of having its own normalizer, now the "plain" time series encoder (used to estimate the autoregressive weights in ArNet) further refines its output if the group combination has a matching normalizer that was trained in the standard time series encoder.


paxcema:
Closing for now; benchmark results show accuracy issues when introducing specialized normalizers in this encoder.

PR #432, opened 2021-05-27
LightwoodEnsemble

paxcema:
## Why
To continuously train predictors from streams of data.

## How
Added a `LightwoodEnsemble` class that holds and uses an array of `Predictor`s to operate.

As this is a proof of concept, predictions for now are extremely simple: mode is taken for classification targets, and mean for regression targets.

### DIFF
--- a/lightwood/api/ensemble.py
+++ b/lightwood/api/ensemble.py
@@ -1,14 +1,13 @@
-from .predictor import Predictor
+from lightwood import Predictor
 from lightwood.constants.lightwood import ColumnDataTypes
 from collections import Counter
 import numpy as np
 import pickle
-import torch
 import os
 
 
 class LightwoodEnsemble:
-    def __init__(self, predictor=None, predictors=None, load_from_path=None):
+    def __init__(self, predictors=None, load_from_path=None):
         self.path_list = None
         if load_from_path is not None:
             with open(os.path.join(load_from_path, 'lightwood_data'), 'rb') as pickle_in:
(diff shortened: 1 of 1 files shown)


lightwood/api/ensemble.py
```diff
@@ -0,0 +1,64 @@
...
+            formatted_predictions[target_name] = {}
+            pred_arr = np.array([p[target_name]['predictions'] for p in predictions])
+            if target['type'] == ColumnDataTypes.NUMERIC:
+                final_preds = np.mean(pred_arr, axis=0).tolist()
+            elif target['type'] == ColumnDataTypes.CATEGORICAL:
+                final_preds = []
+                for idx in range(pred_arr.shape[1]):
+                    final_preds.append(max(Counter(pred_arr[:, idx])))
+
+            formatted_predictions[target_name]['predictions'] = final_preds
```
maximlopin, line 47: If data type is neither categorical nor numeric, this line will raise `NameError`
Solution is to add an `else: raise Exception('only numeric and categorical datatypes are supported for ensembles')` above  in the `if else`

lightwood/api/ensemble.py
```diff
@@ -0,0 +1,64 @@
...
+from lightwood.constants.lightwood import ColumnDataTypes
+from collections import Counter
+import numpy as np
+import pickle
+import torch
+import os
+
+
+class LightwoodEnsemble:
+    def __init__(self, predictor=None, predictors=None, load_from_path=None):
```
maximlopin, line 11: Can check `isinstance(predictors, list)` / `isinstance(predictors, Predictor)` instead of having two parameters

maximlopin (approved):

PR #440, opened 2021-06-25
The big refactor: time series

paxcema:


lightwood/api/generate_config.py
```diff
@@ -145,10 +145,11 @@ def generate_config(type_information: TypeInformation, statistical_analysis: Sta
         timeseries_transformer = {
             'object': 'transform_timeseries',
             'config_args': {
-                'timeseries_settings': 'problem_definition.timeseries_settings'
+                'problem_definition': 'problem_definition'
```
George3d6, line 148: Could you request the particular members you need and not the whole problem defintion, in line with keeping coupling as minimal as possible. Ideally we should aim for requesting no single value that is never used.

George3d6: Note: I know the rest of lightwood isn't perfect at this right now either, but I think it's the goal we should work towards.

tests/integration/advanced/test_timeseries.py
```diff
@@ -1,7 +1,41 @@
+import importlib
+
 from tests.integration.helpers import ClickhouseTest, break_dataset
+from lightwood.api.types import ProblemDefinition
 from lightwood.api import make_predictor
 
 
 class TestTimeseries(ClickhouseTest):
     def test_timeseries(self):
```
George3d6, line 9: This doesn't need to inherit from `ClickhouseTest`

Also could we add a second test (or maybe even 2 x) one testing multiple predictions and one testing anomalies.
Also might be worth adding a text for predicting the "next" value (i.e. target in the following row instead of the current row) for when that is implemented.

George3d6 (changes requested):

PR #443, opened 2021-06-30
Predictor from code string helper function

maximlopin:
very useful inside lightwood and mindsdb to avoid repeating code with importlib


George3d6:
`from lightwood.api.high_level import make_predictor`

Example of usage:

```
 predictor: lightwood.PredictorInterface = make_predictor(df_train, problem_definition)
```

PR #448, opened 2021-07-16
create temp file with os module (fix PermissionError)

maximlopin:


George3d6:
The reason this works and the "proper" way to create a temp file doesn't is... unkonw to me and odd, if you do know let me know, alas it works, so... yeah, this is the new way we do it.

PR #450, opened 2021-07-27
The big refactor: predict_proba

paxcema:
- Add method for estimating class probabilities in `Neural`, `LightGBM` and `Regression` models.
- Add tests for classification tasks.
- Remove a couple of unused datasets in `tests/`


lightwood/api/json_ai.py
```diff
@@ -534,6 +534,19 @@ def code_from_json_ai(json_ai: JsonAI) -> str:
...
+log.info('Cleaning the data')
+data = {call(json_ai.cleaner, json_ai)}
+
+{ts_transform_code}
+
+encoded_ds = lightwood.encode(self.encoders, data, self.target)
+df = self.ensemble(encoded_ds, predict_proba=True)
+return df
+"""
+    predict_proba_body = align(predict_proba_body, 2)
```
George3d6, line 548: maybe share this code between `predict` and `predict_proba`

George3d6: As in the code generation bit should be the same except for the last line that does the actual call

lightwood/encoder/categorical/autoencoder.py
```diff
@@ -101,12 +100,11 @@ def encode(self, column_data):
                 embeddings = self.encoder(oh_encoded_tensor)
                 return embeddings.to('cpu')
 
-    def decode(self, encoded_data):
-        self.onehot_encoder.predict_proba = self.predict_proba
+    def decode(self, encoded_data, predict_proba=False):
```
George3d6, line 103: this feels hacky, I'd prefer to skip the encoder altogether. At *worst* this should be called something like `get_raw_logits` or something like that, since it's basically just "give me the original one-hot vector".

lightwood/encoder/categorical/onehot.py
```diff
@@ -77,7 +76,7 @@ def encode(self, column_data):
 
         return torch.Tensor(ret)
 
-    def decode(self, encoded_data):
+    def decode(self, encoded_data, predict_proba=False):
```
George3d6, line 79: same as above, the encoder has nothing to do with predictions, what's being done here is something else, asking the encoder to... not decode, and instead just return the data it got + some tiny format mods.

lightwood/encoder/categorical/autoencoder.py
George3d6, line 103: !!IMPORATN!! The auto encoder doesn't need to do this period, we never use autoencoders for targets, wouldn't make much sense

lightwood/model/lightgbm.py
```diff
@@ -204,5 +205,12 @@ def __call__(self, ds: EncodedDs) -> pd.DataFrame:
         else:
             decoded_predictions = raw_predictions
 
-        ydf = pd.DataFrame({'prediction': decoded_predictions})
+        if predict_proba and self.ordinal_encoder is not None:
+            predictions = np.hstack([raw_predictions, decoded_predictions.reshape(-1, 1)])
+            cat_labels = [f'__mdb_proba_{label}' for label in self.ordinal_encoder.categories_[0].tolist()]
+            ydf = pd.DataFrame(predictions, columns=cat_labels + ['prediction'])
```
George3d6, line 211: I *think* a better solution here is:
```
ydf = pd.DataFrame({'prediction': decoded_predictions}) --- keep this bit instead of deleting
if predict_proba and self.ordinal_encoder is not None:
    for idx, label in for label in enumerate(self.ordinal_encoder.categories_[0].tolist()):
        ydf.insert(len(ydf), label, decoded_predictions[idx]
```

Note: Haven't ran the code, roughly this you might have to change it a bit.

It will be much faster, it's shorter and avoid an `else` *and* it's much easier to understand what it does.

lightwood/model/neural.py
```diff
@@ -265,12 +269,24 @@ def __call__(self, ds: EncodedDs) -> pd.DataFrame:
...
+                decoded_prediction = self.target_encoder.decode(Yh, **kwargs)
 
             if not self.timeseries_settings.is_timeseries or self.timeseries_settings.nr_predictions == 1:
                 decoded_predictions.extend(decoded_prediction)
             else:
                 decoded_predictions.append(decoded_prediction)
 
-        ydf = pd.DataFrame({'prediction': decoded_predictions})
+        if predict_proba:
+            predictions = np.hstack([np.array(all_probs).squeeze(), np.array(decoded_predictions).reshape(-1, 1)])
```
George3d6, line 286: See comment in the LGBM model, basically same thing here.

George3d6 (changes requested):
Other big things:

* Add `predict_proba` to `Regression` model
* Add a flag that indicates if a predictor can predict probabilities (set after the ensemble is fitted, since for some ensembles based on the model[s] they chose they may or may not have this capacity)
* Add tests!!! We don't have a category predicting test anyway, so just pick a categorical dataset of your choice and add some tests for this.

PR #452, opened 2021-07-30
The big refactor: model analysis

paxcema:
- Restore global column importance feature
- Add model-independent ICP normalizer for dynamic confidence estimation and bounds in numerical, categorical and forecasting tasks
- Add insights (i.e. confidence, acc stats, feature importance from model analysis phase) to `predict_proba()` calls


lightwood/analysis/nc/norm.py
```diff
@@ -1,24 +1,110 @@
...
+
+from lightwood.api.dtype import dtype
+from lightwood.analysis.nc.nc import BaseScorer
 from lightwood.helpers.device import get_devices
 from lightwood.helpers.torch import LightwoodAutocast
-from lightwood.analysis.nc.nc import BaseScorer
+from lightwood.data.encoded_ds import ConcatedEncodedDs
+
+
+class SelfawareNormalizer(BaseScorer):
```
George3d6, line 19: Any reason we ought not to have these normalizers inherit from `BaseModel` instead?

George3d6: I.e. that has the added advantage we could easily get existing models to behave as normalizer and vice/versa.

lightwood/data/encoded_ds.py
```diff
@@ -81,13 +81,25 @@ def get_encoded_column_data(self, column_name: str) -> torch.Tensor:
...
 
+    def get_encoded_data(self, include_target=True) -> torch.Tensor:
+        encoded_dfs = []
+        for col in self.data_frame.columns:
+            if (include_target or col != self.target) and self.encoders.get(col, False):
+                encoded_dfs.append(self.get_encoded_column_data(col))
+
+        return torch.cat(encoded_dfs, 1)
+
+    def clear_cache(self):
```
George3d6, line 92: This doesn't seem to be called anywhere. If so I'd eer on the side of removing it from this and contacted encoded ds.

Or does this lead to e.g. extra pickling volume during `save` ? (If so it should be called there)

paxcema: It's called in the column importance bit. Given it's just one call, I could replicate the behavior as a procedure in the analysis rather than as methods in the classes, but I'd keep it as it is.

lightwood/analysis/nc/norm.py
paxcema: Nope, `BaseScorer` is fairly minimal. I've changed to inherit from `BaseModel` and removed `SelfAwareNet` (basically a copy of `DefaultNet`, ergo redundant now).

Will further tinker with this as part of the `nc` interface simplification effort, so that e.g. `score()` isn't needed anymore.

Issue #462, opened 2021-08-16
Error if "use previous target" set to False


paxcema:
Currently, setting `use_previous_target = False` yields the following error:

```python
 line 109, in learn
    'original_type': self.dtype_dict[col],
KeyError: '__mdb_ts_previous_Traffic'
```

So, an issue when checking dependencies for the encoder. Additionally, this setting does not imply a switch from ArNet to DefaultNet in the Neural model, which is problematic because ArNet _needs_ previous target data to train and predict.

Issue #459, opened 2021-08-16
Missing configurable encoders parameters


paxcema:
For JSON AI editing purposes, the following params should be passed in the initializer signatures:

- TimeSeriesEncoder.encoder_class
- CategoricalAutoEncoder.use_autoencoder

PR #463, opened 2021-08-18
Fixes -- #459 #462

paxcema:
## Why
Fixes:

- #459
- #462

## How

For #459:
- Add respective parameters to `TimeSeriesEncoder` and `CategoricalAutoEncoder` signatures

For #462:
- Use `ArNet` in the Neural model only if `use_previous_target = True`
- Add `SkTime` to predictor models only if `use_previous_target = True`
- Decouple addition of future target values from previous target values in `timeseries_transform`

### Additional changes:
- Revert to not dropping the target column in predict mode if anomaly detection mode is active
- Expanded integration test to cover `anomaly_detection` on and off _and_ `use_previous_target` on vs off
- When needed, target type casting to `float` in analysis and array accuracy helper

### DIFF
--- a/lightwood/analysis/model_analyzer.py
+++ b/lightwood/analysis/model_analyzer.py
@@ -62,6 +62,7 @@ def model_analyzer(
     disable_column_importance = disable_column_importance or ts_cfg.is_timeseries or has_pretrained_text_enc
 
     data = encoded_data.data_frame
+    tr_data = encoded_train_data.data_frame
     runtime_analyzer = {}
     predictions = {}
     input_cols = list([col for col in data.columns if col != target])
@@ -130,11 +131,12 @@ def model_analyzer(
         elif is_multi_ts:
             # we fit ICPs for time series confidence bounds only at t+1 forecast
             icp.nc_function.model.prediction_cache = np.array([p[0] for p in normal_predictions['prediction']])
-
         else:
             icp.nc_function.model.prediction_cache = np.array(normal_predictions['prediction'])
 
         if not is_classification:
+            data[target] = data[target].astype(float)
+            tr_data[target] = tr_data[target].astype(float)
             runtime_analyzer['df_std_dev'] = {'__default': stats_info.df_std_dev}
 
         # fit additional ICPs in time series tasks with grouped columns
--- a/lightwood/analysis/model_analyzer.py
+++ b/lightwood/analysis/model_analyzer.py
@@ -62,7 +62,6 @@ def model_analyzer(
     disable_column_importance = disable_column_importance or ts_cfg.is_timeseries or has_pretrained_text_enc
 
     data = encoded_data.data_frame
-    tr_data = encoded_train_data.data_frame
     runtime_analyzer = {}
     predictions = {}
     input_cols = list([col for col in data.columns if col != target])
@@ -135,8 +134,6 @@ def model_analyzer(
             icp.nc_function.model.prediction_cache = np.array(normal_predictions['prediction'])
 
         if not is_classification:
-            data[target] = data[target].astype(float)
-            tr_data[target] = tr_data[target].astype(float)
             runtime_analyzer['df_std_dev'] = {'__default': stats_info.df_std_dev}
 
         # fit additional ICPs in time series tasks with grouped columns
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -145,15 +145,18 @@ def generate_json_ai(type_information: TypeInformation, statistical_analysis: St
                     'stop_after': '$problem_definition.seconds_per_model',
                     'n_ts_predictions': '$problem_definition.timeseries_settings.nr_predictions'
                 }
-            },
-                {
-                'module': 'SkTime',
-                'args': {
-                    'stop_after': '$problem_definition.seconds_per_model',
-                    'n_ts_predictions': '$problem_definition.timeseries_settings.nr_predictions',
-                },
-            }
-            ])
+            }])
+
+            if problem_definition.timeseries_settings.use_previous_target:
+                models.extend([
+                    {
+                    'module': 'SkTime',
+                    'args': {
+                        'stop_after': '$problem_definition.seconds_per_model',
+                        'n_ts_predictions': '$problem_definition.timeseries_settings.nr_predictions',
+                    },
+                }
+                ])
 
     outputs = {target: Output(
         data_dtype=type_information.dtypes[target],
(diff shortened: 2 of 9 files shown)


tests/integration/advanced/test_timeseries.py
```diff
@@ -44,38 +43,52 @@ def test_0_grouped_regression_timeseries(self):
...
-                                                'order_by': [order_by],
-                                                'window': 5
-                                            }
-                                            })
-
-        a[2] = predictor_from_problem(train, prdb)
-
-        a[2].learn(train)
-        preds = a[2].predict(test)
+        pred_arr = {}
```
paxcema, line 49: I wonder if keeping both tests here is still desired. May be better if we test the "multiple predictors in parallel" someplace else?

George3d6: Oh fuck,  100% remove this test it was a temp thing, shouldn't have been kept there.

lightwood/analysis/model_analyzer.py
```diff
@@ -130,11 +131,12 @@ def model_analyzer(
         elif is_multi_ts:
             # we fit ICPs for time series confidence bounds only at t+1 forecast
             icp.nc_function.model.prediction_cache = np.array([p[0] for p in normal_predictions['prediction']])
-
         else:
             icp.nc_function.model.prediction_cache = np.array(normal_predictions['prediction'])
 
         if not is_classification:
+            data[target] = data[target].astype(float)
+            tr_data[target] = tr_data[target].astype(float)
```
George3d6, line 139: Why does this happen here and not in the data cleaner ?

lightwood/api/json_ai.py
```diff
@@ -298,7 +301,9 @@ def add_implicit_values(json_ai: JsonAI) -> JsonAI:
             models[i]['args']['timeseries_settings'] = models[i]['args'].get(
                 'timeseries_settings', '$problem_definition.timeseries_settings')
             models[i]['args']['net'] = models[i]['args'].get(
-                'net', '"DefaultNet"' if not problem_definition.timeseries_settings.is_timeseries else '"ArNet"')
+                'net', '"DefaultNet"' if not problem_definition.timeseries_settings.is_timeseries
+                or not problem_definition.timeseries_settings.use_previous_target
```
George3d6, line 305: So wait, if we "use previous target" than we want to use the `DefaultNet` not the `ArNet` ... why ?

lightwood/data/cleaner.py
```diff
@@ -103,7 +103,8 @@ def cleaner(
         data: pd.DataFrame, dtype_dict: Dict[str, str],
         pct_invalid: float, ignore_features: List[str],
         identifiers: Dict[str, str],
-        target: str, mode: str, timeseries_settings: TimeseriesSettings) -> pd.DataFrame:
+        target: str, mode: str, timeseries_settings: TimeseriesSettings,
+        anomaly_detection: bool) -> pd.DataFrame:
```
George3d6, line 107: Wrong formatting, please use the `.flake8` file's line limit to format, in this case I think the line did not go over, even if it did it's probably only the last bit

lightwood/data/statistical_analysis.py
```diff
@@ -49,7 +49,8 @@ def statistical_analysis(data: pd.DataFrame,
     seed()
     log.info('Starting statistical analysis')
     df = cleaner(data, dtypes, problem_definition.pct_invalid, problem_definition.ignore_features,
-                 identifiers, problem_definition.target, 'train', problem_definition.timeseries_settings)
+                 identifiers, problem_definition.target, 'train', problem_definition.timeseries_settings,
```
George3d6, line 52: same comment as above about `.flake8` file indentation limit

lightwood/data/timeseries_transform.py
```diff
@@ -236,6 +241,15 @@ def _ts_add_previous_target(df, target, nr_predictions, window, data_dtype, mode
...
 
     df[f'__mdb_ts_previous_{target}'] = previous_target_values_arr
+    return df
+
+
+def _ts_add_future_target(df, target, nr_predictions, data_dtype, mode):
+    if target not in df:
+        return df
+    if data_dtype in (dtype.integer, dtype.float, dtype.array):
+        df[target] = df[target].astype(float)
```
George3d6, line 251: doesn't the casting already happen in the cleaner?

lightwood/encoder/categorical/autoencoder.py
```diff
@@ -11,7 +11,7 @@
 
 
 class CategoricalAutoEncoder(BaseEncoder):
-    def __init__(self, stop_after: int = 3600, is_target=False, max_encoded_length=100):
+    def __init__(self, stop_after: int = 3600, is_target=False, max_encoded_length=100, use_autoencoder=None):
```
George3d6, line 14: Make the default be a string called `'auto'` instead of `None` so it makes more sense.

Provide type for this and the rest of the args please

lightwood/encoder/time_series/rnn.py
```diff
@@ -23,12 +23,11 @@
 class TimeSeriesEncoder(BaseEncoder):
 
     def __init__(self, stop_after: int, is_target=False, original_type: str = None, target: str = None,
-                 grouped_by: List[str] = []):
+                 grouped_by: List[str] = [], encoder_type='rnn'):
```
George3d6, line 26: I'd kinda feel much better if the transformer based encoder as a separate encoder, but if this removes a lot of boilerplate compare to having tow separate ones, fine

lightwood/encoder/time_series/rnn.py
```diff
@@ -169,35 +174,35 @@ def prepare(self, priming_data, dependency_data=None, ts_analysis=None,
...
-                        if group_name != '__default':
-                            idxs, subset = get_group_matches(dep_data, normalizer.combination)
-                            normalized = normalizer.encode(subset).unsqueeze(-1)
-                            data[idxs, :, :] = normalized
-                            all_idxs -= set(idxs)
-                    if len(all_idxs) > 0 and '__default' in self.dep_norms[dep_name].keys():
-                        default_norm = self.dep_norms[dep_name]['__default']
-                        subset = [dep_data['data'][idx] for idx in list(all_idxs)]
-                        data[list(all_idxs), :, :] = torch.Tensor(default_norm.encode(subset)).unsqueeze(-1)
+        normalized_tensors = []
```
George3d6, line 177: Looked for a while, I don't get this bit of logic, maybe double-check it :?

George3d6 (changes requested):

lightwood/encoder/time_series/rnn.py
paxcema: Logic is untouched, I simply un-indented it as the default value for `depedency_data` is now also a Dict. I will add the type hints, though.

lightwood/data/timeseries_transform.py
paxcema: Seems like it's not (may be some edge case), but I agree it's sloppy to do this here, will make sure it happens in the cleaner instead.

lightwood/analysis/model_analyzer.py
paxcema: Agreed, will move to the data cleaner

lightwood/api/json_ai.py
paxcema: Other way around: if we _don't_ "use previous target" then we cannot use ArNet, as it uses the `TsPlainEncoder` that passes raw previous values to the model.

lightwood/data/statistical_analysis.py
paxcema: Hmh, seems like I might have some issues with my `.flake8`.

tests/integration/advanced/test_timeseries.py
paxcema: Actually, I'll keep both as separate tests, to increase coverage (in particular, test TS without anomalies, autoregressiveness, and single-step forecasting)

PR #482, opened 2021-08-27
Seed keyword passed

George3d6:
Batch improvements based on suggestions from @hakunanatasha after her extensive review of the codebase.

* Seed function no longer has a default value for the seed, the seed_nr being passed is configurable from all of the functions that `seed` is called in.

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -627,7 +627,7 @@ class Predictor(PredictorInterface):
     mode: str
 
     def __init__(self):
-        seed()
+        seed({json_ai.problem_definition.seed_nr})
         self.target = '{json_ai.problem_definition.target}'
         self.mode = 'innactive'
 
--- a/lightwood/helpers/seed.py
+++ b/lightwood/helpers/seed.py
@@ -3,7 +3,7 @@ import numpy as np
 import random
 
 
-def seed(seed_nr):
+def seed(seed_nr: int) -> None:
     torch.manual_seed(seed_nr)
     torch.backends.cudnn.deterministic = True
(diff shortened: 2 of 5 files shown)


lightwood/api/json_ai.py
```diff
@@ -627,7 +627,7 @@ class Predictor(PredictorInterface):
     mode: str
 
     def __init__(self):
-        seed()
+        seed({json_ai.problem_definition.seed_nr})
```
hakunanatasha, line 630: is json_ai global? How is this inherited? Shouldn't it be set from self?

lightwood/helpers/seed.py
```diff
@@ -3,7 +3,7 @@
 import random
 
 
-def seed(seed_nr=420):
+def seed(seed_nr):
```
hakunanatasha, line 6: type hint this

hakunanatasha (changes requested):
Mostly good - only issue is with json_ai; should this not be set as self.seed? If not, should be fine - will approve after comments addressed.

lightwood/api/json_ai.py
George3d6: No... this is happening in `json_ai.py` when we are constructing the code for the Predictor class.

Values specified in the problem_definition get inlined in the Predictor class.

I'm not sure the currently built predictor classes are the most readable... but the current questions we are focusing on are more towards json ai design, so I'd rather tackle the generated predictors afterwards.

hakunanatasha: got it

hakunanatasha (approved):
This is fine to merge.

PR #485, opened 2021-08-28
Fix analysis memory usage

paxcema:
## Why
Analysis is currently saving a copy of the model inside each group confidence model. This means for tasks with multiple time series, memory usage is much larger than needed.

## How
Custom `__deepcopy__()` method for `BaseModelNc` does not copy the model when doing `deepcopy(icp)` during analysis.

### Affected modules

 * `lightwood.analysis.nc`


hakunanatasha (commented):
I think this can be merged but some questions:
(1) Do we need something like this for the encoder?
(2) What happens in the unit mixer case where the model inherits from the encoder (the model is effectively in that case a wrapper around the encoder calls). Will we run into issues there?

PR #486, opened 2021-08-28
Adaptive normalizer for numerical time series

paxcema:
## Why
<pending>

## How
<pending>


lightwood/analysis/nc/nc.py
```diff
@@ -337,6 +338,17 @@ def score(self, x, y=None):
...
+    def __deepcopy__(self, memo={}):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            if k != 'model':  # model should not be copied
+                setattr(result, k, deepcopy(v, memo))
+            else:
+                setattr(result, k, v)
+        return result
```
George3d6, line 350: will steal this for another PR since I think it could be a great timesaver, also <i think> the normalizer should also be ignored here.

paxcema:
Will close for now. Benchmarks suggest no clear improvements and this variant is more complex than the current approach.

Issue #488, opened 2021-08-31
Incorrect return value for order column in time series tasks


George3d6:
Code:
```
import lightwood
import pandas as pd


train_data = pd.DataFrame({
    'order_col': list(range(1, 100)),
    'target_col': list(range(101, 200))
})

pred = lightwood.predictor_from_problem(train_data, lightwood.ProblemDefinition.from_dict({
    'target': 'target_col',
    'timeseries_settings': {
        'order_by': ['order_col'],
        'window': 3
    },
    'time_aim': 5
}))

pred.learn(train_data)

test_data = pd.DataFrame({
    'order_col': [105, 104, 110, 0, 5, 144],
    'target_col': [206, 205, 116, 6, 11, 160]
})

predictions = pred.predict(test_data)

print(predictions)
```

Predictions will be:

````
  truth  prediction order_order_col  confidence  lower  upper  anomaly
0  206.0         133         [105.0]      0.9991     69    197     True
1  205.0         145         [104.0]      0.9991     81    209    False
2  116.0         147         [110.0]      0.9991     83    211    False
3    6.0         156           [0.0]      0.9991     92    220     True
4   11.0         151           [5.0]      0.9991     87    215     True
5  160.0         143         [144.0]      0.9991     79    207    False
```

The fact that the order column is an array here seems wrong. I'd expect it to be the same as the original type (int/float). It being a 1-element array doesn't make any sense.

PR #493, opened 2021-09-06
Fix 461, 488, 495

### LINKED ISSUE #461: Better defaults for Neural model in time series tasks
Current default values for early stopping seem too aggressive, as the neural model is not able to correctly fit training data from e.g. arrivals.

paxcema:
## Why
Fixes #461, #488 and #495

## How

- #461: remove warmup step in `fit()` training loop _and_ modify policy for deciding to early stop by giving more importance to most recently seen errors, to avoid premature early stopping.
- #488: cast unitary forecasts from list to first (and only) value to keep original column format
- #495: add special handling for groups where `mean = 0.0`

### Affected modules
- `lightwood/model/neural`
- `lightwood/helpers/ts`
- `lightwood/encoder/numeric/ts_numeric`

### Accuracy changes

Reverting back to no warmup in neural model means that it is once again able to produce sane forecasts. Here is an example (monthly sunspots test split):

Before:
image

After:
image

As reference, new R2 score in one-step-ahead forecasts for the test set is `0.803`. Before, it was `-1.105`. Comparisons against `lightwood <= 0.69.0` + `mindsdb_native` were not performed.

### DIFF
--- a/lightwood/model/neural.py
+++ b/lightwood/model/neural.py
@@ -263,25 +263,18 @@ class Neural(BaseModel):
         criterion = self._select_criterion()
         scaler = GradScaler()
 
-        if not self.timeseries_settings.is_timeseries:
-            for subset_itt in (0, 1):
-                for subset_idx in range(len(dev_ds_arr)):
-                    train_dl = DataLoader(
-                        ConcatedEncodedDs(train_ds_arr[subset_idx * 9: (subset_idx + 1) * 9]),
-                        batch_size=200, shuffle=True)
-
-                    stop_after = self.stop_after / 4
-                    self.model, epoch_to_best_model, err = self._max_fit(
-                        train_dl, dev_dl, criterion, optimizer, scaler, stop_after / 2, 20000 if subset_itt > 0 else 1)
-
-                    self.epochs_to_best += epoch_to_best_model
-        else:
+        for subset_itt in (0, 1):
             for subset_idx in range(len(dev_ds_arr)):
                 train_dl = DataLoader(
                     ConcatedEncodedDs(train_ds_arr[subset_idx * 9: (subset_idx + 1) * 9]),
                     batch_size=200, shuffle=True)
+
+                stop_after = self.stop_after / 4
+                return_model_after = (20000 if subset_itt > 0 else 1) if not self.timeseries_settings.is_timeseries \
+                    else 20000
+
                 self.model, epoch_to_best_model, err = self._max_fit(
-                    train_dl, dev_dl, criterion, optimizer, scaler, self.stop_after, 20000)
+                    train_dl, dev_dl, criterion, optimizer, scaler, stop_after / 2, return_model_after)
 
                 self.epochs_to_best += epoch_to_best_model
 
--- a/lightwood/model/neural.py
+++ b/lightwood/model/neural.py
@@ -199,8 +199,8 @@ class Neural(BaseModel):
                 epochs_to_best = epoch
 
             if len(running_errors) >= 5:
-                delta_mean = np.mean([running_errors[-i - 1] - running_errors[-i]
-                                     for i in range(1, len(running_errors[-5:]))])
+                delta_mean = np.average([running_errors[-i - 1] - running_errors[-i] for i in range(1, 5)],
+                                        weights=[(1/2)**i for i in range(1, 5)])
                 if delta_mean <= 0:
                     break
             elif (time.time() - started) > stop_after:
--- a/lightwood/model/neural.py
+++ b/lightwood/model/neural.py
@@ -200,7 +200,7 @@ class Neural(BaseModel):
 
             if len(running_errors) >= 5:
                 delta_mean = np.average([running_errors[-i - 1] - running_errors[-i] for i in range(1, 5)],
-                                        weights=[(1/2)**i for i in range(1, 5)])
+                                        weights=[(1 / 2)**i for i in range(1, 5)])
                 if delta_mean <= 0:
                     break
             elif (time.time() - started) > stop_after:
--- a/lightwood/model/neural.py
+++ b/lightwood/model/neural.py
@@ -184,7 +184,9 @@ class Neural(BaseModel):
 
             train_error = np.mean(running_losses)
 
-            running_errors.append(self._error(dev_dl, criterion))
+            epoch_error = self._error(dev_dl, criterion)
+            running_errors.append(epoch_error)
+            log.info(f'Loss @ epoch {epoch}: {epoch_error}')
 
             if np.isnan(train_error) or np.isnan(
                     running_errors[-1]) or np.isinf(train_error) or np.isinf(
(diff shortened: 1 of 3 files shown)


lightwood/model/neural.py
```diff
@@ -267,10 +269,12 @@ def fit(self, ds_arr: List[EncodedDs]) -> None:
                     ConcatedEncodedDs(train_ds_arr[subset_idx * 9: (subset_idx + 1) * 9]),
                     batch_size=200, shuffle=True)
 
-                stop_after = self.stop_after / 4
+                stop_after = self.stop_after / 8 if not self.timeseries_settings.is_timeseries else self.stop_after
+                return_model_after = (20000 if subset_itt > 0 else 1) if not self.timeseries_settings.is_timeseries \
```
George3d6, line 273: Isn't this `20000 if subset_it  or self.timeseries_settings.is_timeseries else 1` ?

Also... why do this for time series? The idea in general is to do a "warmup" on each subset (1 epoch than return the model) *then* train until completion (or until ~1/3rd of the time runs out) on each subset.

lightwood/model/neural.py
```diff
@@ -184,7 +184,9 @@ def _max_fit(self, train_dl, dev_dl, criterion, optimizer, scaler, stop_after, r
 
             train_error = np.mean(running_losses)
 
-            running_errors.append(self._error(dev_dl, criterion))
+            epoch_error = self._error(dev_dl, criterion)
+            running_errors.append(epoch_error)
+            log.info(f'Loss @ epoch {epoch}: {epoch_error}')
```
George3d6, line 189: This will log too often in some cases, I think, add as `debug` if you think it's useful so we can disable it in mindsdb

George3d6 (changes requested):

lightwood/model/neural.py
paxcema: > Isn't this 20000 if subset_it or self.timeseries_settings.is_timeseries else 1 ?

Fixed

> Also... why do this for time series? The idea in general is to do a "warmup" on each subset (1 epoch than return the model) then train until completion (or until ~1/3rd of the time runs out) on each subset.

Yes, but this scheme leads to underfitting in seemingly all TS datasets. I’m not sure why, but the issue with current approach seems to be as follows in e.g. monthly sunspots dataset:

* warmup round is quite short -as expected- and loss barely decreases
* in the “full” training round (i.e. subset_itt=1) model loss is stuck, which early stops training based on the error mean being 0.0
* forecasts are bad

If, instead, I crank epochs to 20k in all rounds (as in this PR):
* warmup round does a nice amount of epochs (120 approx.) and early stops correctly but, crucially, the loss this time around did decrease a lot
* “full” round once again is not able to decrease loss by more than a marginal amount (this time, loss is not constant), gets early stopped after 5 epochs
* predictor gives good forecasts

The weird thing to me is that we are sharing optimizer between max_fit calls, and LR does not change, so why would the second loop have stiffer learning dynamics if the DL has shuffle=True, I do not know… :thinking_face:

paxcema: Okay, as discussed: warmup has been scrapped in favour of simplicity. Let's see what the automated benchmarks think of this.

lightwood/model/neural.py
```diff
@@ -197,8 +199,8 @@ def _max_fit(self, train_dl, dev_dl, criterion, optimizer, scaler, stop_after, r
                 epochs_to_best = epoch
 
             if len(running_errors) >= 5:
-                delta_mean = np.mean([running_errors[-i - 1] - running_errors[-i]
-                                     for i in range(1, len(running_errors[-5:]))])
+                delta_mean = np.average([running_errors[-i - 1] - running_errors[-i] for i in range(1, 5)],
+                                        weights=[(1 / 2)**i for i in range(1, 5)])
```
paxcema, line 203: As a note: while this + warmup removal helped, I think it would be nice to compare Neural-only models against results from the latest Native version.

George3d6: I disagree. If it improves benchmark performance overall then it's good to go, if no then it isn't.

Thinking about it this way:
- There's something the neural mixer is good at and some things it isn't
- If it ends up being bad at everything when compared to e.g. lightgbm that's bad and we should fix it or drop it
- If it's good at say 20-50% of things that's fine, and we should focus on those because the other mixers cover the other cases (and this we can track on global benchmarks)
- If we think it can be good at a problem space where it isn't already, this can also be tracked via the global benchmark performance, because "good" should translate to "better than lightgbm"
- As long as a change doesn't make it worst for the problem-space where it's good at, it doesn't really matter if performance drops in other problem spaces.

Long-term maybe it should matter, but we're going to be adding so many new mixers that I'd be way too complex, for now, to give this treatment to all of them. And how would we act on the results, even if we did?

What if a change makes the NN worst overall but better on the problems it's already good at? I'd still say the change is good to go. So would having this suite severely alter our behaviour ?

lightwood/model/neural.py
```diff
@@ -261,18 +263,12 @@ def fit(self, ds_arr: List[EncodedDs]) -> None:
         criterion = self._select_criterion()
         scaler = GradScaler()
 
-        for subset_itt in (0, 1):
-            for subset_idx in range(len(dev_ds_arr)):
-                train_dl = DataLoader(
-                    ConcatedEncodedDs(train_ds_arr[subset_idx * 9: (subset_idx + 1) * 9]),
-                    batch_size=200, shuffle=True)
+        train_dl = DataLoader(ConcatedEncodedDs(train_ds_arr), batch_size=200, shuffle=True)
```
George3d6, line 266: I was suggesting removing the "warmup" subset iteration, I doubt removing it entirely will help, given that I tried that at some point and the results were much worst. But training methodology has changed a bit since then, so who knows. Let's look at the benchmarks.

lightwood/model/neural.py
hakunanatasha: To follow up on a discussion we had this week - the base "behavior" of a model should be expected (i.e. neural should be a vanilla DNN). 

Exotic optimizers, training strategies, and other things should be considered in addition (ideally as abstracted classes etc). I am ok with more "advanced" forms of getting our NN to work (hyper parameter tuning and architecture behavior is non-trivial; realistically getting it to generalize on any dataset with a fixed architecture is not going to work) but let's at least ensure the traditional behavior of a model is captured in the naming scheme.

hakunanatasha: As an aside- what rationale inspired the weights for the average mean? I can understand you want to weight "more recent" differences than later ones, but is this a standard practice?

lightwood/model/neural.py
hakunanatasha: An additional point - batch_size should be a passable parameter

lightwood/model/neural.py
hakunanatasha: TO ADD - from discussing with @paxcema (2021.09.09) we will convert the weights criteria here into an `EarlyStopping` abstraction for modularity

PR #497, opened 2021-09-08
Nonconformist optimizations (for speed)

George3d6:
## Why ?

We need to make lightwood's inferences faster. Non conformist is very slow, `explain` takes 40-90% of inference time, see these tests for detail: https://github.com/mindsdb/magazine_of_magical_minutia/blob/master/performance_benchmarks/Results_Lightwood.md

## How

Optimizations to `explain increase performance across the board.

Example on bench 3

Before:
```
3.376 <module>  lightwood_benchmark_3.py:1
`- 3.367 predict  ../../../tmp/4f35588c6bb1fff0ac002432f937e88e2054d0e88762253d16309827790861123.py:162
   |- 2.931 explain  lightwood/analysis/explain.py:14
```

After:
```
1.747 <module>  lightwood_benchmark_3.py:1
`- 1.733 predict  ../../../tmp/5f283d544a1c1c511a03088822d4e48f62b680f06ec161f816311097846519938.py:162
   |- 1.398 explain  lightwood/analysis/explain.py:13
```

Example on bench 1

Before:
```
10.420 <module>  lightwood_benchmark_1.py:1
`- 10.409 predict  ../../../tmp/9446c53b9c6e2de2a7608e9ba8049b19392d1ce48a727f26163098244731725.py:163
   |- 5.391 explain  lightwood/analysis/explain.py:14
```

After:
```
8.570 <module>  lightwood_benchmark_1.py:1
`- 8.563 predict  ../../../tmp/af76354f4b30c2802dd9b04d3e91a00b43728b8a05f97c6616311102106541724.py:163
   |- 4.338 explain  lightwood/analysis/explain.py:13
```

These are run on my laptop, relative improvements are:

bench 1: `explain` took 51.7% of the runtime, now 50.6% -- NOISE
bench 3: `explain` took 87% of runtime, now 80% -- Signal?

### DIFF
--- a/lightwood/analysis/explain.py
+++ b/lightwood/analysis/explain.py
@@ -4,7 +4,6 @@ import torch
 import numpy as np
 import pandas as pd
 
-# restore_icp_state, clear_icp_state
 from lightwood.analysis.nc.util import get_numerical_conf_range, get_categorical_conf, get_anomalies
 from lightwood.helpers.ts import get_inferred_timestamps, add_tn_conf_bounds
 from lightwood.api.dtype import dtype
@@ -94,6 +93,8 @@ def explain(data: pd.DataFrame,
 
             # base ICP
             X = deepcopy(icp_X)
+            # Ugly optimization but accessing `values` actually wasted a lot of time
+            icp_values = X.values
 
             # get all possible ranges
             if timeseries_settings.is_timeseries and timeseries_settings.nr_predictions > 1 and is_numerical:
@@ -101,11 +102,11 @@ def explain(data: pd.DataFrame,
                 # bounds in time series are only given for the first forecast
                 analysis['icp']['__default'].nc_function.model.prediction_cache = \
                     [p[0] for p in predictions['prediction']]
-                all_confs = analysis['icp']['__default'].predict(X.values)
+                all_confs = analysis['icp']['__default'].predict(icp_values)
 
             elif is_numerical:
                 analysis['icp']['__default'].nc_function.model.prediction_cache = predictions['prediction']
-                all_confs = analysis['icp']['__default'].predict(X.values)
+                all_confs = analysis['icp']['__default'].predict(icp_values)
 
             # categorical
             else:
@@ -122,7 +123,7 @@ def explain(data: pd.DataFrame,
 
                 conf_candidates = list(range(20)) + list(range(20, 100, 10))
                 all_ranges = np.array(
-                    [analysis['icp']['__default'].predict(X.values, significance=s / 100)
+                    [analysis['icp']['__default'].predict(icp_values, significance=s / 100)
                      for s in conf_candidates])
                 all_confs = np.swapaxes(np.swapaxes(all_ranges, 0, 2), 0, 1)
 
--- a/lightwood/analysis/nc/icp.py
+++ b/lightwood/analysis/nc/icp.py
@@ -195,7 +195,7 @@ class IcpClassifier(BaseIcp, ClassifierMixin):
         # TODO: if x == self.last_x ...
         n_test_objects = x.shape[0]
         p = np.zeros((n_test_objects, self.classes.size))
-        
+
         for i, c in enumerate(self.classes):
             test_class = np.zeros(x.shape[0], dtype=self.classes.dtype)
             test_class.fill(c)
--- a/lightwood/analysis/nc/icp.py
+++ b/lightwood/analysis/nc/icp.py
@@ -195,7 +195,10 @@ class IcpClassifier(BaseIcp, ClassifierMixin):
         # TODO: if x == self.last_x ...
         n_test_objects = x.shape[0]
         p = np.zeros((n_test_objects, self.classes.size))
-        random_uniform_dist = np.random.uniform(0, 1, 1)
+        
+        if self.smoothing:
+            random_uniform_dist = np.random.uniform(0, 1, 1)
+        
         for i, c in enumerate(self.classes):
             test_class = np.zeros(x.shape[0], dtype=self.classes.dtype)
             test_class.fill(c)
--- a/lightwood/analysis/nc/icp.py
+++ b/lightwood/analysis/nc/icp.py
@@ -194,6 +194,7 @@ class IcpClassifier(BaseIcp, ClassifierMixin):
         # TODO: if x == self.last_x ...
         n_test_objects = x.shape[0]
         p = np.zeros((n_test_objects, self.classes.size))
+        random_uniform_dist = np.random.uniform(0, 1, 1)
         for i, c in enumerate(self.classes):
             test_class = np.zeros(x.shape[0], dtype=self.classes.dtype)
             test_class.fill(c)
@@ -215,7 +216,7 @@ class IcpClassifier(BaseIcp, ClassifierMixin):
                 p[j, i] = n_gt / (n_cal + 1)
 
                 if self.smoothing:
-                    p[j, i] += (n_eq * np.random.uniform(0, 1, 1)) / (n_cal + 1)
+                    p[j, i] += (n_eq * random_uniform_dist) / (n_cal + 1)
                 else:
                     p[j, i] += n_eq / (n_cal + 1)
 
--- a/lightwood/analysis/nc/icp.py
+++ b/lightwood/analysis/nc/icp.py
@@ -196,9 +196,6 @@ class IcpClassifier(BaseIcp, ClassifierMixin):
         n_test_objects = x.shape[0]
         p = np.zeros((n_test_objects, self.classes.size))
         
-        if self.smoothing:
-            random_uniform_dist = np.random.uniform(0, 1, 1)
-        
         for i, c in enumerate(self.classes):
             test_class = np.zeros(x.shape[0], dtype=self.classes.dtype)
             test_class.fill(c)
@@ -221,7 +218,7 @@ class IcpClassifier(BaseIcp, ClassifierMixin):
                         n_gt += 1
 
                 if self.smoothing:
-                    p[j, i] = (n_gt + n_eq * random_uniform_dist) / (n_cal + 1)
+                    p[j, i] = (n_gt + n_eq * np.random.uniform(0, 1, 1)) / (n_cal + 1)
                 else:
                     p[j, i] = (n_gt + n_eq) / (n_cal + 1)
 
--- a/lightwood/analysis/nc/icp.py
+++ b/lightwood/analysis/nc/icp.py
@@ -208,16 +208,18 @@ class IcpClassifier(BaseIcp, ClassifierMixin):
                 cal_scores = self.cal_scores[self.condition((x[j, :], c))][::-1]
                 n_cal = cal_scores.size
 
-                idx_left = np.searchsorted(cal_scores, nc, 'left')
-                idx_right = np.searchsorted(cal_scores, nc, 'right')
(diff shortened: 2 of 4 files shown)


lightwood/analysis/nc/icp.py
```diff
@@ -207,17 +209,18 @@ def predict(self, x, significance=None):
...
+                        n_eq += 1
+                    elif nc < cal_score:
+                        n_gt += 1
 
                 if self.smoothing:
-                    p[j, i] += (n_eq * np.random.uniform(0, 1, 1)) / (n_cal + 1)
+                    p[j, i] = (n_gt + n_eq) / (n_cal + 1)
                 else:
-                    p[j, i] += n_eq / (n_cal + 1)
+                    p[j, i] = (n_gt + n_eq * random_uniform_dist) / (n_cal + 1)
```
paxcema, line 223: This is backwards, the smoothed version should have `n_eq * random_uniform_dist`, not the other way around.

paxcema (changes requested):
Looks good overall! Interesting to see stuff like accesing `.values` being expensive. 

Apart from the other comment I made, I think this also needs a test for ensuring the `.predict()` output has not changed.

lightwood/analysis/nc/icp.py
George3d6: For the before/after. I printed the `labels` from the nc unittest.

This branch:

```
1st test:
[0.01005034 6.90775528 0.09622803 6.90775528]

2nd test:
[0.5        0.5        0.5        0.57781513 1.5        0.5       ]
```


Staging:

```
1st test:
[0.01005034 6.90775528 0.09622803 6.90775528]

2nd test:
[0.5        0.5        0.5        0.57781513 1.5        0.5       ]
```

paxcema (approved):

lightwood/analysis/explain.py
```diff
@@ -94,18 +93,20 @@ def explain(data: pd.DataFrame,
 
             # base ICP
             X = deepcopy(icp_X)
+            # Ugly optimization but accessing `values` actually wasted a lot of time
```
hakunanatasha, line 96: Remove this comment. `.values` is a standard technique and we don't need a post mortem

lightwood/analysis/explain.py
```diff
@@ -94,18 +93,20 @@ def explain(data: pd.DataFrame,
...
 
                 # bounds in time series are only given for the first forecast
                 analysis['icp']['__default'].nc_function.model.prediction_cache = \
                     [p[0] for p in predictions['prediction']]
-                all_confs = analysis['icp']['__default'].predict(X.values)
+                all_confs = analysis['icp']['__default'].predict(icp_values)
 
             elif is_numerical:
                 analysis['icp']['__default'].nc_function.model.prediction_cache = predictions['prediction']
-                all_confs = analysis['icp']['__default'].predict(X.values)
```
hakunanatasha, line 108: I'm assuming calling X.values twice actually increased the speed. Do you have evidence? As in, can you perform a timing test on how long it actually takes to call this function?

lightwood/analysis/nc/icp.py
```diff
@@ -194,6 +195,7 @@ def predict(self, x, significance=None):
         # TODO: if x == self.last_x ...
         n_test_objects = x.shape[0]
         p = np.zeros((n_test_objects, self.classes.size))
+        random_uniform_dist = np.random.uniform(0, 1, 1)
         for i, c in enumerate(self.classes):
```
hakunanatasha, line 199: What is the use of the uniform dist here?

lightwood/analysis/nc/icp.py
```diff
@@ -207,17 +209,18 @@ def predict(self, x, significance=None):
                 cal_scores = self.cal_scores[self.condition((x[j, :], c))][::-1]
                 n_cal = cal_scores.size
 
-                idx_left = np.searchsorted(cal_scores, nc, 'left')
```
hakunanatasha, line 210: Why is `np.searchsorted` slower than a for loop? Do you have a timing test?

lightwood/analysis/nc/icp.py
```diff
@@ -207,17 +209,18 @@ def predict(self, x, significance=None):
...
-                p[j, i] = n_gt / (n_cal + 1)
+                n_eq = 0
+                n_gt = 0
+                for cal_score in cal_scores:
+                    if cal_score == nc:
+                        n_eq += 1
+                    elif nc < cal_score:
+                        n_gt += 1
 
                 if self.smoothing:
```
hakunanatasha, line 220: Is this flag typically on? In the prior case, we don't make a random uniform distro UNLESS this flag is true, but here, we make the var then pass it.

lightwood/analysis/nc/icp.py
```diff
@@ -207,17 +209,18 @@ def predict(self, x, significance=None):
...
-                    p[j, i] += (n_eq * np.random.uniform(0, 1, 1)) / (n_cal + 1)
+                    p[j, i] = (n_gt + n_eq * random_uniform_dist) / (n_cal + 1)
                 else:
-                    p[j, i] += n_eq / (n_cal + 1)
+                    p[j, i] = (n_gt + n_eq) / (n_cal + 1)
```
hakunanatasha, line 223: The math here has changed; previously you were adding to the cell and now you're replacing it. Why?

lightwood/analysis/nc/nc.py
```diff
@@ -327,23 +327,23 @@ def score(self, x, y=None):
             Nonconformity scores of samples.
         """ # noqa
         prediction = self.model.predict(x)
-        n_test = x.shape[0]
 
-        norm = np.ones(n_test)
+        err = self.err_func.apply(prediction, y)
```
hakunanatasha, line 331: err is not normed here and was before in the return. Is this intended?

lightwood/analysis/nc/nc.py
```diff
@@ -327,23 +327,23 @@ def score(self, x, y=None):
...
         if self.normalizer is not None:
             try:
                 norm = self.normalizer.score(x) + self.beta
+                err = err / norm
             except Exception:
                 pass
 
-        return self.err_func.apply(prediction, y) / norm
```
hakunanatasha, line 339: prior use case err / norm is passed and self.normalizer changes the norm. Is this intended?

lightwood/analysis/nc/nc.py
```diff
@@ -466,19 +466,19 @@ def predict(self, x, nc, significance=None):
         n_test = x.shape[0]
         prediction = self.model.predict(x)
 
-        norm = np.ones(n_test)
+        norm = 1
```
hakunanatasha, line 469: why is it scalar when before it needed to be an array?

lightwood/analysis/nc/nc.py
```diff
@@ -466,19 +466,19 @@ def predict(self, x, nc, significance=None):
...
             except Exception:
                 pass
 
         if significance:
-            intervals = np.zeros((x.shape[0], 2))
             err_dist = self.err_func.apply_inverse(nc, significance)
             err_dist = np.hstack([err_dist] * n_test)
             err_dist *= norm
 
+            intervals = np.zeros((x.shape[0], 2))
```
hakunanatasha, line 481: I like the change of grouping relevant pieces together.

lightwood/analysis/nc/wrappers.py
```diff
@@ -19,25 +18,6 @@ def clear_icp_state(icp):
         icp.normalizer.model = None
 
 
-def restore_icp_state(col, hmd, session):
```
hakunanatasha, line 22: why do we remove this?

hakunanatasha (changes requested):
It is not obvious to me why this is supposed to work faster. The only change I see that may have possible caused extra time is `X.values` being called twice; if this is the case, please time how long it takes to call this command.

lightwood/analysis/explain.py
George3d6: yes but assigning `.values` to something isn't. The weird behavior here is that `.value` actually seems to => some function invocation in the background, it's not just a reference

lightwood/analysis/nc/icp.py
George3d6: No it's not, made it so that the dist is created only when it's on, even faster :)

lightwood/analysis/nc/icp.py
George3d6: because it's faster, we skip 1x assignment and turn 2x divisions into 1x divsion

George3d6: increases speed by removing 2x costly operations (1 assignment, 1 division)

lightwood/analysis/nc/nc.py
George3d6: It is normed bellow ` err = err / norm` we just skip norming when `norm = np.ones(n_test)` aka unit vector aka equivalent to not norming at all... that bit made no sense tbh

lightwood/analysis/nc/nc.py
George3d6: see my other comment, error still gets normed but we streamline the case where `norm=1`

lightwood/analysis/nc/wrappers.py
George3d6: Because it was a dead function, never used.

lightwood/analysis/nc/icp.py
paxcema: See section 5, eq. 13 [here](http://alrw.net/articles/02.pdf) for the reason. 

We actually need to sample a different tie-breaker value for every `p[j, i]`, so the optimization here introduces a bug, please revert it so that `np.random.uniform` gets called for each `cal_score`.

paxcema (changes requested):
See comment above, this introduces a small bug that needs to be fixed before merging.

For reference:

> We actually need to sample a different tie-breaker value for every `p[j, i]`, so the optimization here introduces a bug, please revert it so that `np.random.uniform` gets called for each `cal_score`.

lightwood/analysis/nc/icp.py
paxcema: See above, this introduces a bug :eyes:

lightwood/analysis/nc/nc.py
paxcema: I don't think it _needed_ to be an array. This change forces `numpy` to broadcast it anyway. Not sure if it's faster or not, though.

lightwood/analysis/nc/icp.py
hakunanatasha: Additionally, please make that citation as a comment or within the docstring because for future use, we may want to refer to the research rationale here.

hakunanatasha:
When you did timing tests, did you specifically isolate with .values or did you include all changes? The two areas I could see slow-downs are from (1) the .values and (2) the change in the error function logic. For sanity, try at least a test where .values gets called twice. I'm not averse to just assigning this variable but I want to make sure that is the actual cause.

George3d6:
@hakunanatasha  The tests profile each individual function call, `.values` consumed a significant (~5%?) chunk of time before as opposed to no time now.

@paxcema fixed, now generating a new distribution each time.

hakunanatasha:
If the profiling did show that the .values command was eating time I am ok with the changes so long as @paxcema approves the NC logic.

paxcema (approved):

PR #499, opened 2021-09-09
Add MASE metric to select forecasters

paxcema:
## Why
It is widely acknowledged that MASE (mean absolute scaled error) is a superior error metric for evaluating time series forecasting. Our current method is to get the mean R2 score throughout the forecast horizon, which is not optimal.

## How
This PR adds computation of group-wise naive forecasting residuals in the training set, as part of the `timeseries_analyzer` module. Lightwood then uses these scores to compute the MASE metric, using it in the model selection phase of the `BestOf` ensemble.

### Affected modules
- `lightwood/api/json_ai`
- `lightwood/data/timeseries_analyzer`
- `lightwood/encoder/time_series/helpers/common`
- `lightwood/ensemble/best_of`
- `lightwood/helpers/general`

### Notes
One of the modules planned for the near future is to automatically figure out the seasonality of a series. Once this is implemented, the computation of naive forecasts for seasonal series should change to actually use this offset instead of `t-1`, as it  this PR does.


lightwood/data/timeseries_analyzer.py
```diff
@@ -19,6 +22,11 @@ def timeseries_analyzer(data: pd.DataFrame, dtype_dict: Dict[str, str],
     # @TODO: maybe normalizers should fit using only the training subsets??
     new_data = generate_target_group_normalizers(info)
 
+    if dtype_dict[target] in (dtype.integer, dtype.float, dtype.array):
```
hakunanatasha, line 25: should this be `tsarray`?

lightwood/data/timeseries_analyzer.py
```diff
@@ -56,5 +68,35 @@ def get_delta(df: pd.DataFrame, ts_info: dict, group_combinations: list, order_c
...
+
+def get_naive_residuals(target_data: pd.DataFrame, m: int = 1) -> Tuple[List, float]:
+    """
+    Computes forecasting residuals for the naive method (forecasts for time `t` is the value observed at `t-1`).
+    Useful for computing MASE forecasting error.
+
+    Note: method assumes predictions are all for the same group combination. For a dataframe that contains multiple
+     series, use `get_grouped_naive_resiudals`.
+
+    m: season length. the naive forecasts will be the m-th previously seen value for each series
```
hakunanatasha, line 84: Change to:

:param target_data: The input sequence for forecasting <might be wrong?>
:param m: season length; naive forecasts will be the m-th previously seen value for each series

:returns: <whatever it returns>

lightwood/encoder/time_series/helpers/common.py
```diff
@@ -92,7 +92,7 @@ def get_group_matches(data, combination):
     if isinstance(data['data'], np.ndarray) and len(data['data'].shape) < 2:
         data['data'] = np.expand_dims(data['data'], axis=1)
 
-    if not combination:
+    if not combination or combination == '__default':
```
hakunanatasha, line 95: this one is awkward but I can't come up with a better arg. Is there a reason __default is diff than False?

lightwood/ensemble/best_of.py
```diff
@@ -13,7 +13,8 @@
 class BestOf(BaseEnsemble):
     best_index: int
 
-    def __init__(self, target, mixers: List[BaseMixer], data: List[EncodedDs], accuracy_functions) -> None:
+    def __init__(self, target, mixers: List[BaseMixer], data: List[EncodedDs], accuracy_functions,
+                 ts_analysis: Optional[dict] = None) -> None:
```
hakunanatasha, line 17: double check if Optional[dict] already fills in None as default

lightwood/helpers/general.py
```diff
@@ -50,18 +55,88 @@ def evaluate_multilabel_accuracy(true_values, predictions, **kwargs):
...
+    if fh == 1:
+        predictions = [[p] for p in predictions]
+
+    # only evaluate accuracy for rows with complete historical context
+    if kwargs['ts_analysis'].get('tss', False):
+        true_values = true_values[kwargs['ts_analysis']['tss'].window:]
+        predictions = predictions[kwargs['ts_analysis']['tss'].window:]
 
-        aggregate += base_acc_fn(true_values[i][:valid_horizon], predictions[i][:valid_horizon])
+    for i in range(fh):
```
hakunanatasha, line 139: Does aggregate get trigged appropriately if fh=1?

hakunanatasha (commented):
Few comments to highlight docstrings but otherwise fine 

General approve

lightwood/ensemble/best_of.py
paxcema: Does not, I'm afraid

PR #500, opened 2021-09-10
Make explainer work with categorical targets not present in the training data

### LINKED ISSUE #489: Nonconfrmist issue when test dataset doesn't contain values that existed in the train dataset
When there's a categorical target and nonconformist isn't getting a certain categorical value present in the train set in the validation dataset this seems to lead to errors:

```
    self.model_analysis, self.runtime_analyzer = model_analyzer(data=test_data,train_data=train_data,disable_column_importance=False,fixed_significance=None,confidence_normalizer=False,stats_info=self.statistical_analysis,ts_cfg=self.problem_definition.timeseries_settings,accuracy_functions=self.accuracy_functions,predictor=self.ensemble,target=self.target,dtype_dict=self.dtype_dict,positive_domain=self.statistical_analysis.positive_domain)
  File "/home/george/lightwood/lightwood/analysis/model_analyzer.py", line 153, in model_analyzer
    icp_df, y = clean_df(icp_df, target, is_classification, runtime_analyzer.get('label_encoders', None))
  File "/home/george/lightwood/lightwood/analysis/nc/util.py", line 15, in clean_df
    y = np.array([cats.index(i) for i in y])
  File "/home/george/lightwood/lightwood/analysis/nc/util.py", line 15, in <listcomp>
    y = np.array([cats.index(i) for i in y])
ValueError: 'IT' is not in list
```

Was fixed in native so hopefully we can just copy-paste from there (?)

George3d6:
## Why 

NC crashes if there's a categorical target value in the validation/test data (the one that it's fit on) which it is not present in the train data.

## How

We fix this by just adding an `if this is a value not present in the train data distribution for the target, replace it with the "unknown category" label`

Added a test to catch and replicate this behaviour 

## Other

Added type signatures to a bunch of the explainer modules

closes #489  [Note: the description in the issue is bad because initially I thought the bug was smth else, but then I couldn't replace]


paxcema:
There's lots of shared logic with #497 in here, so I'll wait until that gets merged to staging (and then to this one) before reviewing.

hakunanatasha (commented):
If Patricio approves of the NC logic I am ok with this.

tests/integration/basic/test_weird_target_dist.py
```diff
@@ -0,0 +1,20 @@
...
+        # in all datasets (train/dev/validation) that were not present in the others
+        df = pd.DataFrame({
+            'target': [1 for _ in range(500)] + [f'{i}cat' for i in range(100)],
+            'y': [i for i in range(600)]
+        })
+        target = 'target'
+
+        predictor = predictor_from_problem(df, ProblemDefinition.from_dict({'target': target, 'time_aim': 200}))
+        predictor.learn(df)
+        predictor.predict(df)
```
btseytlin, line 20: How about some assert? Maybe assert that the missing targets were encoded as expected?

George3d6: No point, we just want to make sure it doesn't crash when the validation see contains things that aren't in train, that's all. And that is implicitly happening due to the number of distinct targets we generate. Might be worth explaining this better though...

paxcema: Even if it is implicitly happening (which imo it's not very easy to see unless you're familiar with lightwood internals) I think it's better to have 2 different dataframes here to guarantee there's stuff the ICP hasn't seen in training when doing the `predict()` call.

tests/unit_tests/analysis/test_nc_norm.py
```diff
@@ -10,6 +10,7 @@ def test_compute_numerical_labels(self):
         bounds = [0.5, 1.5]
 
         labels = Normalizer.compute_numerical_labels(preds, truths, bounds)
+        print(labels)
```
paxcema, line 13: Remove `print()` call

tests/unit_tests/analysis/test_nc_norm.py
```diff
@@ -20,6 +21,7 @@ def test_compute_categorical_labels(self):
         truths = np.array([[0, 1], [0, 1], [1, 0], [1, 0]])
 
         labels = Normalizer.compute_categorical_labels(preds, truths)
+        print(labels)
```
paxcema, line 24: Remove `print()` call

lightwood/analysis/nc/util.py
```diff
@@ -12,7 +12,8 @@ def clean_df(df, target, is_classification, label_encoders):
     if is_classification:
         if enc and isinstance(enc.categories_[0][0], str):
             cats = enc.categories_[0].tolist()
-            y = np.array([cats.index(i) for i in y])
+            # the last element is "__mdb_unkonw_cat"
```
paxcema, line 15: Typo, should read `__mdb_unknown_cat`

lightwood/analysis/nc/icp.py
```diff
@@ -104,16 +101,16 @@ def calibrate(self, x, y, increment=False):
...
-
-    def _update_calibration_set(self, x, y, increment):
+    def _update_calibration_set(self, x: np.array, y: np.array, increment: bool) -> np.array:
         if increment and self.cal_x is not None and self.cal_y is not None:
             self.cal_x = np.vstack([self.cal_x, x])
             self.cal_y = np.hstack([self.cal_y, y])
         else:
             self.cal_x, self.cal_y = x, y
 
+    def _calibrate_hook(self, x, y, increment):
```
paxcema, line 111: Missing type hints, same as below would be OK

lightwood/analysis/nc/icp.py
```diff
@@ -104,16 +101,16 @@ def calibrate(self, x, y, increment=False):
             cal_scores = self.nc_function.score(self.cal_x, self.cal_y)
             self.cal_scores = {0: np.sort(cal_scores)[::-1]}
 
-    def _calibrate_hook(self, x, y, increment):
-        pass
-
-    def _update_calibration_set(self, x, y, increment):
+    def _update_calibration_set(self, x: np.array, y: np.array, increment: bool) -> np.array:
```
paxcema, line 104: I believe the return hint here should be `None`

paxcema (changes requested):

tests/integration/basic/test_weird_target_dist.py
George3d6: Then I'd have to call the training functions one by one instead. But in practice you can test the test works (as I did) by running it on staging, and seeing it fails with the same error listed in the issue

paxcema (approved):

PR #501, opened 2021-09-10
Refactor: grouped TS splitter into helper function

paxcema:
## Why

To easily access this method as a helper function from e.g. the forecast visualizer

## How

Simply abstracting the logic into its own function.

### Affected modules

`lightwood/data/splitter`


lightwood/data/splitter.py
```diff
@@ -7,6 +7,9 @@
 
 
 def splitter(data: pd.DataFrame, k: int, tss: TimeseriesSettings) -> List[pd.DataFrame]:
+    """
+    Splits a dataframe into k equally-sized folds.
```
hakunanatasha, line 11: I hate the term folds here as it's not clear what "folds" should be.

Are folds equally sized to the training data? are folds subsets of the data? Please clarify.

lightwood/data/splitter.py
```diff
@@ -19,16 +22,27 @@ def splitter(data: pd.DataFrame, k: int, tss: TimeseriesSettings) -> List[pd.Dat
             folds = np.array_split(data, k)
         else:
             gcols = tss.group_by
-            all_group_combinations = list(product(*[data[gcol].unique() for gcol in tss.group_by]))
-            folds = [pd.DataFrame() for _ in range(k)]
-            for group in all_group_combinations:
-                subframe = data
-                for idx, gcol in enumerate(gcols):
-                    subframe = subframe[subframe[gcol] == group[idx]]
+            folds = grouped_ts_splitter(data, k, gcols)
```
hakunanatasha, line 25: happy about this abstraction, immensely.

lightwood/data/splitter.py
```diff
@@ -19,16 +22,27 @@ def splitter(data: pd.DataFrame, k: int, tss: TimeseriesSettings) -> List[pd.Dat
...
+    return folds
+
+
+def grouped_ts_splitter(data: pd.DataFrame, k: int, gcols: List[str]):
+    """
+    Splitter for grouped time series tasks, where there is a set of `gcols` columns by which data is grouped.
+    Each group yields a different time series, and the splitter generates `k` folds from `data`,
+    with equally-sized sub-series for each group.
+    """
+    all_group_combinations = list(product(*[data[gcol].unique() for gcol in gcols]))
```
hakunanatasha, line 36: I'm actually not fully aware what product(*<var>) does. Can you elaborate?

lightwood/data/splitter.py
```diff
@@ -19,16 +22,27 @@ def splitter(data: pd.DataFrame, k: int, tss: TimeseriesSettings) -> List[pd.Dat
...
+    folds = [pd.DataFrame() for _ in range(k)]
```
hakunanatasha, line 37: please explicitly define folds, alternatively, call them subsets (as I believe it's closer to the intended goal). I recognize we will likely change splitter's intended role, but for the sake of clarity it helps.

lightwood/data/splitter.py
```diff
@@ -19,16 +22,27 @@ def splitter(data: pd.DataFrame, k: int, tss: TimeseriesSettings) -> List[pd.Dat
...
+    for group in all_group_combinations:
+        subframe = data
```
hakunanatasha, line 39: This might not apply: do we want to use deepcopy here as opposed to subframe=data?

hakunanatasha (changes requested):

hakunanatasha (changes requested):
As an aside, why wouldn't we add this to the helpers function folder?

lightwood/data/splitter.py
paxcema: Yes, agreed on renaming to `subsets`

lightwood/data/splitter.py
paxcema: Indeed, equally sized. As discussed elsewhere, will be renaming to `subset`.

lightwood/data/splitter.py
paxcema: It yields the cartesian product of all observed values for each `group_by` column.

So, if `gby_col_a` has `A, B` as possible values, and `gby_col_b` has `C, D` as possible values, this line of code determines that all possible time series are given by the following combinations:

`[(A, C), (B, C), (A, D), (B, D)]`

lightwood/data/splitter.py
paxcema: Well, these subsets should form a partition over the original dataframe, so I'd think it's not a problem.

paxcema:
> As an aside, why wouldn't we add this to the helpers function folder?

We could, though I'm partial to leaving all splitting logic here, else this module would end up empty. I defer the choice to you, though.

lightwood/data/splitter.py
hakunanatasha: Great - be nice to make a small note of that in the docstring explicitly with the phrase you used earlier ("It yields the cartesian product of all observed values for each group_by column.")

hakunanatasha (approved):

PR #502, opened 2021-09-13
Enhanced date-time + tag histograms

### LINKED ISSUE #479: Generate histogram for all types
Title says it all, for types where generating one is pointless just have an empty one, I guess, because otherwise the GUI can't display it...

George3d6:
## Why

We weren't generating histograms for DateTime and tag types, our histograms for dates were arguably bad (using categorical histogram).

## How

Added  a numeric-like histogram for dates, using categorical histogram for tags

Resolves #479  | Obviously histograms can still be improved but this is the  bare minimum to get some decent graphics in the GUI in the vast majority of cases.

### DIFF
--- a/lightwood/data/statistical_analysis.py
+++ b/lightwood/data/statistical_analysis.py
@@ -1,7 +1,8 @@
 from typing import Dict
 import pandas as pd
 import numpy as np
-
+import datetime
+from dateutil.parser import parse as parse_dt
 from lightwood.api import StatisticalAnalysis, ProblemDefinition
 from lightwood.helpers.numeric import filter_nan
 from lightwood.helpers.seed import seed
@@ -10,6 +11,24 @@ from lightwood.helpers.log import log
 from lightwood.api.dtype import dtype
 from scipy.stats import entropy
 from lightwood.data.cleaner import _clean_float_or_none
+
+
+def get_datetime_histogram(data, data_dtype):
+    data = [_clean_float_or_none(parse_dt(str(x)).timestamp()) for x in data]
+
+    Y, X = np.histogram(data, bins=min(50, len(set(data))),
+                        range=(min(data), max(data)), density=False)
+    if data_dtype == dtype.integer:
+        Y, X = np.histogram(data, bins=[int(round(x)) for x in X], density=False)
+
+    X = X[:-1].tolist()
+    Y = Y.tolist()
+
+    X = [datetime.datetime.fromtimestamp(x) for x in X]
+    return {
+        'x': X,
+        'y': Y
+    }
 
 
 def get_numeric_histogram(data, data_dtype):
@@ -84,7 +103,7 @@ def statistical_analysis(data: pd.DataFrame,
     for col in df.columns:
         histograms[col] = None
         buckets[col] = None
-        if dtypes[col] in (dtype.categorical, dtype.binary, dtype.date):
+        if dtypes[col] in (dtype.categorical, dtype.binary, dtype.tags):
             hist = dict(df[col].value_counts().apply(lambda x: x / len(df[col])))
             histograms[col] = {
                 'x': list([str(x) for x in hist.keys()]),
@@ -94,8 +113,10 @@ def statistical_analysis(data: pd.DataFrame,
         elif dtypes[col] in (dtype.integer, dtype.float, dtype.array):
             histograms[col] = get_numeric_histogram(filter_nan(df[col]), dtypes[col])
             buckets[col] = histograms[col]['x']
+        elif dtypes[col] in (dtype.date, dtype.datetime):
+            histograms[col] = get_datetime_histogram(filter_nan(df[col]), dtypes[col])
         else:
-            histograms[col] = {'x': [], 'y': []}
+            histograms[col] = {'x': ['Unknown'], 'y': [1]}
             buckets[col] = []
 
     # get observed classes, used in analysis
--- a/lightwood/data/statistical_analysis.py
+++ b/lightwood/data/statistical_analysis.py
@@ -4,7 +4,7 @@ import numpy as np
 import datetime
 from dateutil.parser import parse as parse_dt
 from lightwood.api import StatisticalAnalysis, ProblemDefinition
-from lightwood.helpers.numeric import filter_nan
+from lightwood.helpers.numeric import filter_nan_and_none
 from lightwood.helpers.seed import seed
 from lightwood.data.cleaner import cleaner
 from lightwood.helpers.log import log
@@ -38,7 +38,7 @@ def get_numeric_histogram(data: pd.Series, data_dtype: dtype, bins: int) -> Dict
     """Generate the histogram for integer and float typed data
     """
     data = [_clean_float_or_none(x) for x in data]
-
+    
     Y, X = np.histogram(data, bins=min(bins, len(set(data))),
                         range=(min(data), max(data)), density=False)
     if data_dtype == dtype.integer:
@@ -116,12 +116,12 @@ def statistical_analysis(data: pd.DataFrame,
             }
             buckets[col] = histograms[col]['x']
         elif dtypes[col] in (dtype.integer, dtype.float, dtype.array):
-            histograms[col] = get_numeric_histogram(filter_nan(df[col]), dtypes[col], 50)
+            histograms[col] = get_numeric_histogram(filter_nan_and_none(df[col]), dtypes[col], 50)
             buckets[col] = histograms[col]['x']
         elif dtypes[col] in (dtype.date, dtype.datetime):
-            histograms[col] = get_datetime_histogram(filter_nan(df[col]), 50)
+            histograms[col] = get_datetime_histogram(filter_nan_and_none(df[col]), 50)
         else:
-            histograms[col] = {'x': ['Unknown'], 'y': [len(filter_nan(df[col]))]}
+            histograms[col] = {'x': ['Unknown'], 'y': [len(df[col])]}
             buckets[col] = []
 
     # get observed classes, used in analysis
--- a/lightwood/data/statistical_analysis.py
+++ b/lightwood/data/statistical_analysis.py
@@ -13,7 +13,9 @@ from scipy.stats import entropy
 from lightwood.data.cleaner import _clean_float_or_none
 
 
-def get_datetime_histogram(data, data_dtype):
+def get_datetime_histogram(data: pd.Series) -> Dict[str, list]:
+    """Generates the histogram for date and datetime types
+    """
     if isinstance(data[0], float) or isinstance(data[0], int):
         data = [_clean_float_or_none(x) for x in data]
     else:
@@ -21,8 +23,6 @@ def get_datetime_histogram(data, data_dtype):
 
     Y, X = np.histogram(data, bins=min(50, len(set(data))),
                         range=(min(data), max(data)), density=False)
-    if data_dtype == dtype.integer:
-        Y, X = np.histogram(data, bins=[int(round(x)) for x in X], density=False)
 
     X = X[:-1].tolist()
     Y = Y.tolist()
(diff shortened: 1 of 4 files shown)


lightwood/data/statistical_analysis.py
```diff
@@ -84,7 +106,7 @@ def statistical_analysis(data: pd.DataFrame,
     for col in df.columns:
         histograms[col] = None
         buckets[col] = None
-        if dtypes[col] in (dtype.categorical, dtype.binary, dtype.date):
+        if dtypes[col] in (dtype.categorical, dtype.binary, dtype.tags):
```
paxcema, line 109: Maybe split `tag` values and turn them into a set? That way you cover the case where a tag column does not have ordered tags.

lightwood/data/statistical_analysis.py
```diff
@@ -12,6 +13,27 @@
...
+def get_datetime_histogram(data, data_dtype):
+    if isinstance(data[0], float) or isinstance(data[0], int):
+        data = [_clean_float_or_none(x) for x in data]
+    else:
+        data = [_clean_float_or_none(parse_dt(str(x)).timestamp()) for x in data]
+
+    Y, X = np.histogram(data, bins=min(50, len(set(data))),
+                        range=(min(data), max(data)), density=False)
+    if data_dtype == dtype.integer:
+        Y, X = np.histogram(data, bins=[int(round(x)) for x in X], density=False)
```
paxcema, line 25: For `integer` timestamps it may be useful to compute the minimum interval, then fill with `None`s if it's unevenly spaced data.

paxcema (approved):
Couple of observations that, as you said, could be tackled in future iterations. As it currently stands, I think this is OK.

lightwood/data/statistical_analysis.py
hakunanatasha: I dislike the logic implemented above.  I'd prefer a uniform, if we want integer, interval for bins. This looks like a bad hack to get integer bins for visualization.

A simple approach could be bins = np.arange(min_value-1, max_value+1, delta) where delta is the frequency of the interval. 

RE @paxcema : Nones is not a preference; the histogram function should automatically return 0 for any "unpopulated" terms. For unevenly spaced data, the graphs will demonstrate "gaps". Users should be able to (if necessary) rerun an analysis histogram with a different set of bins.

To follow up on my earlier comment; I would prefer users specify the delta between bins frequency and we can defer to the min/max of the data distribution for the thresholds. If we note outliers, the user can be at least alerted and drop/handle them as necessary.

hakunanatasha: Alternatively, they can specify the bins themselves.

lightwood/data/statistical_analysis.py
```diff
@@ -12,6 +13,27 @@
...
+
+    X = X[:-1].tolist()
```
hakunanatasha, line 27: I'm assuming this if for plotting purposes to enable X and Y to be of the same length.

lightwood/data/statistical_analysis.py
```diff
@@ -12,6 +13,27 @@
...
+    Y = Y.tolist()
+
+    X = [datetime.datetime.fromtimestamp(x) for x in X]
```
hakunanatasha, line 30: Do we have a unit-test to demonstrate this works as intended?

lightwood/data/statistical_analysis.py
```diff
@@ -12,6 +13,27 @@
...
+    return {
+        'x': X,
+        'y': Y
+    }
+
+
 def get_numeric_histogram(data, data_dtype):
```
hakunanatasha, line 37: type hint + docstring missing

lightwood/data/statistical_analysis.py
hakunanatasha: Where is the split necessary? This is specifically checking the type of the column.

lightwood/data/statistical_analysis.py
```diff
@@ -94,8 +116,10 @@ def statistical_analysis(data: pd.DataFrame,
         elif dtypes[col] in (dtype.integer, dtype.float, dtype.array):
             histograms[col] = get_numeric_histogram(filter_nan(df[col]), dtypes[col])
             buckets[col] = histograms[col]['x']
+        elif dtypes[col] in (dtype.date, dtype.datetime):
+            histograms[col] = get_datetime_histogram(filter_nan(df[col]), dtypes[col])
         else:
-            histograms[col] = {'x': [], 'y': []}
+            histograms[col] = {'x': ['Unknown'], 'y': [1]}
```
hakunanatasha, line 122: Do not arbitrarily set 'y' to 1. At least make it None.

Additionally may make sense to set 'x' to None also.

hakunanatasha (changes requested):
There is an issue with this file to begin with in line 85 and line 89. For this reason, I do not think this is ready to PR. Additionally see edits.

Secondly: More type-specific histogram is not a descriptive title. This should be specifically called 'Enables appropriate histograms for datetime'.

This PR is missing the "modules affected". I need you to add this in the description.

lightwood/data/statistical_analysis.py
paxcema: In `hist.keys()`. It would require using something other than `value_counts` though. 

Example: a movie dataset with a `genres` tag column. We'd like both `action,thriller` and `thriller,action` to count for the same bin, I'd presume.

lightwood/data/statistical_analysis.py
paxcema: > Nones is not a preference; the histogram function should automatically return 0 for any "unpopulated" terms.

Right, I had envisioned the behavior you just described. Somehow I silly thought of `None`s as `== 0`, but indeed `0` is the correct choice.

lightwood/data/statistical_analysis.py
```diff
@@ -12,6 +13,27 @@
 from lightwood.data.cleaner import _clean_float_or_none
 
 
+def get_datetime_histogram(data, data_dtype):
```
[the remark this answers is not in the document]
George3d6: added and even included some extra

lightwood/data/statistical_analysis.py
```diff
@@ -12,6 +13,27 @@
...
+    if isinstance(data[0], float) or isinstance(data[0], int):
+        data = [_clean_float_or_none(x) for x in data]
+    else:
+        data = [_clean_float_or_none(parse_dt(str(x)).timestamp()) for x in data]
+
+    Y, X = np.histogram(data, bins=min(50, len(set(data))),
```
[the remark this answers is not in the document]
George3d6: Added as argument to functions, it's a magic number... is it any good? Dunno, we need to review the whole statistical analysis and where it's used internally vs in the GUI and go over magic number like these, but that's a separate issue

lightwood/data/statistical_analysis.py
```diff
@@ -12,6 +13,27 @@
...
+                        range=(min(data), max(data)), density=False)
```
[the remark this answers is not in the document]
George3d6: I'd rather change that as part of a different PR, see above

lightwood/data/statistical_analysis.py
George3d6: Hmh, no, but I can add histogram checking to some unittests

lightwood/data/statistical_analysis.py
George3d6: Yes

tests/integration/basic/test_analyze_dataset.py
```diff
@@ -16,11 +16,11 @@
 
 
 class TestInferTypes(unittest.TestCase):
-    def test_infer_types_on_home_rentlas(self):
+    def test_analyze_home_rentlas(self):
         datasource = FileDS(
             "https://raw.githubusercontent.com/mindsdb/mindsdb-examples/master/classics/home_rentals/dataset/train.csv"
         )
-        type_information = infer_types(datasource.df, pct_invalid=0)
+        type_information = analyze_dataset(datasource.df).type_information
```
hakunanatasha, line 23: how is `analyze_dataset` diff from `infer_types` now?

tests/integration/basic/test_analyze_dataset.py
```diff
@@ -47,8 +47,9 @@ def test_infer_types_on_home_rentlas(self):
         for k in type_information.identifiers:
             self.assertTrue(type_information.identifiers[k] is None)
 
-    def test_type_deduction(self):
-        n_points = 100
+    def test_with_generated(self):
+        # Must be an even number
+        n_points = 134
```
hakunanatasha, line 52: why this number? when it was 100?

lightwood/helpers/numeric.py
```diff
@@ -7,9 +7,10 @@
 
 Sorry, python is dumb, but the spirit of your investigation wasn't, try again
 '''
+from typing import Iterable
 
 
-def can_be_nan_numeric(value):
+def can_be_nan_numeric(value: object) -> bool:
```
hakunanatasha, line 13: The logic of this function is not obvious at all; I recognize this is to explicitly handle np.nan-like situations where they can be converted into a str/float but not int representation, but we need to highlight this for others to understand. Please docstring the logic; it can be crude

lightwood/helpers/numeric.py
```diff
@@ -25,5 +26,5 @@ def can_be_nan_numeric(value):
     return isnan
 
 
-def filter_nan(series):
-    return [x for x in series if not can_be_nan_numeric(x)]
+def filter_nan_and_none(series: Iterable) -> list:
```
hakunanatasha, line 29: Iterable is a little vague.

lightwood/helpers/numeric.py
```diff
@@ -25,5 +26,5 @@ def can_be_nan_numeric(value):
...
+    return [x for x in series if not can_be_nan_numeric(x) and x is not None]
```
hakunanatasha, line 30: This is problematic; there is no guarantee for the length of the array to be preserved:

See following:
```
from lightwood.helpers.numeric import *
import numpy as np

def filter_nan_and_none(series) -> list:
    return [x for x in series if not can_be_nan_numeric(x) and x is not None]

x = [0, np.nan, 5, 1, 5, None, np.nan]

print("Original length of list", len(x))

y = filter_nan_and_none(x)
print("New Filtered length", len(y))
```

tests/integration/basic/test_analyze_dataset.py
hakunanatasha, line 52: please indicate why 134 is used as opposed to 100.

tests/integration/basic/test_analyze_dataset.py
```diff
@@ -47,8 +47,9 @@ def test_infer_types_on_home_rentlas(self):
         for k in type_information.identifiers:
             self.assertTrue(type_information.identifiers[k] is None)
 
-    def test_type_deduction(self):
-        n_points = 100
+    def test_with_generated(self):
+        # Must be an even number
```
hakunanatasha, line 51: this comment is out of date, remove or justify

hakunanatasha (changes requested):
Helpers numeric function doesn't make sense to me. The proposed situation filters out Nones, which is what it's expected to do, but then will not preserve length. This can lead to issues with cleaning not preserving the net length of the sequence order. Is this only used in the statistical analysis? Otherwise, this is going to lead to issues later where the wrong rows are assigned a term.

The better place to filter is within stats analysis to ensure order is preserved.

tests/integration/basic/test_analyze_dataset.py
George3d6: when checking the length I wanted to make sure it's understood length == length of data. With `100` people might think it's 100% (0 to 100)

tests/integration/basic/test_analyze_dataset.py
George3d6: It just says that the magic number must be even, due to a quirk in data generation it won't work if it's odd.

lightwood/helpers/numeric.py
George3d6: Will do

tests/integration/basic/test_analyze_dataset.py
George3d6: `analyze_dataset` = > return a data analysis class that contains both the type information and the stat analysis

lightwood/helpers/numeric.py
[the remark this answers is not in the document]
George3d6: Not really an issue, this is a function used purely in stat analysis to remove nan and None for the purpose of generating the histograms

PR #512, opened 2021-09-15
Better detection, cleaning and encoding of arrays

### LINKED ISSUE #433: Could not convert string to float error
Using [twitter airline sentiment](https://www.kaggle.com/crowdflower/twitter-airline-sentiment) dataset to predict the `airline_sentiment`, the bellow error happens on model training start:
```
Some weights of the model checkpoint at distilbert-base-uncased were not used when initializing DistilBertForSequenceClassification: ['vocab_transform.weight', 'vocab_transform.bias', 'vocab_layer_norm.weight', 'vocab_layer_norm.bias', 'vocab_projector.weight', 'vocab_projector.bias']
- This IS expected if you are initializing DistilBertForSequenceClassification from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPretraining model).
- This IS NOT expected if you are initializing DistilBertForSequenceClassification from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of DistilBertForSequenceClassification were not initialized from the model checkpoint at distilbert-base-uncased and are newly initialized: ['pre_classifier.weight', 'pre_classifier.bias', 'classifier.weight', 'classifier.bias']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
/opt/conda/lib/python3.8/site-packages/lightwood/encoders/text/helpers/pretrained_helpers.py:22: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
item['labels'] = torch.tensor(self.labels[idx])
ERROR:mindsdb-logger-1534b558-bd85-11eb-9ca6-0242ac110002---417f6474-2893-458f-8695-3859450f31dc:/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/controllers/transaction.py:173 - Could not load module ModelInterface
ERROR:mindsdb-logger-1534b558-bd85-11eb-9ca6-0242ac110002---417f6474-2893-458f-8695-3859450f31dc:/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/controllers/transaction.py:263 - could not convert string to float: '0.0 0.0'
Predictor learn error: could not convert string to float: '0.0 0.0'
File "/opt/conda/lib/python3.8/site-packages/mindsdb/interfaces/model/learn_process.py", line 67, in run_learn
mdb.learn(
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/controllers/predictor.py", line 314, in learn
self.transaction.run()
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/controllers/transaction.py", line 269, in run
self._run()
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/controllers/transaction.py", line 265, in _run
raise e
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/controllers/transaction.py", line 238, in _run
self._call_phase_module(module_name='ModelInterface', mode='train')
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/controllers/transaction.py", line 170, in _call_phase_module
ret = module(self.session, self)(**kwargs)
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/phases/base_module.py", line 53, in __call__
ret = self.run(**kwargs)
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/phases/model_interface/model_interface.py", line 18, in run
self.transaction.model_backend.train()
File "/opt/conda/lib/python3.8/site-packages/mindsdb_native/libs/phases/model_interface/lightwood_backend.py", line 407, in train
lightwood_train_ds = lightwood.api.data_source.DataSource(
File "/opt/conda/lib/python3.8/site-packages/lightwood/api/data_source.py", line 116, in __init__
self.encoders = self._prepare_encoders()
File "/opt/conda/lib/python3.8/site-packages/lightwood/api/data_source.py", line 411, in _prepare_encoders
encoder_instance = self._prepare_column_encoder(config,
File "/opt/conda/lib/python3.8/site-packages/lightwood/api/data_source.py", line 359, in _prepare_column_encoder
encoder_instance.prepare(column_data, previous_target_data=training_data['previous'])
File "/opt/conda/lib/python3.8/site-packages/lightwood/encoders/time_series/rnn.py", line 130, in prepare
priming_data, lengths_data = self._prepare_raw_data(priming_data)
File "/opt/conda/lib/python3.8/site-packages/lightwood/encoders/time_series/rnn.py", line 101, in _prepare_raw_data
e = np.array(e, dtype=float)
ValueError: could not convert string to float: '0.0 0.0'
```

George3d6:
Fixes #433 which was failing because we couldn't handle coordinate data.

#502 Did most of the fixing for array-type detection and handling that was necessary to get this PR going, but this PR:

1. Changes the `cleaner` so that it turns arrays into python lists instead of strings of numbers separated by ` ` (this is the same behavior as the timeseries array reshaper)
2. Makes the `ArrayEncoder` the default for handling arrays (timeseries arrays are still handled by the RNN)
3. Renames the `TsPlainEncoder` to the `ArrayEncoder` and makes some minor changes to it to allow it to handle `None` values


George3d6:
Really confusing as to why this fails here but not locally. Will wait until the other things are merged to staging then try agian.

lightwood/api/json_ai.py
```diff
@@ -7,8 +7,8 @@
                                  Output, ProblemDefinition)
 
 
-trainable_encoders = ('PretrainedLangEncoder', 'CategoricalAutoEncoder', 'TimeSeriesEncoder', 'TimeSeriesPlainEncoder')
-ts_encoders = ('TimeSeriesEncoder', 'TimeSeriesPlainEncoder', 'TsNumericEncoder')
+trainable_encoders = ('PretrainedLangEncoder', 'CategoricalAutoEncoder', 'TimeSeriesEncoder', 'ArrayEncoder')
```
hakunanatasha, line 10: We need to figure out a better solution for this @paxcema @George3d6 but for now ok

lightwood/api/json_ai.py
```diff
@@ -26,7 +26,7 @@ def lookup_encoder(
         dtype.image: 'Image.Img2VecEncoder',
         dtype.rich_text: 'Rich_Text.PretrainedLangEncoder',
         dtype.short_text: 'Short_Text.CategoricalAutoEncoder',
-        dtype.array: 'Array.TimeSeriesEncoder',
+        dtype.array: 'Array.ArrayEncoder',
```
hakunanatasha, line 29: ts_array needs to go in here as well with an addition to `lw.api.dtype`.

hakunanatasha: @paxcema

lightwood/encoder/time_series/__init__.py
```diff
@@ -1,5 +1,5 @@
 from lightwood.encoder.time_series.rnn import TimeSeriesEncoder
-from lightwood.encoder.time_series.plain import TimeSeriesPlainEncoder
+from lightwood.encoder.array.array import ArrayEncoder
 
 
-__all__ = ['TimeSeriesEncoder', 'TimeSeriesPlainEncoder']
+__all__ = ['TimeSeriesEncoder', 'ArrayEncoder']
```
hakunanatasha, line 5: the fact array is within the time_series init is very weird given the goal of array is supposed to be sequences in general.

lightwood/data/cleaner.py
```diff
@@ -63,6 +63,8 @@ def _standardize_array(element):
         # Handles cases where arrays are numbers
         if ' ' not in element:
             element = _clean_float_or_none(element)
+        else:
+            element = [float(x) for x in element.split(' ')]
```
hakunanatasha, line 67: Are you indicated sequences that look like this? "1 2 3 5 6 1 5 5" which is why you split? What happens if a user has them comma separated or something else?

hakunanatasha: Also what if they happen to be integers as opposed to floats?

hakunanatasha (changes requested):
The fact array is in the time-series encoder subsection is a little misleading; I recognize they might borrow very similar principles but it needs to be thought through more carefully.

I had a few comments I'd like to address - let's chat on Monday with details.

lightwood/api/json_ai.py
George3d6: > The fact array is in the time-series encoder subsection is a little misleading; I recognize they might borrow very similar principles but it needs to be thought through more carefully.

That was a mistake on my end, I think it will behave perfectly if it's not part of those.

lightwood/api/json_ai.py
George3d6: But it needn't be part of this PR... I think

lightwood/encoder/time_series/__init__.py
George3d6: Again, I think this is a mistake, it should indeed not be here

lightwood/data/cleaner.py
George3d6: See above lines, this case is handled, arrays are standardized to a ` ` split string and then split... though this might be a bad way of doing it, since it was done when our array columns were typed as `str` in the dataframe, now they use `list`.

hakunanatasha (approved):

hakunanatasha:
The bulk of this PR actually doesn't handle coordinates explicitly. Please change the name to "array/sequence handling" or something along the lines of that.

PR #524, opened 2021-09-16
Fix #510 - Sktime dependency

### LINKED ISSUE #510: `sktime` missing dependency
In requirements.txt
`sktime[arima] >= 0.5.0`
bat 
`(mindsdb) C:\Users\User>pip install sktime[arima]>=0.5.0
WARNING: sktime 0.7.0 does not provide the extra 'arima'`
install
`(mindsdb) C:\Users\User>pip install sktime[arima]==0.5.0`

paxcema:
## Why
To fix #510 -- a warning that pops up while installing lightwood dependencies

## How
Remove `[arima]` from `sktime` dependency, as `pmdarima` is enough.

Unrelated, but this PR also supresses warnings from `statsmethods` in the `AutoARIMA` model, as the user cannot act upon them and they are not very informative.

### Affected modules
- `requirements.txt`

### DIFF
--- a/lightwood/model/sktime.py
+++ b/lightwood/model/sktime.py
@@ -42,6 +42,7 @@ class SkTime(BaseModel):
                                for gcol in self.grouped_by} if self.ts_analysis['tss'].group_by else {}}
 
         for group in self.ts_analysis['group_combinations']:
+            # many warnings might be thrown inside of statsmodels during stepwise procedure
             self.models[group] = self.model_class(suppress_warnings=True)
 
             if self.grouped_by == ['__default']:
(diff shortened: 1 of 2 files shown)


George3d6 (approved):

lightwood/model/sktime.py
```diff
@@ -42,7 +42,7 @@ def fit(self, ds_arr: List[EncodedDs]) -> None:
                                for gcol in self.grouped_by} if self.ts_analysis['tss'].group_by else {}}
 
         for group in self.ts_analysis['group_combinations']:
-            self.models[group] = self.model_class()
+            self.models[group] = self.model_class(suppress_warnings=True)
```
hakunanatasha, line 45: maybe slight preference to make this a default flag but its a small note

hakunanatasha (approved):
Maybe make suppress_warnings a flag that's default False in case for debugging but honestly fine as is

PR #525, opened 2021-09-16
Fix #508 - `allow_incomplete_history` for time series predictors

### LINKED ISSUE #508: Reinstate `allow_incomplete_history` in time series forecasting
At the moment, `transform_timeseries()` is not checking for the `allow_incomplete_history` parameter, this should be reinstated.

paxcema:
## Why
Fix for #508. Some of our time series forecasters have "cold starts", meaning they will not output good predictions until there is enough historical context (as defined during training with the `window` parameter). The user could still want to predict with an incomplete history. We had the `allow_incomplete_history` flag, but it was dropped at some point. This PR restores it.

## How
- Add `allow_incomplete_history` to `TimeseriesSettings`, with `False` as default value for now (as 2/3 forecasters exhibit cold starts).
- Add checks for this in the numerical TS tests.
- Additionally, reduced `JsonAi` verbosity by aliasing `problem_definition.timeseries_settings` to `tss` throughout the code.
- Fix an (as of yet undetected) bug in one of the conditionals for non-TS predictors with some TS arguments enabled.

## Affected modules
- `api/json_ai`
- `api/types`
- `data/timeseries_transform`
- `tests/integration/advanced/test_timeseries`


lightwood/data/timeseries_transform.py
```diff
@@ -132,11 +133,14 @@ def transform_timeseries(
         combined_df = pd.DataFrame(combined_df[combined_df['__mdb_make_predictions'].astype(bool).isin([True])])
         del combined_df['__mdb_make_predictions']
 
-    if len(combined_df) == 0:
-        raise Exception(f'Not enough historical context to make a timeseries prediction. Please provide a number of rows greater or equal to the window size. If you can\'t get enough rows, consider lowering your window size. If you want to force timeseries predictions lacking historical context please set the `allow_incomplete_history` advanced argument to `True`, but this might lead to subpar predictions.') # noqa
+    if len(combined_df) < n_groups * tss.window:
```
George3d6, line 136: will this actually work though? As in, shouldn't we test to see that forecasting with incomplete history doesn't result in a crash at the encoder level. I'm still not 100% up to date on the RNN encoder design so maybe if it's "obvious" it would work, forget about the test, since it's kind of an edge case anyway.

Also `log.warning` rather than `log.info`... maybe (?)

paxcema: The encoders should be able to handle it, because by this point the data has been reshaped into a regular size (even if the historical context has lots of `None`s). I will add a test though, to be sure.

lightwood/api/json_ai.py
```diff
@@ -177,8 +178,7 @@ def generate_json_ai(
 
         }]
 
-        if not problem_definition.timeseries_settings.is_timeseries or \
-                problem_definition.timeseries_settings.nr_predictions <= 1:
+        if not tss.is_timeseries or tss.nr_predictions <= 1:
```
hakunanatasha, line 181: when will ``tss.nr_predictions`` ever be less than 1? Might as well do == or 2?

lightwood/api/json_ai.py
```diff
@@ -270,6 +266,7 @@ def generate_json_ai(
     elif list(outputs.values())[0].data_dtype in [dtype.array]:
```
hakunanatasha, line 266: seems like `list(outputs.values())[0].data_dtype` is called earlier. might as well make this a var as otherwise since it's gonna get called. I don't know if this is expensive or a moot point to do tho but it does reduce the number of times we need to evaluate this.

hakunanatasha (approved):
Minor comment on lw.api.json_ai but otherwise logic seems fine.

lightwood/api/json_ai.py
paxcema: Good point, will go with `==`

PR #528, opened 2021-09-17
Deployment

George3d6:
Solves the deployment issue :)


George3d6:
Hard to read but only thing the PR does is:

Moves `docs` to `docssrc`
Adds a simple `make github` command to the makefile
This `make github` command moves `build/html` to `docs` which github automatically turns into the page at: https://mindsdb.github.io/lightwood/ (currently using the `github_docs` branch, but in the future, I'd use the `stable` branch)
The `make github` thing we can do manually when committing to stable *or* we can automate via github actions

hakunanatasha (approved):
i don't know why the tests are running but so far no complaints. I'll need to later merge the tutorials but this is fine

PR #530, opened 2021-09-17
Add `dtype.tsarray`

### LINKED ISSUE #484: Incorrect assumption for arrays
* Lightwood version: 1.1 / staging

Some bits of time series logic assume a column with `dtype.array` holds floats, but this is not necessarily always the case. Example:

```
[...]

elif data_dtype in (dtype.float, dtype.array):
    data[name] = data[name].astype(float)

[...]
```

### LINKED ISSUE #504: Dtypes target refers to dtype.array in if and elif statement
https://github.com/mindsdb/lightwood/blob/96b4dd64aa3f55d2c66db145d5151df1253cf21d/lightwood/data/statistical_analysis.py#L85

The "elif" component of the loop (89) will not be triggered due to line 85; this needs to be rectified.

paxcema:
## Why

To fix #484 and #504, and establish clearer dtype distinction between actual (non-temporal) arrays, and processed numerical time series.

## How
- Add `dtype.tsarray`, changing where needed (e.g. in `JsonAi`, timeseries transformer, analyzer)
- Fix logic in `StatisticalAnalyzer` to fix #504 
- Move time series encoder to its own type encoder mapping


George3d6:
Given that this has *a lot* of changes please feel free to branch off `airline_sentiment_fix` but PR directly into staging, as we'd ideally merge `airline_sentiment_fix` and then this, rather than consolidate into an even greater PR.

lightwood/analysis/model_analyzer.py
```diff
@@ -50,7 +50,7 @@ def model_analyzer(
     data_type = dtype_dict[target]
     data_subtype = data_type
 
-    is_numerical = data_type in [dtype.integer, dtype.float] or data_type in [dtype.array]
+    is_numerical = data_type in [dtype.integer, dtype.float, dtype.array, dtype.tsarray]
```
hakunanatasha, line 53: is there a reason this is a "[]" as opposed to "()"; earlier convention had tuples but this wont change anything

lightwood/encoder/__init__.py
```diff
@@ -17,10 +17,10 @@
 from lightwood.encoder.categorical.multihot import MultiHotEncoder
 from lightwood.encoder.text.pretrained import PretrainedLangEncoder
 from lightwood.encoder.type_encoder_maps import (Array, Binary, Categorical, Date, Datetime, Float, Image, Integer,
-                                                 Quantity, Rich_Text, Short_Text, Tags)
+                                                 TimeSeries, Quantity, Rich_Text, Short_Text, Tags)
 
 
-# Encoders that depend on optiona dependencies
+# Encoders that depend on optional dependencies
```
hakunanatasha, line 23: nice 8)

hakunanatasha (approved):
Is fine; 

analysis.model_analyzer and analysis.nc.util has a [] type instead of a () type for the data_dtype in [list_of_types] which breaks convention in some other points. This doesn't change anything as the order is not sensitive but for style it would be nice to stick to one convention.

PR #532, opened 2021-09-20
Use correct accuracy score for binary data

George3d6:
This fixes a bunch of issues:

1. We had a weird "fallback" behaviour when we couldn't find an accuracy function for a type, where we just used `accuracy_score`. This can result in bugs and unexpected behaviour, so now if an accuracy function for an output type is not found or specified by the user, we'll raise
2. Made `balanced_accuracy_score` the default for `binary` and `r2_score` the default for `date` and `datetime`
3. Fixed the `call` method used to build the json ai, it had a deprecated argument that took the entire json ai
4. Fixed the `call` method to raise an exception instead of silently fail is a wrong value is provided (by "wrong" here I mean "has potential for code injection so we don't allow it")
5. Fixed an issue where if the encoder was a target *and* it had to be provided the target class distribution it's previous arguments were overridden (when constructing the JSON ai)
6. Modified tests to check if the `unbias` flag works both for it's `True` and it's `False` values
7. Unbiasing target (`unbiase_target = True`) as the default, since it seems more reasonable to do that if our current assumption in the benchmarks is that balanced_accuracy is what matters


hakunanatasha (approved):

lightwood/helpers/templating.py
```diff
@@ -69,8 +68,13 @@ def is_allowed(v):
     return True
 
 
-def call(entity: dict, json_ai: JsonAI) -> str:
-    args = [f'{k}={v}' for k, v in entity['args'].items() if not str(v).startswith('$') and is_allowed(v)]
+def call(entity: dict) -> str:
```
paxcema, line 71: `json_ai` argument was removed because it was unused, yes?

paxcema: Nevermind, it's in the description 😛

paxcema (approved):

hakunanatasha (commented):
Why are leaving unbias_target default as True but then flagging as False later?

PR #534, opened 2021-09-21
Rename model to mixer and folds to subsets

### LINKED ISSUE #513: Rename model to mixer

### LINKED ISSUE #520: Rename 'folds' to 'subsets' across codebase
*Folds* has a different academic connotation than is typically used in our data. Please rename all calls to 'folds' as 'subsets'.

Later, I would like to explicitly reserve this for a user to call a "train", "val", "dev" and "test" subset (this is handled internally but not in an obvious way) for increased transparency.

George3d6:
Resolves #513, resolves #520


lightwood/encoder/text/helpers/rnn_helpers.py
```diff
@@ -34,7 +34,7 @@
 to sequence network <https://arxiv.org/abs/1409.3215>`__, in which two
 recurrent neural networks work together to transform one sequence to
 another. An encoder network condenses an input sequence into a vector,
-and a decoder network unfolds that vector into a new sequence.
```
paxcema, line 37: This one should not be changed

tests/integration/basic/test_model_selection.py
```diff
@@ -5,41 +5,41 @@
 from lightwood.api.types import ProblemDefinition
 
 
-class TestModelSelection(unittest.TestCase):
-    def get_models(self, df: pd.DataFrame, target: str, prob_kwargs: dict = None):
+class TestmixerSelection(unittest.TestCase):
```
paxcema, line 8: Typo: `TestMixerSelection`

paxcema (changes requested):
A few minor observations, apart from that this looks OK to me.

paxcema (approved):

lightwood/analysis/nc/nc.py
```diff
@@ -251,7 +251,7 @@ def score(self, x, y=None):
         return norm
 
 
-class BaseModelNc(BaseScorer):
+class BaseMixerNc(BaseScorer):
```
hakunanatasha: in theory, can analyzers work on encoder models? this is just to see if we can keep the basemodel nomenclature here, because any *model* can be analyzed, but mixers specifically go after the target.

either way this isnt an issue, just a comment.

lightwood/api/encode.py
```diff
@@ -4,11 +4,11 @@
...
-def encode(encoders: List[BaseEncoder], folds: List[pd.DataFrame], target: str) -> List[EncodedDs]:
-    if isinstance(folds, pd.DataFrame):
-        folds = [folds]
+def encode(encoders: List[BaseEncoder], subsets: List[pd.DataFrame], target: str) -> List[EncodedDs]:
+    if isinstance(subsets, pd.DataFrame):
+        subsets = [subsets]
 
     encoded_ds_arr: List[EncodedDs] = []
-    for fold in folds:
+    for fold in subsets:
```
hakunanatasha, line 12: maybe nice to call this "s" instead of fold but this is also fine.

lightwood/api/json_ai.py
```diff
@@ -105,56 +105,56 @@ def generate_json_ai(type_information: TypeInformation, statistical_analysis: St
         is_target_predicting_encoder = True
 
     if is_target_predicting_encoder:
-        models = [{
+        mixers = [{
             'module': 'Unit',
             'args': {
                 'target_encoder': '$encoders[self.target]',
-                'stop_after': '$problem_definition.seconds_per_model'
+                'stop_after': '$problem_definition.seconds_per_mixer'
```
hakunanatasha, line 112: this is so much clear in that seconds_per_mixer only affects mixers

hakunanatasha (approved):
This look fine to me - all the appropriate changes are made as far as I can tell (is there anything in lightwood/helpers??)

I want to change some elements of json-ai, but I think this PR should go through first before I do so. Additionally, I assume there will be changes to splitter hence had no comments there.

lightwood/analysis/nc/nc.py
paxcema: This particular approach (conformal prediction) should work for any object that generates a prediction given some input. If, say, you used a text encoder as the whole model, then this should work provided an appropriate interface is in place.

lightwood/api/encode.py
George3d6: should be `subset` will rename, not sure why it's this way rn

Issue #516, opened 2021-09-15
Modularize analyzer


paxcema:
Currently, the analyzer is the biggest phase of the pipeline, and it is not easily customizable if the idea is to extend it rather than replace it. It should be refactored into smaller units (ideally actors for better parallelization).

PR #539, opened 2021-09-22
Modular Analysis

paxcema:
## Why

To facilitate an easier extension of the core model analysis and explainability procedures with custom user-made "analysis blocks" (see #516).

## How
Introduces `BaseAnalysisBlock` class to encapsulate related analysis and inference logic under the same object. These blocks can have both types of logic (or either one of them). This PR modularizes the old `analysis` and `explain` procedures into sequential executions of core and user-defined blocks instead. All core logic has been moved to the new analysis block paradigm (see `analysis/nc/calibrate`, `analysis/helpers/feature_importance` and `analysis/helpers/acc_stats`).

There's also a bunch of restructuring in the rest of `analysis` files.


lightwood/analysis/helpers/feature_importance.py
```diff
@@ -0,0 +1,59 @@
...
+import torch
+import numpy as np
+import pandas as pd
+
+from lightwood.analysis.base import BaseAnalysisBlock
+from lightwood.helpers.general import evaluate_accuracy
+from lightwood.analysis.nc.util import t_softmax
+
+
+class GlobalFeatureImportance(BaseAnalysisBlock):
```
hakunanatasha, line 14: please potentially annotate what this function does - people do care about how column importances are calculated

lightwood/encoder/base.py
```diff
@@ -28,7 +28,7 @@ def encode(self, column_data) -> torch.Tensor:
     def decode(self, encoded_data) -> List[object]:
         raise NotImplementedError
 
-    # Should work for all troch-based encoders, but custom behavior may have to be implemented for very weird models
+    # Should work for all torch-based encoders, but custom behavior may have to be implemented for weird models
```
hakunanatasha, line 31: <3 thank you

hakunanatasha (approved):
looks fine to me so long as george approves

this is a fantastic streamlined approach!

minor note- major thanks for making doc strings + typo fixing

lightwood/analysis/analyze.py
```diff
@@ -103,7 +104,13 @@ def model_analyzer(
...
-    for block in analysis_blocks:
-        runtime_analyzer = block.compute(runtime_analyzer, **{})
+    if len(analysis_blocks) > 0:
+        exec(lightwood.api.json_ai.IMPORTS_FOR_EXTERNAL_DIRS, globals())
+        exec(lightwood.api.json_ai.IMPORT_EXTERNAL_DIRS, globals())
+
+        for dirpath in analysis_blocks:
+            module, block_name = dirpath.split(".")
+            block = getattr(eval(module), block_name)()
+            runtime_analyzer = block.analyze(runtime_analyzer, **{})
```
George3d6, line 114: Ok, I think this is a really bad idea to be honest.

As in, this is basically moving dynamic code generation from json_ai, where it's at least constrained, into the model_analyzer.

If we go the route of "every module can do dynamic importing" (outside of very specific cases, e.g. analyzer can dynamically import sklearn acc function from strings) then the design can get messy.

Why not have the user's specify the blocks in the json_ai and import them at json_ai creation time?

lightwood/analysis/explain.py
```diff
@@ -80,7 +81,13 @@ def explain(data: pd.DataFrame,
...
-    for block in explainer_blocks:
-        row_insights, global_insights = block.explain(insights, **{})
+    if len(explainer_blocks) > 0:
+        exec(lightwood.api.json_ai.IMPORTS_FOR_EXTERNAL_DIRS, globals())
+        exec(lightwood.api.json_ai.IMPORT_EXTERNAL_DIRS, globals())
+
+        for dirpath in explainer_blocks:
+            module, block_name = dirpath.split(".")
+            block = getattr(eval(module), block_name)()
+            row_insights, global_insights = block.explain(insights, **kwargs)
```
George3d6, line 91: Same comment as above, don't think this should be happening here.

George3d6: I'd be better for `explainer_blocks` to be a list of already existing objects that's e.g. instantiate when `class Predictor` is instantiated and inlined into the predictor code by `json_ai.py`

George3d6 (changes requested):

lightwood/analysis/analyze.py
paxcema: Very good point, let's do that instead.

tests/integration/advanced/test_custom_modules.py
```diff
@@ -62,3 +62,45 @@ def test_0_add_throwing_cleaner(self):
...
+            'args': {}
+        }]
+
+        json_ai = JsonAI.from_dict(json_ai_dump)
+
+        # create a predictor from it
+        code = code_from_json_ai(json_ai)
+        predictor = predictor_from_code(code)
+        predictor.learn(df)
+        row_insights = predictor.predict(df)
```
George3d6, line 103: I thought this test had a `.save` and a `.load` called on the predictor object (and hence that's why you spotted the import bug)... anyway, outside the scope of this test or this PR to do this but made this issue: https://github.com/mindsdb/lightwood/issues/556 | so thanks for pointing this out :) \

George3d6 (changes requested):
- The `sys.modules` bit is fine, if we run into namespace collision issues with many users we can add a custom_modules. "namespace" to the custom modules for now this is more streamlined
- Would have failed in all cases so nice catch, I was not trying saving and loading in my test, hindsight I should have
- The handling of the list looks all good, that's what I'd have done, will review the logic now

tests/integration/advanced/test_custom_modules.py
paxcema: Yeah, caught it with an external test script that did `.save()` and `.load()`. I agree on following up with a PR for #556.

tests/integration/advanced/test_custom_modules.py
```diff
@@ -62,3 +62,45 @@ def test_0_add_throwing_cleaner(self):
...
+
+class {cname}(BaseAnalysisBlock):
+    def __init__(self):
+        super().__init__(deps=None)
+
+    def analyze(self, info, **kwargs):
+        info['test'] = 'test'
+        return info
+
+    def explain(self, row_insights, global_insights, **kwargs):
```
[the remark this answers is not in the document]
paxcema: We don't. I tried doing it (and works) but I thought we should first discuss if it makes sense for predict to return that. Relevant issue is #557

George3d6 (approved):

PR #540, opened 2021-09-23
Better imports

### LINKED ISSUE #515: Make directory for storing custom modules
As a replacement for the current `JsonAI.imports` mechanism, Lightwood should create a directory where all custom ML blocks are stored (potentially following the same structure as core Lightwood). This directory would be turned into a Python module so that at runtime Lightwood automatically imports any custom block that is discovered, streamlining the process.

George3d6:
* Removed import key from Json AI
* Import everything that we may want to use in the top-level predictor by default
* Custom modules are imported by default from `~/lightwood_modules` and `/etc/lightwood_modules` if they exist (maybe add an extra path for windows?)

Resolves #515


tests/integration/advanced/test_custom_modules.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+test_err_message = 'This ! Is ! A ! Testing ! Error !'
+
+
+def create_custom_module():
+    mdir = os.path.expanduser('~/lightwood_modules')
+
+    try:
+        shutil.rmtree(mdir)
```
paxcema, line 16: Wouldn't this test remove all custom modules a user might have added?

George3d6: Yeap, we should probably only remove the file we create,, good point.

hakunanatasha: Might be worth considering a "refresh modules" by removing anything custom function. Either way, I like it.

Dumb question, how do users import custom accuracy functions as in if they want to do ```from sklearn.metrics import hinge_loss``` with this new approach?

hakunanatasha (approved):
General approval on the strategy, let's chat in standup.

The only caveat I see is that I don't follow how a user can bring a custom metric imported from another package but I may not be reading too carefully.

tests/integration/advanced/test_custom_modules.py
George3d6: Any accuracy function gets auto-imported:
`accuracy_function = getattr(importlib.import_module('sklearn.metrics'), accuracy_function_str)`

PR #542, opened 2021-09-23
Data splitting and passing architecture refactor

George3d6:
Switched from lightwood operating with subsets to it operating with 3x dataframes / EncodedDses (train, dev and test).
Added stratification procedure for categorical and binary targets when splitting.


lightwood/data/splitter.py
```diff
@@ -16,7 +16,7 @@ def splitter(data: pd.DataFrame, k: int, tss: TimeseriesSettings,
 
     if not tss.is_timeseries:
         # shuffle
-        data = data.sample(frac=1, seed=seed if seed is not None else len(data)).reset_index(drop=True)
+        data = data.sample(frac=1, random_state=seed if seed is not None else len(data)).reset_index(drop=True)
```
hakunanatasha, line 19: honestly just make seed = 1 in the kwargs default this is too convoluted

hakunanatasha: I'm not sure why you want it to be the len(data) when you can just as easily do something like

```def splitter(...., seed=1):```

**Review by hakunanatasha — COMMENTED** [review]
Approved but please dont make the seed len(data) if None, just literally set the seed=1 or 420 or some arbitrary number. It doesn't need to change per dataset and not per len(data). It's easier for a user to also reproduce and it removes the if/else in the random_state situation

File: lightwood/data/splitter.py (hunk 1 above)
Comment (hakunanatasha) [reply lightwood/data/splitter.py]: As an aside - I've made some other changes to this file @George3d6

File: lightwood/data/splitter.py [hunk 2]
```diff
@@ -1,48 +1,84 @@
...
         data = data.sample(frac=1, random_state=seed).reset_index(drop=True)
 
-        # split
-        subsets = np.array_split(data, k)
-
+    stratify_on = None
+    if tss.is_timeseries or dtype_dict[target] in (dtype.categorical, dtype.binary) and target is not None:
+        stratify_on = [target]
+        if isinstance(tss.group_by, list):
+            stratify_on = stratify_on + tss.group_by
```
paxcema, line 47: I believe this will impact the stratification for TS tasks. At the moment it is only done using `tss.group_by`. May be worth trying, but I'd default to doing `stratify_on = tss.group_by`.

George3d6: I assume timeseries tasks with categorical targets are rare, in the cases they exist, it seems reasonable to also stratify on that

paxcema: Fair point, that is nice. However, does this work for numerical TS? Unless I'm reading this wrong, it would still stratify on the target then, which I'm not sure is correct.

lightwood/data/splitter.py
```diff
@@ -1,48 +1,107 @@
...
+) -> Dict[str, pd.DataFrame]:
     """
+    Splits a dataset into stratified training/test. First shuffles the data within the dataframe (via ``df.sample``).
+
+    :param data: Input dataset to be split
+    :param tss: time-series specific details for splitting
+    :param pct_train: training fraction of data; must be less than 1
+    :param dtype_dict: Dictionary with the data type of all columns
+    :param seed: Random state for pandas data-frame shuffling
+    :param n_subsets: Number of subsets to create from data (for time-series)
```
paxcema, line 29: remove, deprecated

Issue #536, opened 2021-09-22
Makr encoder as trainable and timesereis


George3d6:
Currently, there are hardcoded lists of trainable and time-series encoders, which indicate that we ought to pass an extra argument to them. Instead, we should handle this in a way that makes it easier for the user to add a custom ts or trainable encoders.

Ideas:

1. Add `trainable` and `timeseries` bool flags to encoders ( default to `False` in `BaseEncoder`)
2. Introspect signature of encoder class and mark as trainable or timeseries if the appropriate arguments are found.

hakunanatasha:
@paxcema and I discussed; it's almost better to look for an attribute passed to the encoder called "is_trained" instead. This enables the processing pipeline to look for data passed in to adjust. We may want to even consider if you want to train an encoder on NEW data that has nothing to do with the target.

George3d6:
oh yes, ****, I remember, I was in that discussion. But I guess an issue never came out of it? Do you still have any notes from that?

George3d6:
Note: We have to be able to determine this *before* we instantiate the encoders, since this dictates how they are instantiated. So I'm afraid we'll have to add 2x new class variables `is_timeseries` and `is_trainable` to the base encoder... hopefully static members can work around this issue.

hakunanatasha:
We didn't make an issue because we had decided that we needed to visit the encoder design together as opposed to figuring it out this week. 

That's fine - I'm actually a fan of these two flags.

Issue #514, opened 2021-09-15
Inherit default module arguments in JsonAI


paxcema:
When developing and using new custom modules in JsonAI, the default arguments should be implicitly included, so that instead of writing this:

```python
json_ai.analyzer = {
        'module': 'shap_model_analyzer',
        'args': {
            'stats_info': '$statistical_analysis',
            'predictor': '$ensemble',
            'data': 'test_data',
            'train_data': 'train_data',
            'target': '$target',
            'dtype_dict': '$dtype_dict',
            'new_custom_arg1': 'x'
        }
    }
```

You can opt for:

```python
json_ai.analyzer = {
        'module': 'shap_model_analyzer',
        'args': {
            'new_custom_arg1': 'x'
        }
```

Or similar. Note how all of the original arguments are used in the default model analyzer. Thus, this custom analyzer will use a subset of them in addition to a new argument. Default arguments should be implicitly retrieved and passed to the custom module if needed.

hakunanatasha:
Strongly agree.

PR #543, opened 2021-09-24
Passing default arguments

George3d6:
If the user overrides a hidden module, but doesn't pass all of the arguments **and** if we detect some of the argument names for this module are the same as the "default" argument names, we'll implicitly pass them.

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -369,95 +369,88 @@ def add_implicit_values(json_ai: JsonAI) -> JsonAI:
                 json_ai.features[name].encoder['module'].split(".")[0].lower()
             )
 
-    # Add implicit phases
-    # @TODO: Consider removing once we have a proper editor in studio
-    if json_ai.cleaner is None:
-        json_ai.cleaner = {
-            "module": "cleaner",
-            "args": {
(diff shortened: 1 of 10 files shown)


lightwood/api/json_ai.py
```diff
@@ -369,96 +404,94 @@ def add_implicit_values(json_ai: JsonAI) -> JsonAI:
...
+
+    for field_name, implicit_value in hidden_fields:
+        field = json_ai.__getattribute__(field_name)
+        if field is None:
+            if tss.is_timeseries or field_name not in ('timeseries_analyzer', 'timeseries_transformer'):
+                field = implicit_value
+        else:
+            args = eval(field['module']).__code__.co_varnames
+            for arg in args:
+                if 'args' not in field:
```
hakunanatasha, line 488: I'd prefer blocking these up into small sub-functions but the gist of this is clear

hakunanatasha (approved):
Looks good - the only caveat is that I tend to prefer making smaller functions so you call blocks ex:
```
for i, j in values:
        if is_timeseries:
                _do_some_ts_function(<args>):
        if args in j:
                _do_some_other_fun(<args.)

George3d6:
Closes #536 
Closes #514

PR #546, opened 2021-09-27
Allow quantity as target

### LINKED ISSUE #545: Mixers don't support quantitiy target
I think this applies to all of them, we should introduce it, basically just a number.

George3d6:
Allows `quantitiy` to be the dtype of the target column, modifies mixers, statistical analysis and model analysis modules to account for this.

Adds test for this case.

Closes #545


paxcema (approved):
General approve, but it seems like the test is missing?

lightwood/analysis/explain.py
```diff
@@ -223,7 +223,7 @@ def explain(data: pd.DataFrame,
         insights['prediction'] = insights['prediction'].astype(int)
         insights['upper'] = insights['upper'].astype(int)
         insights['lower'] = insights['lower'].astype(int)
-    elif target_dtype in (dtype.float):
+    elif target_dtype in (dtype.float, dtype.quantity):
```
hakunanatasha, line 226: im ok with quant being a float but what sort of things are taken as a quant INSTEAD of float/int in use cases?

hakunanatasha (approved):
In general approved but one major high level question is that what data is categorized as quantity? from `data.infer_types` it seems like its when its not date-time/date or i guess numeric etc.

The reason I ask is because I want to know why we cast quantity as a float (which IS reasonable and im OK with that default) as opposed to an int? What are the cases (if any) that may be beneficial to have an int?

With questions answered this can be merged.

Issue #547, opened 2021-09-27
Fix OneHotEncoder in JSONAI if fewer than 100 (N) categories


hakunanatasha:
Right now we call CatAE; under the hood, this defaults to OHE when under 100 categories, but the logic can be moved outside to make things more simple.

George3d6:
getting this done today or tomorrow

PR #549, opened 2021-09-27
Select OHE at json-ai build time and fix pandas None bugs

George3d6:
Categorical AE and implicitly Short text encoder no longer switch to OHE mode for < 100 categories

Fixed bugs where pandas is replacing `None` with various stuff that's something else by aggressively checking all values pandas might replace `None` with and converting them back to `None` at cleaning time.

Addresses #547

### DIFF
--- a/lightwood/data/cleaner.py
+++ b/lightwood/data/cleaner.py
@@ -54,7 +54,7 @@ def cleaner(
         # If you want to customize the cleaner, it's likely you can to modify ``get_cleaning_func``
         for nan_val in VALUES_FOR_NAN_AND_NONE_IN_PANDAS:
             data[col] = data[col].apply(get_cleaning_func(dtype_dict[col], custom_cleaning_functions)
-                                    ).replace({nan_val: None})
+                                        ).replace({nan_val: None})
         # If a column has too many None values, raise an Excpetion
         # Figure out how to reintroduce later, maybe a custom flag, `crash for too much invalid data`?
         # _check_if_invalid(data[col], pct_invalid, col)
--- a/lightwood/data/cleaner.py
+++ b/lightwood/data/cleaner.py
@@ -15,6 +15,9 @@ import numpy as np
 from typing import Dict, List, Optional, Tuple, Callable, Union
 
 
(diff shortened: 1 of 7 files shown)


lightwood/data/cleaner.py
```diff
@@ -49,12 +52,13 @@ def cleaner(
...
-        data[col] = data[col].apply(get_cleaning_func(dtype_dict[col], custom_cleaning_functions)
-                                    ).replace({np.nan: None})
+        for nan_val in VALUES_FOR_NAN_AND_NONE_IN_PANDAS:
+            data[col] = data[col].apply(get_cleaning_func(dtype_dict[col], custom_cleaning_functions)
+                                        ).replace({nan_val: None})
         # If a column has too many None values, raise an Excpetion
         # Figure out how to reintroduce later, maybe a custom flag, `crash for too much invalid data`?
         # _check_if_invalid(data[col], pct_invalid, col)
-
+    pd.set_option("display.max_rows", None, "display.max_columns", None)
```
paxcema, line 61: Why is this being set?

lightwood/encoder/categorical/autoencoder.py
```diff
@@ -24,18 +22,15 @@ def __init__(self, stop_after: int = 3600, is_target: bool = False, max_encoded_
...
-        if self.is_target:
-            self.max_encoded_length = None
-        else:
-            self.max_encoded_length = max_encoded_length
+        self.max_encoded_length = max_encoded_length
 
     def _train_callback(self, error, real_buff, predicted_buff):
-        log.info(f'{self.name} reached a loss of {error} while training !')
+        pass
+        #  log.info(f'{self.name} reached a loss of {error} while training !')
```
paxcema, line 33: Leftover comment, though I think it may be useful to leave this one as a `log.debug()` call

paxcema (approved):

lightwood/encoder/categorical/autoencoder.py
George3d6: whole function can be removed tbh, a debug log would be useful but I'll leave it w/o to motivate us to phase out the Gym when we get to resisting this encoder.

PR #563, opened 2021-10-04
Restore icp args

### LINKED ISSUE #562: Restore arguments for ICP customization
The ICP analysis block gave the option to customize its behavior during prediction, which was specially useful for the anomaly detection pipeline.

Currently, these arguments are only using the (static) value provided inside the problem definition:

* `anomaly_cooldown`
* `anomaly_error_rate`
* `anomaly_detection`

We should fix it so that they can be dynamically modified once a predictor has been trained.

paxcema:
## Why 

To fix #562.

## How

Introduces a `PredictionArguments` dataclass that aims at generating defaults and setting any user-defined arguments for a `predict()` call. A couple statements are added in existing tests to account for the correct setting of these arguments.

Also, fixes a bug in the ICP block where the `fixed_confidence` setting would be incorrectly overridden in grouped TS tasks.

### DIFF
--- a/lightwood/api/types.py
+++ b/lightwood/api/types.py
@@ -298,16 +298,10 @@ class ProblemDefinition:
         encoders and models.
     :param target_weights:
     :param positive_domain: For numerical taks, force predictor output to be positive (integer or float).
-    :param fixed_confidence: For analyzer module, specifies a fixed `alpha` confidence for the model calibration so \
-        that predictions, in average, are correct `alpha` percent of the time.
     :param timeseries_settings: TimeseriesSettings object for time-series tasks, refer to its documentation for \
          available settings.
     :param anomaly_detection: Whether to conduct unsupervised anomaly detection; currently supported only for time-\
         series.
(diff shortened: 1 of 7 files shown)


lightwood/api/types.py
```diff
@@ -542,3 +527,54 @@ class ModelAnalysis:
     confusion_matrix: object
     histograms: object
     dtypes: object
+
+
+@dataclass
+class PredictionArguments:
```
George3d6, line 533: Can we remove `predict_proba` and instead add it as an argument here? It's fine if the `predict` interface changes to some extent we can always have the return type be a `Union` of the various types returned by the various modes.

George3d6 (changes requested):

lightwood/api/types.py
George3d6: Also can you add a "all_mixers" mode or something like that which returns a prediction for every mixer? We needed it for the benchmarks anyway

paxcema: Both are really good ideas, let me take a stab at them.

PR #569, opened 2021-10-05
Add Module type

George3d6:
`JsonAI` and `Feature` now correctly use the type `Module` for the modules they contain, should behave exactly as the dictionaries we used before


paxcema (approved):
Looks good :+1:

PR #597, opened 2021-10-09
changes: Reduce spacing and beautify the code in high_level.html #596

idivyanshbansal:
#596

### DIFF
--- a/docs/api/high_level.html
+++ b/docs/api/high_level.html
@@ -1,342 +1,329 @@
-
-
 <!DOCTYPE html>
-<html class="writer-html5" lang="en" >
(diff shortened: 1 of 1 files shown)


paxcema:
Hey @idivyanshbansal, our docs are auto-generated so I'm afraid this would get overridden in the next release. As mentioned in #590 among others, feel free to open a **_single_** issue with the **_discussion_** tag to talk about this.

Moreover, if you are looking to participate in the Hacktoberfest 2021 event, be on the lookout for issues with the `hacktoberfest` tag, as only those count. Cheers!

Issue #601, opened 2021-10-10
 Reduce spacing and beautify the code in api folder


idivyanshbansal:
There is a requirement that will clean all unwanted spacing in the code that will reduce the page size. Beautifying the code will make it easy for a developer to read the code and work on it. Even sir formatting code can also make changes in server response time too. Better formatting means less server response time.

paxcema:
Okay, 

As mentioned before, docs are currently auto-generated using sphinx. Feel free to suggest here any proposals regarding html spacing and formatting when automatically generating the docs.

PR #602, opened 2021-10-10
Changes: Reduce spacing and beautify the code in api folder

idivyanshbansal:
#601

### DIFF
--- a/docs/api/types.html
+++ b/docs/api/types.html
@@ -1,630 +1,796 @@
-
-
 <!DOCTYPE html>
-<html class="writer-html5" lang="en" >
+<html class="writer-html5" lang="en">
+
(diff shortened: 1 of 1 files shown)


idivyanshbansal:
> Hi, thanks for the PR, but I'm afraid the docs directory is auto-generated from the files inside dossrc
> 
> If you think beautified HTML would help make the docs more readable (maybe some people like to read the source) feel free to suggest ways of making sphinx better at generating pretty HTML.

Thank you for reviewing @George3d6 sir. I have beautified using brackets which works better than VS code for this job. There should  NOT be an issue with functioning. Moreover, I see that on my website when code is minified it increases the loading time by a little bit.

PR #604, opened 2021-10-10
Add-vaibhavi-branch

alteregoprofile:
# Why is it needed ?
To ensure that incase we require something to be returned as it is, we can have it decoded
# What does it do ?
Goes from list of arrays to tensors and tensors to list of arrays.


alteregoprofile:
I have attempted this but be warned I am new to all of this.

George3d6:
Also, fyi, the commits are made by this account: https://github.com/vdeodataseer | which is different than the one you're commenting from, not sure if intended.

Issue #614, opened 2021-10-11
Implement an audio encoder


paxcema:
## Task

Lightwood handles many different data types, and we're always looking to include more.

One possible data type to experiment with is an [audio encoder](https://github.com/mindsdb/lightwood/blob/staging/lightwood/encoder/audio/amplitude_ts.py); if you are interested in tackling this, _please_ ping the research team (either me, @hakunanatasha or @George3d6) in the discussion here, to propose your initial approach and scope your contribution together ✨. 

Ensure you implement a unit test plus any modifications needed to obtain an encoded representation from audio in order for this PR to be accepted.

## Steps :male_detective: :female_detective: 

- Fork the Lightwood repository, checkout the `staging` branch and from it create a new one.
- Implement a unit test that generates an instance of your encoder, some (potentially synthetic) audio data, and proceed to encode-decode it. [This](https://github.com/mindsdb/lightwood/blob/f469fd49f1859bd0444094fd378c9508e99871cf/tests/unit_tests/encoder/time_series/test_timeseries_rnn.py#L22) unit test can be useful as a reference, and the (outdated) [`AmplitudeTs` encoder](https://github.com/mindsdb/lightwood/blob/staging/lightwood/encoder/audio/amplitude_ts.py) as inspiration or starting point for your own encoder (in fact, feel free to replace it).
- Accuracy of the encoder is to be discussed with our team, so don't get discouraged if encoded representations are not perfect; this is a hard problem!
- Make the PR and address any comments that reviewers might make.

## Additional rewards :1st_place_medal: 

Each encoder PR brings 5️⃣  points for entry into the draw for a :computer:  Deep Learning Laptop powered by the NVIDIA RTX 3080 Max-Q GPU or other swag :shirt: :bear: .  For more info check out https://mindsdb.com/hacktoberfest/

paxcema:
Hey @mrandri19, that's great! 

By all means, feel free to work on this. I will assign you to this issue preemptively, though note that others can also participate, concurrently.

As stated in the original post, we look forward to any questions or implementation ideas you might have. Good luck!

hakunanatasha:
Looking forward : ) feel free to ping one of us (or all of us) if you need anything - we're happy to help!

paxcema:
That sounds nice! We're always open to discuss collaborations :)

[...]

PR #625, opened 2021-10-12
Implement MFCC-based audio encoder

mrandri19:
Related to #614 

# Why is it needed ?

To encode audio files into a vector form, to be used by models downstream.

# What does it do ?

Uses [librosa](https://librosa.org/doc/latest/index.html) to compute the Mel-frequency cepstral coefficients (MFCCs) of the audio file. They are a common feature used in speech and audio processing. The features are a 2D array, flattened into a 1D one.

## Parameter choices

We set the following two parameters:
```
NUM_TIME_BUCKETS = 100
N_MFCC_COEFFICIENTS = 20
```
The first one means that the audio file will be split into 100 sequential time intervals before computing the Fourier transform needed for the MFCCs. It was chosen because it will split a 1s audio file in 10ms intervals, which are enough for speech recognition, and because is will split a 3 minutes song in 1.8s intevals, which are still small enough to capture enough detail for genre recognition (At 128BPM, 1.8 seconds are 4 beats or 1 bar).

The second one is librosa's current default. We set it explicitly because we aim for `20 * 100 = 2000` elements in our feature vector.

## Does it work?

The encoder has a unit test, but I suppose that adding an audio benchmark wouldn't be a bad idea.

### DIFF
--- a/lightwood/encoder/audio/mfcc.py
+++ b/lightwood/encoder/audio/mfcc.py
@@ -5,6 +5,7 @@ import numpy as np
 import torch
 import warnings
 from lightwood.encoder.base import BaseEncoder
+from lightwood.helpers.io import read_from_path_or_url
 from lightwood.helpers.log import log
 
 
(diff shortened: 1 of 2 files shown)


George3d6:
Overall this looks amazing, just what we need, thanks a ton

Issue #674, opened 2021-10-22
Unexpected `nan` cleaning behavior


paxcema:
[This line](https://github.com/mindsdb/lightwood/blob/353e37970dbf610b14fdf932ee9b10eed194f5a0/lightwood/data/cleaner.py#L55) in `lightwood/data/cleaner.py`:

```python
data[col] = data[col].replace(to_replace=VALUES_FOR_NAN_AND_NONE_IN_PANDAS, value=None)
```

Has an unexpected behavior, where `nan`s are seemingly replaced with a forward fill mode instead of imputed with `None`s.

George3d6:
Closing, should be handled in #700 , doing this it too hard. You can replace `is None` checks with the `is_none` helper when referring to a cell from a pd Series.

PR #695, opened 2021-10-27
fixed ```nan``` cleaning behaviour #674

ma1VAR3:
This is in reference to the issue: 
#### Fix:
Changed the following line:
```data[col] = data[col].replace(to_replace=VALUES_FOR_NAN_AND_NONE_IN_PANDAS, value=None)```
to:
```data[col] = data[col].apply(lambda x: None if x in VALUES_FOR_NAN_AND_NONE_IN_PANDAS else x)```


George3d6:
Closing, should be handled in #700 , doing this it too hard. You can replace `is None` checks with the `is_none` helper when referring to a cell from a pd Series.

PR #707, opened 2021-10-30
Removed all usages of print, removed unused tests

George3d6:
* Replace all usage of `print` (in the lightwood codebase, not in the tests) with the lightwood's logger `debug` print
* Removed test that wasn't running and related helpers, realistically it wasn't needed anyway and adapting it would have been too much of a hassle. Also this means we are officially 100% decoupled from the mindsdb remote testing env :)


George3d6:
FYI, do not merge yet, I plan to add the fixes for datasets failing in 1.6 (stack overflow and us consumption) in this PR

PR #712, opened 2021-11-01
Docs - Add predictor update tutorial

paxcema:
Notebook briefly showcases `partial_fit()` interface in a quick example.


George3d6 (changes requested):
Force the ensemble to be `BestOf` , otherwise, this guide will not hold once we have other default ensembles.

paxcema:
Try to use `PredictorInterface.adjust()`

PR #715, opened 2021-11-02
Documentation for /most/ lightwood modules

George3d6:
We need to migrate to the latest sphinx to use something like autosummary in order to not document like we're in the stone age.

*But* until we can do that we need to at least display most of what's documented in the actual docs.

This adds the doctree for the `__init__.py` file of each top-level lightwood directory to the index.

It also removes some extra headings which caused really nasty hierarchies with 1 to 3 elements that had a single child.

### DIFF
--- a/docssrc/source/encoder.rst
+++ b/docssrc/source/encoder.rst
@@ -1,0 +1,7 @@
+:mod:`Encoders`
+==========================
+
+Used for encoding data into pytorch tensorsand decoding it from pytorch tensors
+
+.. automodule:: encoder
+   :members:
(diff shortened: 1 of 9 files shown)


docssrc/source/encoder.rst
```diff
@@ -0,0 +1,7 @@
+:mod:`Encoders`
+==========================
+
+Used for encoding data into pytorch tensorsand decoding it from pytorch tensors
```
paxcema, line 4: Typo: "tensorsand" -> "tensors and"

paxcema: Also, maybe "pytorch" -> "PyTorch"?

paxcema (approved):

PR #720, opened 2021-11-04
Making sure we don't decode nans

George3d6:
- Encoded DS now tests all the values that get encoded to make sure no `nans` or `infs` are present in the tensors
- SImplified the preparation of the `NumericEncoder` and removed old edge-case handling where the numeric encoder could accept string values
- Documented the `NumericEncoder` and all of its methods (constructor included)
- Added a test that checks if the `NumericEncoder` can handle encoding and being prepared with corrupt values (nan, inf, None) without outputting a `nans` or `infs` in the encoded representation, as well as if these invalid values are properly decoded to `None`
- Added tests similar above to the `DateTimeEncoder` as part of an existing test, much simpler since this encoder needs no preparation, so we only have to test `decode`


George3d6:
Detected some potential problems with timeseries, pinging @paxcema to take a look at this.

lightwood/data/encoded_ds.py
```diff
@@ -72,6 +72,8 @@ def __getitem__(self, idx: int) -> Tuple[torch.Tensor, torch.Tensor]:
 
                 data = self.data_frame[cols].iloc[idx].tolist()
                 encoded_tensor = self.encoders[col].encode(data, **kwargs)[0]
+                if torch.isnan(encoded_tensor).any() or torch.isinf(encoded_tensor).any():
```
hakunanatasha, line 75: just to make sure, there is no weird buggy edge cases on nan/infs? I always shiver when I think of isnan() and isinf() calls : /

lightwood/encoder/numeric/numeric.py
```diff
@@ -1,46 +1,63 @@
...
+    The ``absolute_mean`` is computed in the ``prepare`` method and is just the mean of the absolute values of all numbers feed to prepare (which are not none)
+
+    ``none`` stands for any number that is an actual python ``None`` value or any sort of non-numeric value (a string, nan, inf)
+    """ # noqa
+
+    def __init__(self, data_type: dtype = None, is_target: bool = False, positive_domain: bool = False):
+        """
+        :param data_type: The data type of the number (integer, float, quantity)
+        :param is_target: If we are encoding a target value or an input value (feature)
+        :param positive_domain: Forces the encoder to always output positive values
```
hakunanatasha, line 28: in what cases do we use this or need to use this?

hakunanatasha: it's helpful to explain why we do have this flag

lightwood/encoder/numeric/numeric.py
```diff
@@ -1,46 +1,63 @@
...
+from torch.types import Number
 from lightwood.encoder.base import BaseEncoder
 from lightwood.helpers.log import log
 from lightwood.helpers.general import is_none
+from lightwood.api.dtype import dtype
 
 
 class NumericEncoder(BaseEncoder):
+    """
+    The numeric encoder takes numbers (float or integer) and converts it into tensors of the form:
```
hakunanatasha, line 14: Would be nice to explain why we choose to do it this way. 

One interesting thing that comes out of it is that you now have features that point to modalities of *why* a prediction goes the way it does, but I don't actually know the rationale of why this is the case.

lightwood/encoder/numeric/numeric.py
```diff
@@ -1,46 +1,63 @@
...
+        """
+        :param data_type: The data type of the number (integer, float, quantity)
+        :param is_target: If we are encoding a target value or an input value (feature)
+        :param positive_domain: Forces the encoder to always output positive values
+        """
         super().__init__(is_target)
         self._type = data_type
         self._abs_mean = None
         self.positive_domain = positive_domain
         self.decode_log = False
```
hakunanatasha, line 34: what is decode_log used for and why dont other encoders have it?

lightwood/encoder/numeric/numeric.py
```diff
@@ -1,46 +1,63 @@
...
         self.output_size = 4 if not self.is_target else 3
 
-    def prepare(self, priming_data):
+    def prepare(self, priming_data: Iterable):
+        """
```
hakunanatasha, line 38: Provide a description. Can be as simple as

"NumericalEncoder" uses a rule-based form to prepare results on training (priming) data. The averages etc. are taken from *this* distribution.

Some people often normalize data using the stddev + mean of the *entire* dataset; by explicitly choosing the training, we should convey that decision.

lightwood/encoder/numeric/numeric.py
```diff
@@ -1,46 +1,63 @@
...
 from lightwood.encoder.base import BaseEncoder
 from lightwood.helpers.log import log
 from lightwood.helpers.general import is_none
+from lightwood.api.dtype import dtype
 
 
 class NumericEncoder(BaseEncoder):
+    """
+    The numeric encoder takes numbers (float or integer) and converts it into tensors of the form:
+    ``[is_none(number), is_positive(number), natural_log(abs(number)), number/absolute_mean]``
```
hakunanatasha, line 15: you might want to actually mention that is_none and is_positive are in fact boolean outputs of the number itself, maybe even call it inp_number for less ambiguity.

lightwood/encoder/numeric/numeric.py
```diff
@@ -1,46 +1,63 @@
...
+
+    def __init__(self, data_type: dtype = None, is_target: bool = False, positive_domain: bool = False):
+        """
+        :param data_type: The data type of the number (integer, float, quantity)
+        :param is_target: If we are encoding a target value or an input value (feature)
+        :param positive_domain: Forces the encoder to always output positive values
+        """
         super().__init__(is_target)
         self._type = data_type
         self._abs_mean = None
```
hakunanatasha, line 32: what happens if _abs_mean is 0? This should in theory NEVER happen unless the input column are all 0s.

I can't recall in our stats-analysis if we throw away columns with extremely low entropy (i.e. columns where there is no difference at all in the values of each row. I recall some entropy calculation there for identifiers which are the opposite side). @George3d6 we should add this if it doesn't exist.

lightwood/encoder/numeric/numeric.py
```diff
@@ -1,46 +1,63 @@
...
+                    value_type = 'float'
 
         self._type = value_type if self._type is None else self._type
-        non_null_priming_data = [float(str(x).replace(',', '.')) for x in priming_data if not is_none(x)]
+        non_null_priming_data = [x for x in priming_data if not is_none(x)]
         self._abs_mean = np.mean(np.abs(non_null_priming_data))
         self.is_prepared = True
 
-    def encode(self, data):
+    def encode(self, data: Iterable):
```
hakunanatasha, line 55: describe the process

lightwood/encoder/numeric/numeric.py
```diff
@@ -80,7 +94,13 @@ def encode(self, data):
 
         return torch.Tensor(ret)
 
-    def decode(self, encoded_values, decode_log=None) -> list:
+    def decode(self, encoded_values: Union[List[Number], torch.Tensor], decode_log: bool = None) -> list:
```
hakunanatasha, line 97: describe the process

lightwood/mixer/regression.py
```diff
@@ -12,6 +12,8 @@
 
 
 class Regression(BaseMixer):
+    '''
```
hakunanatasha, line 15: might conflict with george's docs

tests/integration/advanced/test_timeseries.py
```diff
@@ -51,10 +51,10 @@ def calculate_duration(self, predictor, train, time_aim_expected):
         start = time.process_time()
         predictor.learn(train)
         time_aim_actual = (time.process_time() - start)
-        if((time_aim_expected * 5) < time_aim_actual):
+        if((time_aim_expected * 10) < time_aim_actual):
```
hakunanatasha, line 54: why 10?

tests/unit_tests/encoder/numeric/test_numeric.py
```diff
@@ -1,7 +1,17 @@
 import unittest
 import numpy as np
+import torch
 from lightwood.encoder.numeric import NumericEncoder
 from lightwood.encoder.numeric import TsNumericEncoder
+from lightwood.helpers.general import is_none
+
+
+def _polute(array):
```
hakunanatasha, line 9: should be _pollute() but i like the spirit of this

lightwood/encoder/numeric/numeric.py
```diff
@@ -80,7 +94,13 @@ def encode(self, data):
...
+        """
+        :param encoded_values: The encoded values to decode into single numbers
+        :param decode_log: Wether to decode the ``log`` or ``linear`` part of the representation, since the encoded vector contains both a log and a linear part
+
+        :returns: A number
```
hakunanatasha, line 102: can we say "returns the number expected from decoding"

hakunanatasha (changes requested):
Please see changes and add

hakunanatasha:
I would rather you make the nan-handling changes separate from the `Numerical` documentation; that way the NaN handling changes can be approved and Numeric can be worked on; otherwise it makes it difficult for me to just approve since the other changes need work.

lightwood/data/encoded_ds.py
George3d6: There isn't because this is torch not python numbers and/or numpy numbers

lightwood/encoder/numeric/numeric.py
George3d6: Used in a bunch of places such as time series, but I don't think we need to exemplify it here. After all, we don't do this for any other args and we'd go insane if we did, unless I'm misunderstanding.

lightwood/encoder/numeric/numeric.py
George3d6: I honestly don't remember why we chose it, i.e. I mentioned the experiments we did during the call but they were inconclusive, I don't think we can motivate this until we have benchmarks to show it's good in specific scenarios.

lightwood/encoder/numeric/numeric.py
George3d6: Other encoders don't have it because they don't decode numbers (this is for decoding, not encoding). Made comment explaining it.

lightwood/encoder/numeric/numeric.py
George3d6: I thought the top-level comment already explained it quite well, but added you comment in this case.

lightwood/encoder/numeric/numeric.py
George3d6: done

lightwood/encoder/numeric/numeric.py
George3d6: The column will be removed before this happens because of the information content check.

George3d6: Also, if it's an input, it's a meaningless input, and even if we alter it such that it doesn't get removed, it will always be [0,0,0,0] (see try except in encoding inputs)... which is as well as having any other constant value to represent `0`

lightwood/encoder/numeric/numeric.py
George3d6: I feel like this is already well described under `NumeriEncoder` (indeed, too well, the fallacy of writing down code as comments, the code will change and the comments won't, the code/comments will have bugs that the other won't)

I can move that description here, or otherwise, feel free to suggest what's missing from the description.

lightwood/mixer/regression.py
George3d6: these are my docs too :p will fix conflict if it arises, not sure why this is here.

tests/integration/advanced/test_timeseries.py
George3d6: because tests were failing locally becuase the RNN encoder doesn't handle time aim yet, but we still keep the tests "for the future" when it will properly handle it.

lightwood/encoder/numeric/numeric.py
George3d6: Not sure what `expected` represents in that context.

George3d6:
> I would rather you make the nan-handling changes separate from the Numerical documentation; that way the NaN handling changes can be approved and Numeric can be worked on; otherwise it makes it difficult for me to just approve since the other changes need work.

The numeric encoder suffered heavy changes in this PR so they ended up being together, might have made more sense to do them sequentially, but alas, that's hindsight, now they are way too annoying to split.

PR #721, opened 2021-11-04
More docs: image and audio encoders; LightGBM mixer

paxcema:


lightwood/encoder/audio/mfcc.py
```diff
@@ -9,20 +9,40 @@
 
 class MFCCEncoder(BaseEncoder):
     """
-    Audio encoder. Uses `librosa` to compute the Mel-frequency spectral coefficients (MFCCs) of the audio file. They are a common feature used in speech and audio processing. The features are a 2D array, flattened into a 1D one.
+    Audio encoder.
+
+    Uses `librosa` to compute the Mel-frequency spectral coefficients (MFCCs) of the audio file. They are a common feature used in speech and audio processing.
```
hakunanatasha, line 14: might be worth linking a paper or even wiki if it's a standard enough technique. I am not familiar with audio processing approaches.

lightwood/encoder/audio/mfcc.py
```diff
@@ -9,20 +9,40 @@
 
 class MFCCEncoder(BaseEncoder):
     """
-    Audio encoder. Uses `librosa` to compute the Mel-frequency spectral coefficients (MFCCs) of the audio file. They are a common feature used in speech and audio processing. The features are a 2D array, flattened into a 1D one.
+    Audio encoder.
```
hakunanatasha, line 12: ive been putting my docstrings in the init; since rst will auto-type arguments, i dont know if init is automatically recognized if the docstring is up here (hence i dunno if the params will be shown).

it's very possible that's the case - if so, I will move my docstrings up under the class definition, otherwise, let's put it in the init call.

lightwood/encoder/audio/mfcc.py
```diff
@@ -9,20 +9,40 @@
...
 
     def prepare(self, priming_data: pd.Series):
+        """
+        Retrieves and stores the expected output_size.
+
+        :param priming_data: training data
+        """
         self.is_prepared = True
         priming_data = list(priming_data)
         ele = self.encode([str(priming_data[0])])[0]
```
hakunanatasha, line 37: do we have a guarantee that the first example will always yield the same size? I don't know enough about this encoder to say either way.

lightwood/encoder/audio/mfcc.py
```diff
@@ -9,20 +9,40 @@
...
     def __init__(self, is_target: bool = False):
+        """
+        :param is_target: whether this encoder's column is the target. Should be false as encoder is not bi-directional. 
+        """  # noqa
+        assert not is_target
         super().__init__(is_target)
 
     def prepare(self, priming_data: pd.Series):
+        """
+        Retrieves and stores the expected output_size.
```
hakunanatasha, line 31: "The audio encoder undergoes rule-based processing. Thus, the prepare statement only returns the output dimension size"

is the output dimension size *fixed* always? as in, no matter what input audio goes in, do we expect the same size vector? cause if this is the case we can literally just so self.output_size = number

lightwood/encoder/audio/mfcc.py
```diff
@@ -9,20 +9,40 @@
...
+        :param priming_data: training data
+        """
         self.is_prepared = True
         priming_data = list(priming_data)
         ele = self.encode([str(priming_data[0])])[0]
         self.output_size = len(ele)
 
     def encode(self, column_data):
+        """
+        Encode a list of audio files.
```
hakunanatasha, line 42: this is *very important*; it means the input here is expected to be an audio file NAME i.e. a str not some meta-object.

lightwood/encoder/image/img_2_vec.py
```diff
@@ -9,12 +9,25 @@
 
 
 class Img2VecEncoder(BaseEncoder):
+    """
+    This encoder generates encoded representations for images using a pre-trained deep neural network.
+
+    All input images are rescaled to a standard size of 224x224, and normalized using the mean and standard deviation of the ImageNet dataset (as it was used to train the underlying NN).
```
hakunanatasha, line 15: im ok with the size but is there a reason we picked 224 x 224? could this be a kwarg?

lightwood/encoder/image/img_2_vec.py
```diff
@@ -9,12 +9,25 @@
...
+    
+    Note that this encoder does not have a .decode() method yet. As such, models that predict images as output are not supported at this time.
```
hakunanatasha, line 17: oh man a decode would be *wild*

lightwood/encoder/image/img_2_vec.py
```diff
@@ -40,12 +56,20 @@ def prepare(self, train_priming_data: pd.Series, dev_priming_data: pd.Series):
...
+
+        :return: same object but moved to the target device.
+        """
         self.model.to(device, available_devices)
         return self
 
     def encode(self, images: List[str]) -> torch.Tensor:
         """
-        Encode list of images
+        Encode a list of images.
```
hakunanatasha, line 72: May want to use the term "Creates an encoding for images; each image is referenced by the filepath or url"

lightwood/mixer/lightgbm.py
```diff
@@ -47,10 +47,31 @@ class LightGBM(BaseMixer):
     use_optuna: bool
     supports_proba: bool
 
+    """
+    Gradient boosting mixer, uses LightGBM as backbone.
+    
+    This mixer is a good all-rounder, due to the generally great performance of tree-based ML algorithms for supervised learning tasks with tabular data.
+    If you want more information regarding the techniques that set apart LightGBM from other gradient boosters, please refer to their technical paper: "LightGBM: A Highly Efficient Gradient Boosting Decision Tree" (2017).
+    
+    We can basically think of this mixer as a wrapper to the LightGBM interface. To do so, there are a few things the user may want to be aware about:
```
hakunanatasha, line 56: "As this mixer uses the formal implementation of LightGBM, here are some caveats to be aware of:"

lightwood/mixer/lightgbm.py
```diff
@@ -47,10 +47,31 @@ class LightGBM(BaseMixer):
     use_optuna: bool
     supports_proba: bool
 
+    """
+    Gradient boosting mixer, uses LightGBM as backbone.
```
hakunanatasha, line 51: gradient boosted tree-method i assume a mixer called lightgbm uses lgbm as its backbone ;)

lightwood/mixer/lightgbm.py
```diff
@@ -193,6 +227,12 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
             self.partial_fit(dev_data, train_data)
```
hakunanatasha, line 227: why do we auto-partial fit on dev??????? This is not acceptable! We call partial_fit in the `learn()` call!

hakunanatasha (changes requested):
I have a couple of issues with the logic flow of lgbm that we should discuss before merging.

lightwood/encoder/audio/mfcc.py
paxcema: I am fairly confident this is the case, as the method in `encode()` is essentially a call to `librosa.feature.mfcc()` with the same arguments every time.

lightwood/encoder/audio/mfcc.py
paxcema: It _is_ fixed, but inside `.encode()`. I agree this should be refactored to go for best practices, but I'd argue it should be on a separate PR.

lightwood/encoder/audio/mfcc.py
paxcema: Rephrased for clarity.

lightwood/encoder/image/img_2_vec.py
paxcema: Because it's the standard for torchvision pretrained models. We could turn it into a kwarg, but then we'd have to rewrite this to be trainable from scratch with different architectures, which I'd rather not do here.

lightwood/mixer/lightgbm.py
paxcema: Discussed on today's call.

PR #722, opened 2021-11-04
Documentation for the linear regression

George3d6:


lightwood/mixer/regression.py
```diff
@@ -41,10 +57,24 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
         log.info(f'Regression based correlation of: {self.model.score(X, Y)}')
 
     def partial_fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
+        """
+        Fits the Neural mixer on some data, this refits the model entirely rather than updating it
```
paxcema, line 61: Typo: 
> Fits the _regression_ mixer

lightwood/mixer/regression.py
```diff
@@ -12,11 +12,21 @@
...
+    A mixer that runs a simple linear regression using the (Encoded) features to predict the target.
+    Supports all types because they are all encoded numerically.
+    """ # noqa
     model: LinearRegression
     label_map: dict
     supports_proba: bool
 
     def __init__(self, stop_after: int, target_encoder: BaseEncoder, dtype_dict: dict, target: str):
+        """
+        :param stop_after: Maximum amount of time it should train for, currently ignored
```
hakunanatasha, line 31: in seconds? minutes? add the time interval.

lightwood/mixer/regression.py
```diff
@@ -25,6 +41,12 @@ def __init__(self, stop_after: int, target_encoder: BaseEncoder, dtype_dict: dic
         self.stable = False
 
     def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
+        """
+        Fits `LinearRegression` model on input feature data to provide predictions.
+
+        :param train_data: The EncodedDs on which to train the network
```
hakunanatasha, line 47: what network?

lightwood/mixer/regression.py
```diff
@@ -25,6 +41,12 @@ def __init__(self, stop_after: int, target_encoder: BaseEncoder, dtype_dict: dic
...
+        :param dev_data: Data used for early stopping and hyperparameter determination
```
hakunanatasha, line 48: for linear regression, there are no hyper parameters or early stopping

lightwood/mixer/regression.py
```diff
@@ -41,10 +63,24 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
         log.info(f'Regression based correlation of: {self.model.score(X, Y)}')
 
     def partial_fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
+        """
+        Fits the Neural mixer on some data, this refits the model entirely rather than updating it
+
+        :param train_data: The EncodedDs on which to train the network
```
hakunanatasha, line 69: there is no network. this is the simplest model.

lightwood/mixer/regression.py
```diff
@@ -41,10 +63,24 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
...
+        :param dev_data: Data used for early stopping and hyperparameter determination
```
hakunanatasha, line 70: likewise no network, no hyperparams, no early stopping

lightwood/mixer/regression.py
```diff
@@ -41,10 +63,24 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
         log.info(f'Regression based correlation of: {self.model.score(X, Y)}')
 
     def partial_fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
```
hakunanatasha, line 65: The inputs to partial-fit are incorrectly named. We should change it to `old_data` and `new_data`. I suspect this is a holdover of train's inputs, where dev data is specifically reserved for hyperparam/earlystopping etc.

In this case, the name "dev_data" is actually used to adjust the model with new information; this is *unintuitive*. We need to fix this for the entire codebase.

lightwood/mixer/regression.py
```diff
@@ -41,10 +63,24 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
...
+
+        :param train_data: The EncodedDs on which to train the network
+        :param dev_data: Data used for early stopping and hyperparameter determination
+        """
         self.fit(train_data, dev_data)
 
     def __call__(self, ds: EncodedDs,
                  args: PredictionArguments = PredictionArguments()) -> pd.DataFrame:
+        """
+        Make predictions based on datasource similar to the one used to fit (sans the target column)
```
hakunanatasha, line 77: This doesn't even make sense, what do you mean similar? The same data structure? This should be obvious.

lightwood/mixer/regression.py
```diff
@@ -41,10 +63,24 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
...
+
+        :param ds: The EncodedDs for which to generate the predictions
```
hakunanatasha, line 79: Sphinx already type-hints you don't need to call it by the dataclass.

hakunanatasha: ":param ds: The featurized dataset used to generate predictions"

lightwood/mixer/regression.py
```diff
@@ -41,10 +63,24 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
...
+        :param arg: Argument for predicting
```
hakunanatasha, line 80: ":param arg: Any additional arguments used in predicting"

hakunanatasha: does this even have additional args? I don't know if this is just there to make things modular to the mixer format, but ideally we don't even need to have this extra term.

lightwood/mixer/regression.py
hakunanatasha: You need to actually discuss how partial-fit is handled in LinearRegression. LinearRegression is cheap, hence we simply re-fit the model with more data (old data + new data). This is NOT the same as in other models, for example, neural networks, where you can get away with training on just new data assuming the old data representations have been "learned" or fine-tuning.

hakunanatasha (changes requested):
These functions are incorrectly annotated (I assume because they are borrowed from other areas). Please make sure to detail arguments and actually provide a deeper reference as to what's going on.

lightwood/mixer/regression.py
George3d6: Oh, good point. This is an issue in like half a dozen places (we don't specify) the unit. So I went ahead and did a more generic fix where `stop_after` and `time_aim` are now typed as `seconds` so that it's obvious what they are all the time.

lightwood/mixer/regression.py
hakunanatasha: There is no choice to motivate. If you retrain Linear regression with a new dataset, it's as if the old dataset was never seen. 

The problem is the term `partial_fit` is ambiguous, so we need to be clear what `refit` means, because in certain contexts `refitting` means that we can train without the old data (i.e. neural nets can be fine-tuned with new data). Terms like `update` and `refit` do not discuss *how* changes are made. It doesn't hurt to be explicit.

lightwood/api/types.py
```diff
@@ -15,6 +15,9 @@
 import json
 
 
+seconds = float
+
+
 # See: https://www.python.org/dev/peps/pep-0589/ for how this works
 # Not very intuitive but very powerful abstraction, might be useful in other places (@TODO)
```
hakunanatasha, line 22: why is this here? please remove it

hakunanatasha (changes requested):
This is an over-engineered solution. When I said time interval, I merely mean to annotate in the docs "this is in seconds". Making a weird reserved type that is a float is pretty confusing, as people will then need to look around for what weird data type we've enforced `seconds` to be. It can be a float or int param argument, it's fine.

When we actually DO have a time-aim, the package we use to explicitly "stop" training has an expectation for the input. That's what we want to make clear.

PR #724, opened 2021-11-04
Remove type mapping

George3d6:
- Json ai now *requires* explicit `data_type`s for all the features instead of sourcing them from the namespace of the encoder
- Removed "typed" encoder namespaces and their usage


hakunanatasha (approved):
looks fine.

PR #732, opened 2021-11-08
Build and deploy wheel file

paxcema:
# Why

To try and add lightwood in offline Kaggle kernels, having a wheel file _might_ be useful.

# How

Modifies the existing command to generate a .whl file too.


George3d6:
No, don't push wheels to pip, because if we don't build wheels for multiples OSes we end up with people downloading incompatible wheels instead of pip building from source (which would work)

Put the wheel on AWS

Issue #735, opened 2021-11-08
Test get_identifier_description


George3d6:
Important logic, we should unittest it with real-world examples and generated data

paxcema:
As discussed, may be worth moving some modules in `lightwood/helpers/text` to a separate "identification" module or similar.

PR #736, opened 2021-11-08
Drop column with no information

George3d6:
Dropping columns that don't have any information (a single unique value) 

Currently no test for this since `get_identifier_description` has no tests (maybe todo for later) and this is so dumb it's not the first thing I'd want to test, see #735

### DIFF
--- a/lightwood/helpers/text.py
+++ b/lightwood/helpers/text.py
@@ -215,7 +215,7 @@ def get_identifier_description(data, column_name, data_dtype):
     if nr_unique == 1:
         return 'No Information'
 
-    unquie_pct = nr_unique / len(data)
+    unique_pct = nr_unique / len(data)
 
     spaces = [len(str(x).split(' ')) - 1 for x in data]
     mean_spaces = np.mean(spaces)
@@ -223,7 +223,7 @@ def get_identifier_description(data, column_name, data_dtype):
     # Detect auto incrementing index
     # -- some cases where I guess people do want to use this for learning, so ignoring this check for now...
     # if data_dtype == dtype.integer:
-    #    if get_pct_auto_increment(data) > 0.98 and unquie_pct > 0.99:
+    #    if get_pct_auto_increment(data) > 0.98 and unique_pct > 0.99:
     #        return 'Auto-incrementing identifier'
 
     # Detect hash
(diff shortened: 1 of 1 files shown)


paxcema (approved):
As discussed, may be worth moving these modules to a separate _identification_ module, but as far as dropping uninformative columns this seems OK to me.

lightwood/helpers/text.py
```diff
@@ -210,7 +210,12 @@ def get_identifier_description_mp(arg_tup):
 
 def get_identifier_description(data, column_name, data_dtype):
```
hakunanatasha, line 211: type hint

data: Iterable, column_name: str, data_dtype: str

lightwood/helpers/text.py
```diff
@@ -210,7 +210,12 @@ def get_identifier_description_mp(arg_tup):
...
     data = list(data)
```
hakunanatasha, line 212: why do we make this a list?

hakunanatasha: if it's already in a series form you can simply do something like len(data.unique())

lightwood/helpers/text.py
```diff
@@ -210,7 +210,12 @@ def get_identifier_description_mp(arg_tup):
...
-    unquie_pct = len(set(data)) / len(data)
+    nr_unique = len(set(data))
+
+    if nr_unique == 1:
+        return 'No Information'
+
+    unquie_pct = nr_unique / len(data)
```
hakunanatasha, line 218: fix the spelling

lightwood/helpers/text.py
```diff
@@ -210,7 +210,12 @@ def get_identifier_description_mp(arg_tup):
...
 
     spaces = [len(str(x).split(' ')) - 1 for x in data]
     mean_spaces = np.mean(spaces)
```
hakunanatasha, line 221: why are we doing this?

hakunanatasha (changes requested):

lightwood/helpers/text.py
George3d6: In order to mark the column for dropping under the "identifier / foreign-keys" rule, if this function returns a string (rather than `None`) the column is dropped, maybe the interface is confusing. And maybe we should broaden the definition of "identifier" to something like "no_information" ?

PR #740, opened 2021-11-09
Separate doc branch

George3d6:
Doc builds are pushed to a separate `gh-pages` branch

Currently https://lightwood.io/ is source from that branch and everything looks peachy :)


ZoranPandovski (approved):

docs/CNAME
```diff
@@ -1 +0,0 @@
-lightwood.io
```
paxcema, line 1: Is it okay for this to be gone?

paxcema: nevermind, it's still in the Makefile :+1:

ZoranPandovski: I think this is required when connecting GitHub pages to a custom domains?

paxcema: Yeah, but it seems it's being generated when doing `make github` in the `doc_build.yml` file, so it _should_ be okay :thinking:

PR #741, opened 2021-11-10
Test notebooks and docs

George3d6:
- [x] Parse out all the code from the README and run it during testing in order to make sure it's always valid
- [x] Run jupyter notebooks every time we build the docs
- [x] Run jupyter notebooks as part of the test suite
- [x] Added `load_custom_module` function to API, takes a file path and create modules dir (if needed) and moves (or replaces) it in that dir
- [x] Removed pre-built notebooks so that they always *have* to be newly generated in order to be added to the docs
- [x] Test *must* pass before doc is built, and test outputs in the form of the notebook HTML files are used by the doc generation  
- [x] Added instructions for building *new* tutorials (based on how current tutorials had to be standardized)
- [x] Edited every single tutorial in order for it to actually run on github


tests/docs/test_readme.py
```diff
@@ -0,0 +1,18 @@
+import os
+import unittest
+
+
+class TestReamde(unittest.TestCase):
```
hakunanatasha, line 5: this is very neat :)

hakunanatasha (approved):

hakunanatasha:
I'm assuming this is WIP for other things

George3d6:
Yes, markdown I had for this is wrong, but basically `[]` means WIP, `[x]` means done, I'd want to get README tests + notebook running here.

The inlining of test code inside `.rst` files, as discussed, will be forgotten until needed

George3d6:
Closing to avoid action spam!

paxcema (approved):

hakunanatasha:
hold off on this until my review

docssrc/source/tutorials/custom_cleaner/custom.json
```diff
@@ -14,12 +14,13 @@
     "features": {
         "excerpt": {
             "encoder": {
-                "module": "Rich_Text.PretrainedLangEncoder",
+                "module": "PretrainedLangEncoder",
                 "args": {
                     "output_type": "False",
```
hakunanatasha, line 19: how can output type be false?

we should standardize data types going in

docssrc/source/tutorials/custom_splitter/README.md
```diff
@@ -0,0 +1 @@
+Disabled unitl we can get a link to the data
```
hakunanatasha: why would you disable the custom splitter???

lightwood/api/high_level.py
```diff
@@ -17,6 +17,18 @@
...
+from shutil import copyfile
+
+
+def load_custom_module(file_path: str):
+    modules_dir = os.path.join(os.path.expanduser('~/lightwood_modules'))
+    custom_module_path = os.path.join(modules_dir, os.path.split(file_path)[-1])
+    if not os.path.exists(modules_dir):
+        os.mkdir(modules_dir)
+    if os.path.exists(custom_module_path):
+        os.remove(custom_module_path)
```
hakunanatasha, line 29: I'm not sure if removing is the best idea

hakunanatasha (changes requested):
there are a few changes i dislike; moreover since the encoders are going to change, we may need to rework some of the tutorials.

I need to think more on the load_module feature; i don't see why it's useful for the notebooks and I'm averse to deleting folders that we don't have business deleting. 

Also, I think disabling the custom splitter tutorial is a bad idea; we may want to just not run that doc. It seems silly to omit certain tutorials when we know they can run.

docssrc/source/tutorials/custom_cleaner/custom.json
George3d6: This is a mistake, types are standardized, but not enforced because enforcign this in python has to do with 3rd party tools

George3d6: Also that is and incorrect file that's not used anywhere, this shluld be deleted, ups xD

docssrc/source/tutorials/custom_splitter/README.md
George3d6: Because I don't have the data to run it, do you have those .csvs somewhere.

paxcema (changes requested):
Please remove the `.zip` file, I don't think we should include it in this repo. I suggest moving it to the public benchmarks.

George3d6:
Will move all data to s3 for now, doesn't make sense in private benchmarks either since we don't bench on this and I wouldn't want to (licensing issues making it hard to open source long term in a very public fashion)

out.txt
```diff
@@ -0,0 +1,212 @@
+[NbConvertApp] Converting notebook ./docssrc/source/tutorials/custom_explainer/custom_explainer.ipynb to notebook
```
paxcema, line 1: Sneaky leftover file 👀

tests/docs/test_readme.py
paxcema: Typo in the class name

lightwood/api/high_level.py
paxcema: Module loading based on the package version is out of scope for this PR, I assume?

paxcema: Also +1 to removing being too harsh. Maybe move contents to `custom_module_path/old` or something like that.

lightwood/analysis/explain.py
```diff
@@ -50,11 +50,6 @@ def explain(data: pd.DataFrame,
     global_insights = {}
     row_insights['prediction'] = predictions['prediction']
 
-    if target_name in data.columns:
```
paxcema, line 53: For reference, I double checked and as far as I can tell this is okay 👍

.gitignore
```diff
@@ -64,4 +64,9 @@ docssrc/source/tutorials/*.html
 docssrc/source/tutorials/*.pickle
 docssrc/source/tutorials/*.py
 docs
-docs/*
\ No newline at end of file
+<<<<<<< HEAD
```
paxcema, line 67: Heads up, some pending merge conflicts it seems

George3d6:
New PR here: https://github.com/mindsdb/lightwood/pull/764 | closing this one

PR #742, opened 2021-11-10
Reduce cat complexity

hakunanatasha:
Categorical OHE and Binary need to be less complex than they currently are. Here is a WIP proposal.

(1) Binary is a specific case of OHE, we may want to make a generic OHE encoder and enforce a style
(2) OHE has a few changes
- *max_dimensions* is not a useful criteria due to handling of OHE earlier in the pipeline (We don't run this if len(categories) > 100)
- unknown mode should be Boolean, there are only 2 criteria and str handling is not optimal
- `decode` doesn't obey the expected output criteria because of return_probabilities; this has been abstracted to a new function
- `decode_probabilities` handles the case in which we want to create a bounded estimate [0, 1]
- unit_tests previously could exhibit silent errors (i.e. returning an encoded vector where unk dimension = 1 + token)

I will be standardizing some of the type hints in the meanwhile.


hakunanatasha:
Moved to 'standardize encoders'

PR #748, opened 2021-11-11
Fix #738

### LINKED ISSUE #738: Fix encode/decode in TsArrayNumericEncoder
Current methods are functionally equivalent to `encode_one()` and `decode_one()`, which is confusing. We need to rename these and write the wrappers `encode()` and `decode()`, all the while documenting the encoder.

paxcema:
## Why

To get a consistent behavior across encoders. See #738 for additional details, but `TsArrayNumericEncoder` was not behaving like a normal encoder would.

## How

Add `encode_one` and `decode_one` methods that are called for each data point in their respective wrappers, `encode` and `decode`. Given that this encoder has receives input from multiple columns, the behavior in `EncodedDs.__getitem__()` has been slightly modified.

Finally, and as an aside, `Neural()` now does not perform a final tuning on time series tasks (was incorrectly doing so without dependency data being passed).

### DIFF
--- a/lightwood/data/encoded_ds.py
+++ b/lightwood/data/encoded_ds.py
@@ -67,10 +67,11 @@ class EncodedDs(Dataset):
                 if hasattr(self.encoders[col], 'data_window'):
                     cols = [self.target] + [f'{self.target}_timestep_{i}'
                                             for i in range(1, self.encoders[col].data_window)]
+                    data = [self.data_frame[cols].iloc[idx].tolist()]
                 else:
                     cols = [col]
+                    data = self.data_frame[cols].iloc[idx].tolist()
 
-                data = self.data_frame[cols].iloc[idx].tolist()
                 encoded_tensor = self.encoders[col].encode(data, **kwargs)[0]
                 if torch.isnan(encoded_tensor).any() or torch.isinf(encoded_tensor).any():
                     raise Exception(f'Encoded tensor: {encoded_tensor} contains nan or inf values')
--- a/lightwood/encoder/numeric/ts_array_numeric.py
+++ b/lightwood/encoder/numeric/ts_array_numeric.py
@@ -42,6 +42,14 @@ class TsArrayNumericEncoder(BaseEncoder):
             dependency_data = {'__default': [None] * len(data)}
 
         ret = []
+        for series in data:
+            ret.append(self.encode_one(series, dependency_data=dependency_data))
+
+        return torch.vstack(ret)
+
+    def encode_one(self, data, dependency_data={}):
+        ret = []
+
         for data_point in data:
             ret.append(self.sub_encoder.encode([data_point], dependency_data=dependency_data))
 
@@ -62,6 +70,16 @@ class TsArrayNumericEncoder(BaseEncoder):
                                                 self.sub_encoder.output_size)
 
         ret = []
-        for encoded_timestep in torch.split(encoded_values, 1, dim=1):
+        for tensor in torch.split(encoded_values, 1, dim=0):
+            ret.append(self.decode_one(tensor, dependency_data=dependency_data))
+
+        if encoded_values.shape[0] == 1:
+            return ret[0]  # for a single value, we omit wrapping inside a list (TBD in #744)
+        else:
+            return ret
+
+    def decode_one(self, encoded_value, dependency_data={}):
+        ret = []
+        for encoded_timestep in torch.split(encoded_value, 1, dim=1):
             ret.extend(self.sub_encoder.decode(encoded_timestep.squeeze(1), dependency_data=dependency_data))
         return ret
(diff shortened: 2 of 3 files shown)


lightwood/data/encoded_ds.py
```diff
@@ -67,10 +67,11 @@ def __getitem__(self, idx: int) -> Tuple[torch.Tensor, torch.Tensor]:
                 if hasattr(self.encoders[col], 'data_window'):
                     cols = [self.target] + [f'{self.target}_timestep_{i}'
                                             for i in range(1, self.encoders[col].data_window)]
+                    data = [self.data_frame[cols].iloc[idx].tolist()]
```
George3d6, line 70: I'm really confused by what's happening here...

Originally we do `tolist()` to convert the single element (which ends up being a pd.Series anyway) to a list to fit the signature.

In this case we need a list of lists... will the above code have `.tolist()` convert the dataframe composed of the selected columns at that index into a list ? If so that'd make sense.

lightwood/encoder/numeric/ts_array_numeric.py
```diff
@@ -58,6 +70,16 @@ def decode(self, encoded_values, dependency_data=None, return_all=False):
                                                 self.sub_encoder.output_size)
 
         ret = []
-        for encoded_timestep in torch.split(encoded_values, 1, dim=1):
+        for tensor in torch.split(encoded_values, 1, dim=0):
+            ret.append(self.decode_one(tensor, dependency_data=dependency_data))
+
+        if encoded_values.shape[0] == 1:
+            return ret[0]  # for a single value, we omit wrapping inside a list (TBD in #744)
```
George3d6, line 77: No, we shouldn't `decode` and `encode` should have consistent behaviour, always take and return Iterables (well, take Iterables and Tensors respectively, return `List`, but you get the point)

George3d6: Sorry if this requires more annoying changes to usage :(

George3d6: But `if edge case then return changes` is a pattern that gets really weird really quick

lightwood/mixer/neural.py
```diff
@@ -289,7 +289,9 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
 
         if self.fit_on_dev:
             self.partial_fit(dev_data, train_data)
-        self._final_tuning(dev_data)
+
+        if not self.timeseries_settings.is_timeseries:
+            self._final_tuning(dev_data)
```
George3d6, line 294: Wait... why are we not doing final tuning for timeseries? Why is the change part of this PR?

George3d6 (changes requested):

lightwood/mixer/neural.py
paxcema: It's on the PR description:

> Finally, and as an aside, Neural() now does not perform a final tuning on time series tasks (was incorrectly doing so without dependency data being passed).

I guess it could be a separate PR, if you prefer that.

lightwood/data/encoded_ds.py
paxcema: Not quite. This has to do with the following: is the expected input to encode() a list of items to encode?
If so, the list wrapper is there because of the following: say we have two rows for which we'll retrieve target_timestep_0 and target_timestep_1  (which is what this encoder does, i.e. arrays are its inputs). Then, the input to encode would be:

```
[
  [row1_time_0_value, row1_time_1_value], 
  [row2_time_0_value, row2_time_1_value], 
]
```

Which I believe conforms to the discussed interface, unless I'm off the mark here. Added types for good measure.

lightwood/encoder/numeric/ts_array_numeric.py
paxcema: You're right. Fixed!

lightwood/mixer/neural.py
```diff
@@ -341,10 +343,7 @@ def __call__(self, ds: EncodedDs,
                 else:
                     decoded_prediction = self.target_encoder.decode(Yh, **kwargs)
 
-                if not self.timeseries_settings.is_timeseries or self.timeseries_settings.nr_predictions == 1:
-                    decoded_predictions.extend(decoded_prediction)
-                else:
-                    decoded_predictions.append(decoded_prediction)
+                decoded_predictions.extend(decoded_prediction)
```
paxcema, line 346: This solves the "edge case" behavior noted in an earlier comment by simplifying `Neural.__call__()`

lightwood/mixer/neural.py
George3d6: My bad... but shouldn't there be a PR to fix final tuning then?

lightwood/data/encoded_ds.py
George3d6: yeap, all good

PR #749, opened 2021-11-14
Various fixes

George3d6:
- `target_weights` argument is now used properly
- light gbm mixer supports weights
- OHE and binary encoder logic around weights makes a bit more sense now due to saner names


lightwood/encoder/categorical/onehot.py
```diff
@@ -61,17 +70,16 @@ def prepare(self, priming_data, max_dimensions=20000):
...
+        # Note: Is target assume that we are operating in "unknown_token" mode
         if self.is_target:
-            self.index_weights = [None] * self._lang.n_words
-            if self.target_class_distribution is not None:
-                self.index_weights[0] = np.mean(list(self.target_class_distribution.values()))
-            else:
-                self.index_weights[0] = 1
+            self.index_weights = [1] * self._lang.n_words
+            if self.target_weights is not None:
+                self.index_weights[0] = np.mean(list(self.target_weights.values()))
```
hakunanatasha, line 77: Why do we do this?
```
                self.index_weights[0] = np.mean(list(self.target_weights.values()))
                self.target_weights[UNCOMMON_WORD] = self.index_weights[0]
```

This implies that the unknown weight is the average likelihood of any other class; set this to 0. We're surprisingly up-weighting unknowns.

lightwood/mixer/lightgbm.py
```diff
@@ -127,6 +131,8 @@ def _to_dataset(self, data: Dict[str, Dict], output_dtype: str):
                     self.ordinal_encoder.fit(np.array(list(self.label_set)).reshape(-1, 1))
 
                 label_data = [x if x in self.label_set else '__mdb_unknown_cat' for x in label_data]
+                if weight_map is not None:
+                    data[subset_name]['weights'] = [weight_map[x] for x in label_data]
```
hakunanatasha, line 135: weights are on columns-  wouldn't this be on rows?

lightwood/mixer/lightgbm.py
```diff
@@ -182,10 +188,22 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
...
+
+        Because it guards against the following crash:
+
+        WARNING:lightwood-1613058:Exception: Cannot change feature_pre_filter after constructed Dataset handle. when training mixer: <lightwood.mixer.lightgbm.LightGBM object at 0x7f39148eae20>
+        feature_fraction, val_score: inf:   0%|                                 | 0/7 [00:00<?, ?it/s]
+        free(): double free detected in tcache 2
+
+        Only happens sometimes and I can find no pattern as to when, happens for multiple input and target types.
+
+        Why does the following crash happen and what does it mean? No idea, closest relationships I can find is /w optuna modifying parameters after the dataset is create: https://github.com/microsoft/LightGBM/issues/4019 | But why this would apply here makes no sense. Could have to do with the `train` process of lightgbm itself setting a "set only once" property on a dataset when it starts. Dunno, if you find out replace this comment with the real reason.
```
hakunanatasha, line 202: this is really bad if we can't figure out what's wrong

hakunanatasha (commented):
Some major comments but I don't want to block this since it apparently has a hotfix. Please respond to comments and then merge.

lightwood/mixer/lightgbm.py
George3d6: For Lightgbm weighting is on a per-row basis (there's only ever a single target column) ... only way to do a value-map is using the sklearn API, which just converts to this, and which is buggier

lightwood/mixer/lightgbm.py
George3d6: It is.... but I've spent some time trying to. Even if we did, likely no fix, when I have more time I will do a self-contained replication and report the bug (maybe contribute a fix) in the LightGBM repo.

Ultimately there's a bunch of "design around bug of our libraries" things in lightwood, including the fact that all encoders are designed using that stupid `is_none` function because pandas can't ****** guarantee that values in a column don't change. It's non ideal, but this is hardly the worst offender.

PR #754, opened 2021-11-16
Dielectron problems

George3d6:
- using custom r2 to account for sklearn weirdness where it can overflow to a positive value
- added new acc function lookup hierarchy with our own function taking precedence over sklearn (note, we should aim to move `evaluate_array_accuracy` there)
- Added better warning logs for when an encoded rep contains nan or inf
- Fixed target dropping behavior


lightwood/api/types.py
```diff
@@ -351,13 +351,13 @@ def from_dict(obj: Dict):
...
         seconds_per_encoder = obj.get('seconds_per_encoder', None)
-        time_aim = obj.get('time_aim', None)
+        time_aim = obj.get('time_aim', 20)
         target_weights = obj.get('target_weights', None)
         positive_domain = obj.get('positive_domain', False)
         timeseries_settings = TimeseriesSettings.from_dict(obj.get('timeseries_settings', {}))
         anomaly_detection = obj.get('anomaly_detection', True)
         ignore_features = obj.get('ignore_features', [])
-        fit_on_all = obj.get('fit_on_all', True)
+        fit_on_all = obj.get('fit_on_all', False)
```
hakunanatasha: what does flipping to false do?

lightwood/api/types.py
```diff
@@ -351,13 +351,13 @@ def from_dict(obj: Dict):
         unbias_target = obj.get('unbias_target', True)
         seconds_per_mixer = obj.get('seconds_per_mixer', None)
         seconds_per_encoder = obj.get('seconds_per_encoder', None)
-        time_aim = obj.get('time_aim', None)
+        time_aim = obj.get('time_aim', 20)
```
hakunanatasha: is this going to affect everything? if so why 20?

lightwood/helpers/accuracy.py
```diff
@@ -0,0 +1,17 @@
+from sklearn.metrics import r2_score as sk_r2_score
```
hakunanatasha, line 1: please dont do this, we will never ever catch a bug again if so

hakunanatasha: i guess there is an exception here but i'd almost rather it just blow up

hakunanatasha (commented):
dumb questions for my own understanding but otherwise hotfix looks fine

i really hate the wrapper for the sklearn r2 weirdness but I am assuming the criteria of "if the exception triggers we stop the benchmarks"

lightwood/data/cleaner.py
```diff
@@ -289,6 +289,7 @@ def _remove_columns(data: pd.DataFrame, identifiers: Dict[str, object], target:
     if mode == "predict":
         if (
             target in data.columns
+            and timeseries_settings.is_timeseries
             and not timeseries_settings.use_previous_target
             and not anomaly_detection
```
paxcema, line 294: I think we need to change the default value in `anomaly_detection` for this to work

PR #755, opened 2021-11-16
Standardize encoders

hakunanatasha:
Update: 2021.11.29

Major refactor of binary + OHE; index_weights is preserved and passed to neural. I could not find references to LGBM. Streamlines input/outputs and updated docs (among other encoders).

# ----------------

Reworked some encoders to have significantly more details than previously, and to standardize the I/O of each argument (and some signatures).

Major changes to categorical: OHE + Binary, some changes to text/base/audio/image mainly for docs.

WIP.


lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
-            word_to_remove = None
-            for word in least_occuring_words:
-                if word not in necessary_words:
-                    word_to_remove = word
-                    break
-
-            self._lang.removeWord(word_to_remove)
-
-        # Note: Is target assume that we are operating in "unknown_token" mode
+        unq_cats = np.unique([i for i in priming_data if i is not None]).tolist()
```
George3d6, line 58: same as /w the other one

lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
+        if self.use_unknown:
+            log.info("Encoding UNK categories as index 0")
+            self.map = {cat: indx + 1 for indx, cat in enumerate(unq_cats)}
+            self.map.update({_UNCOMMON_WORD: 0})
+            self.rev_map = {indx: cat for cat, indx in self.map.items()}
+        else:
+            log.info("Encoding UNK categories as vector of all 0s")
```
George3d6, line 65: probably better to say `unkown` since `UNK` might confuse users and idk if there's a reason to use that abbreviation. Also `unkown categories and missing values` might be more explicit as to what this is doing.

lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
+            if index is not None:
+                ret[idx, index] = 1
+
+            if self.use_unknown and index is None:
+                ret[idx, 0] = 1
 
         return torch.Tensor(ret)
 
-    def decode(self, encoded_data, return_raw=False):
+    def decode(self, encoded_data: torch.Tensor):
```
George3d6, line 124: add `str` return type

lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
+                ret.append(_UNCOMMON_WORD)
+            else:
+                ret.append(self.rev_map[np.argmax(vector)])
+
+            probs.append(self._norm_vec(vector))
+
+        return ret, probs, self.rev_map
+
+    @staticmethod
+    def _norm_vec(vec: List[float]):
```
George3d6, line 167: In general, I think we prefer to just inline the method in the file rather than making it static but small style nitpick, if you like it better this way keep it.

lightwood/encoder/text/pretrained.py
```diff
@@ -135,35 +80,43 @@ def __init__(
...
         if self.embed_mode:
             log.info("Embedding mode on. [CLS] embedding dim output of encode()")
         else:
             log.info("Embedding mode off. Logits are output of encode()")
 
-    def prepare(self, train_priming_data: pd.Series, dev_priming_data: pd.Series, encoded_target_values: torch.Tensor):
+    def prepare(
+        self,
+        train_priming_data: pd.Series,
+        dev_priming_data: Optional[pd.Series],
```
George3d6, line 91: make these `Iterable` instead of pd series.

lightwood/encoder/text/pretrained.py
```diff
@@ -135,35 +80,43 @@ def __init__(
...
         if self.is_prepared:
             raise Exception("Encoder is already prepared.")
 
-        # TODO: Make tokenizer custom with partial function; feed custom->model
-        if self._tokenizer is None:
-            self._tokenizer = self._tokenizer_class.from_pretrained(self._pretrained_model_name)
+        os.environ['TOKENIZERS_PARALLELISM'] = 'true'
+
+        # TODO -> we shouldn't be concatenating these together
+        if dev_priming_data is not None:
```
George3d6, line 111: I don't think we should allow for null dev data, will make it a mess once we do try to implement early stopping or other things that require a dev set. Currently it's easy but won't be so in the future, + it changes the API from that which all other trainable encoders use (mandatory train and dev)

lightwood/encoder/categorical/binary.py
```diff
@@ -1,76 +1,162 @@
...
+
+        for vector in encoded_data_list:
+            if not np.any(vector):  # Vector of all 0s -> unknown category
+                ret.append(_UNCOMMON_WORD)
+            else:
+                ret.append(self.rev_map[np.argmax(vector)])
+
+        return ret
+
+    def decode_probabilities(self, encoded_data: torch.Tensor):
```
George3d6, line 132: Adding this method without changes to `Neural` (and regression?) will result in probability predictions not working and tests failing.

Also this should return `Tuple[List[str], List[List[float]], Dict[int,str]` ... I think, or whatever the types are, not 100% sure.

lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
 
-            if return_raw:
-                probs.append(softmax(vector).tolist())
+        return ret
 
-        if return_raw:
-            return ret, probs, self.rev_map
-        else:
-            return ret
+    def decode_probabilities(self, encoded_data: torch.Tensor):
```
George3d6, line 144: same comment as with the other one, predicting probabilities will fail unless appropriate changes are made to `Neural` and please specify return type.

George3d6 (changes requested):

lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
+    Creates a one-hot encoding (OHE) of categorical data. This creates a vector where each individual dimension corresponds to a category. A category has a 1:1 mapping between dimension indicated by a "1" in that position.
 
-    Why are we handling target weighting inside encoders? Simple: we'd otherwise have to compute per-index weighting inside the mixers, rather than having that code unified inside 2x encoders. So moving this to the mixer will still involve having to pass the target encoder to the mixer, but will add the additional complexity of having to pass a weighting map to the mixer and adding class-to-index translation boilerplate + weight setting for each mixer
-    """ # noqa
-    def __init__(self, is_target=False, target_weights=None, handle_unknown='unknown_token'):
-        super().__init__(is_target)
-        self._lang = None
-        self.rev_map = {}
+    OHE operates in 2 modes:
+        (1) "use_unknown=True": Makes an :math:`N+1` length vector for :math:`N` categories, the first index always corresponds to the unknown category.
```
paxcema, line 15: Ooh, `:math:` seems useful, there's a bunch of docstrings that could use this :eyes:

lightwood/encoder/categorical/binary.py
```diff
@@ -1,76 +1,162 @@
...
+    .. math::
+
+       A &= [1, 0] \\
+       B &= [0, 1]
+
+    This encoder is a specialized case of one-hot encoding (OHE); unknown categories are explicitly handled as [0, 0]. Unknowns may only be reported if the input row value is NULL (or python `None` type) or if new data, after the encoder is prepared, has examples outside the feature map.
+
+    When data is typed with Lightwood, this class is only deployed if an input data type is explicitly recognized as binary (i.e. the column has only 2 unique values like True/False). If future data shows a new category (thus the data is no longer truly binary), this encoder will no longer be appropriate unless you are comfortable mapping ALL new classes as [0, 0].
+
+    An encoder can also represent the target column; in this case, `is_target` is `True`, and `target_weights`. The `target_weights` parameter enables users to specify how heavily each class should be weighted within a mixer - useful in imbalanced classes.
```
paxcema, line 22: End of first sentence sounds weird

lightwood/encoder/categorical/binary.py
```diff
@@ -1,76 +1,162 @@
...
+
+    When data is typed with Lightwood, this class is only deployed if an input data type is explicitly recognized as binary (i.e. the column has only 2 unique values like True/False). If future data shows a new category (thus the data is no longer truly binary), this encoder will no longer be appropriate unless you are comfortable mapping ALL new classes as [0, 0].
+
+    An encoder can also represent the target column; in this case, `is_target` is `True`, and `target_weights`. The `target_weights` parameter enables users to specify how heavily each class should be weighted within a mixer - useful in imbalanced classes. 
+
+    By default, the `StatisticalAnalysis` phase will provide `target_weights` as the relative fraction of each class in the data which is important for imbalanced populations; for example, suppose there is a 80/10/10 imbalanced representation across 3 different classes - `target_weights` will be a vector as such::
+
+    target_weights = {"class1": 0.9, "class2": 0.1, "class3": 0.1}
```
paxcema, line 26: Probably meant to be `0.8`, `0.1`, `0.1`?

lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
-            raise ValueError(f"handle_unknown should be either 'unknown_token' or 'return_zeros', got {handle_unknown}")
-        else:
-            self.handle_unknown = handle_unknown
+        (2) "use_unknown=False": Makes an :math:`N` length vector for :math:`N` categories, where an empty vector of 0s indicates an unknown/missing category.
 
-        if self.is_target:
-            if self.handle_unknown != 'unknown_token':
-                raise ValueError(f'One Hot Encoders used for target encoding can only be used with `handle_unknown` \
-                                   set to `unknown_token`. The option: "{self.handle_unknown}" is not supported!')
+    An encoder can also represent the target column; in this case, `is_target` is `True`, and `target_weights`. The `target_weights` parameter enables users to specify how heavily each class should be weighted within a mixer - useful in imbalanced classes.
```
paxcema, line 19: Same as in the binary docstring regarding the first sentence.

lightwood/encoder/categorical/onehot.py
```diff
@@ -1,140 +1,174 @@
...
+
+    By default, the `StatisticalAnalysis` phase will provide `target_weights` as the relative fraction of each class in the data which is important for imbalanced populations; for example, suppose there is a 80/10/10 imbalanced representation across 3 different classes - `target_weights` will be a vector as such::
+
+    target_weights = {"class1": 0.9, "class2": 0.1, "class3": 0.1}
```
paxcema, line 23: Same as in the binary docstring regarding the typo in `target_weights`

tests/unit_tests/encoder/categorical/test_binary.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.assertTrue(
+            (
+                enc_data
+                == Tensor(
+                    [
+                        [
+                            1.0,
+                            0.0,
+                        ],  # category 1
```
paxcema, line 42: Are comments correct here? I spot a couple odd ones

tests/unit_tests/encoder/categorical/test_binary.py
```diff
@@ -0,0 +1,145 @@
...
+
+        enc = BinaryEncoder(is_target=True, target_weights=tweights)
+        enc.prepare(data)
+
+        # Check inverse weights correct
+        self.assertTrue(np.all(((enc.inv_target_weights - iweights) == 0).tolist()))
+
+    def test_distro_nonzeroweights(self):
+        """
+        Tests if target wts do not sum to 1 properly handled.
```
paxcema, line 116: wts -> weights?

lightwood/encoder/array/array.py
```diff
@@ -84,6 +96,12 @@ def encode(self, column_data: Union[list, np.ndarray, torch.Tensor]) -> torch.Te
 
         return data
 
-    def decode(self, data) -> torch.tensor:
+    def decode(self, data) -> List[Union[list, np.ndarray, torch.Tensor]]:
```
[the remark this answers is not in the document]
hakunanatasha: yep I think the iterable intuition hit me later as I declare that in certain areas; will change

lightwood/encoder/array/array.py
```diff
@@ -67,6 +74,11 @@ def prepare(self, train_priming_data, dev_priming_data):
         self.is_prepared = True
 
     def encode(self, column_data: Union[list, np.ndarray, torch.Tensor]) -> torch.Tensor:
```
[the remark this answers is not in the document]
hakunanatasha: yep will fix

lightwood/encoder/categorical/binary.py
```diff
@@ -1,76 +1,162 @@
...
+        """  # noqa
         if self.is_prepared:
             raise Exception('You can only call "prepare" once for a given encoder.')
 
-        for x in priming_data:
-            x = str(x)
-            if x not in self.map:
-                self.map[x] = len(self.map)
-                self.rev_map[len(self.rev_map)] = x
+        unq_cats = np.unique([i for i in priming_data if i is not None]).tolist()
```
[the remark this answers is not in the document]
hakunanatasha: `list(set(thing))`  is fine by me; probably nicer to use python internals anyway

lightwood/encoder/categorical/binary.py
```diff
@@ -1,76 +1,162 @@
...
 
-            self.index_weights = torch.Tensor(self.index_weights)
+            self.inv_target_weights = torch.Tensor([1, 1])  # Equally wt. both classes
+
+            # If target weights provided, weight by inverse
+            if self.target_weights is not None:
+
+                # Check target weights properly specified
+                if sum([np.isclose(i, 0) for i in self.target_weights.values()]) > 0:
+                    raise ValueError('Target weights cannot be 0')
```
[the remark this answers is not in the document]
hakunanatasha: If I recall correctly, `target_weights` will multiply each element by the respective component in the loss function; if we have a 0, it's akin to entirely ignoring a loss term. Pytorch already has an 'ignore_index' arg for this in most loss functions, moreover, I'm not 100% sure what this would do to numerical stability while training.

lightwood/encoder/categorical/onehot.py
hakunanatasha: list(set()) it is

lightwood/encoder/categorical/onehot.py
hakunanatasha: will use `unknown` albeit in caps

lightwood/encoder/categorical/onehot.py
hakunanatasha: got it

lightwood/encoder/text/pretrained.py
hakunanatasha: sounds good

lightwood/encoder/text/pretrained.py
hakunanatasha: we need to seriously revamp the way we use dev right now; early stopping is not a bulletproof solution, and if we dont provide a dev set, it seems to me that the more powerful strategies requiring dev simply can't run. This is as simple as asking `if dev is None -> pass`.

lightwood/encoder/categorical/binary.py
hakunanatasha: yep; I made changes in neural but this is WIP

lightwood/encoder/categorical/onehot.py
hakunanatasha: :eyes: im probably not using it correctly sigh

tests/unit_tests/encoder/categorical/test_binary.py
hakunanatasha: yes

lightwood/encoder/categorical/binary.py
hakunanatasha: fixed to `target_weights = {"class1": 0.8, "class2": 0.05, "class3": 0.15}` to avoid class ambiguity

tests/unit_tests/encoder/categorical/test_binary.py
hakunanatasha: ah must be a holdover will fix

lightwood/encoder/text/pretrained.py
George3d6: So, my argument here is that giving an option where that option isn't needed is horrible and should never be done it.

1) Allows for unit tests to be unrepresentative of production use-cases
2) Asks maintainer to deal with more complexity in the future
3) Makes the API more complicated

While not adding anything.

`if dev is None -> pass` works when dev is not used for anything, once we start using dev what are we going to do to handle this edge case?

If you so desire I'm fine with removing the dev set from all encoders period, and then if they want to do early stopping they can split their own data. I think it would be a *horrible* decision, but it is better than this.

George3d6 (changes requested):
Ok, not much to add besides the previous comment on pretrained which I think needs addressing still.

As discussed, I think it makes sense to remove unknown handling for OHE when used for targets, but that's not necessarily in the scope of this PR.

lightwood/encoder/text/pretrained.py
paxcema: How about passing an empty dev instead for cases where it's not needed? That way we keep the interface uniform and the check is now `if len(dev_priming_data) > 0:` or similar. I think the splitter is currently compatible with this as well.

lightwood/encoder/categorical/binary.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 from lightwood.encoder.base import BaseEncoder
 from lightwood.helpers.constants import _UNCOMMON_WORD
+from copy import deepcopy as dc
```
paxcema, line 5: Very small nitpick, I think `dc` is a bit nondescriptive and would suggest using it as `deepcopy`

lightwood/encoder/categorical/onehot.py
```diff
@@ -3,6 +3,7 @@
 from lightwood.helpers.log import log
 from lightwood.encoder.base import BaseEncoder
 from lightwood.helpers.constants import _UNCOMMON_WORD
+from copy import deepcopy as dc
```
paxcema, line 6: Very small nitpick, I think dc is a bit nondescriptive and would suggest using it as deepcopy

lightwood/encoder/categorical/onehot.py
hakunanatasha: im actually keeping the numpy unique() option because it does a sort that makes the output feature intuitive; fixing it means I'd have to rewrite the unit-tests so unless there is a concern for using numpy (but it seems like you mentioned earlier if it's a pain just keep as is)

lightwood/encoder/categorical/onehot.py
hakunanatasha: added

lightwood/encoder/categorical/binary.py
hakunanatasha: done

hakunanatasha:
> Ok, not much to add besides the previous comment on pretrained which I think needs addressing still.
> 
> As discussed, I think it makes sense to remove unknown handling for OHE when used for targets, but that's not necessarily in the scope of this PR.

no problem; though originally the use of holding a reserve UNK dimension in OHE allows an "the input doesnt fit any known target distribution", without this extra dimension, the model needs to choose between classes. I'm indifferent to the approach (not in the scope of this PR) as they probably require more philosophical thought around which is the better practice in general.  To add, we eventually made a mixer that explicitly handles this criteria at one point @paxcema (Where the encoded space uses an UNK token, but the mixer will re-weight across the "known" classes and fundamentally provide a vector sized by the number of classes)

lightwood/encoder/categorical/binary.py
```diff
@@ -1,76 +1,162 @@
...
+                ret.append(self.rev_map[np.argmax(vector)])
+
+            probs.append(self._norm_vec(vector))
+
+        return ret, probs, self.rev_map
 
-            if return_raw:
-                probs.append(softmax(vector).tolist())
+    @staticmethod
+    def _norm_vec(vec: List[float]):
```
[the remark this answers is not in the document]
hakunanatasha: I'm leaving it as is but I've made a note of it; when I get to generalizing helpers, let's revamp.

lightwood/encoder/categorical/onehot.py
hakunanatasha: fixed

lightwood/encoder/text/pretrained.py
hakunanatasha: This is a good suggestion and a safe alternative IMO; if it doesn't work im happy to enforce a criteria with dev but right now it's strange we just concatenate the two together and do nothing with it.

PR #758, opened 2021-11-18
SkTime mixer 2.0

### LINKED ISSUE #757: Enable `sktime.model_class` as JsonAI argument
This will let the user choose from a wide array of forecasters that [sktime](https://www.sktime.org/en/stable/api_reference/forecasting.html) offers. 

It might also be useful to automate model selection within the mixer (or, alternatively, do a `BestOf` mixer using a bunch of possible underlying forecasters).

paxcema:
## Why
To improve SkTime mixer's behavior for time series forecasting tasks.

## How

* Streamlines behavior based on `ForecastHorizon` usage, so that the user can specify (relative to latest data point seen during training) what is the starting point of the forecast that should be produced. This is done with the `PredictionArguments.forecast_offset` new argument.
* Fixes #757 with an `__init__` parameter to indicate what type of forecaster (from `sktime.forecasting`) should be used.
* Adds a test to check that automated usage through the mixer yields equal results to manual usage of the library.
* Fixes wrong offsets in `__call__()`
* Removes row dropping procedure for incomplete historical context. This burden is moved onto mixers and accuracy functions due to some mixers (like sktime) being able to leverage said rows.
* Adds docstrings to the mixer.

#### Update Nov. 25th:
* `timeseries_analyzer` now infers and stores seasonal periods for all groups. `sktime` makes use of this information for improved models. 
* Add `interval_periods` to `TimeseriesSettings` to specify period length for any given temporal sampling. For example, if you have a daily series where you expect weekly seasonality effects, you would override the default value of `1` by passing `interval_periods={'daily': 7}`.


lightwood/api/types.py
```diff
@@ -561,6 +561,7 @@ class PredictionArguments:
     fixed_confidence: Union[int, float, None] = None
     anomaly_error_rate: Union[float, None] = None
     anomaly_cooldown: int = 1
+    forecast_offset: int = 0
```
hakunanatasha, line 564: did this automatically default to 0 before?

lightwood/data/timeseries_transform.py
```diff
@@ -162,7 +162,7 @@ def transform_timeseries(
         if tss.allow_incomplete_history:
             log.warning("Forecasting with incomplete historical context, predictions might be subpar")
         else:
-            raise Exception(f'Not enough historical context to make a timeseries prediction. Please provide a number of rows greater or equal to the window size. If you can\'t get enough rows, consider lowering your window size. If you want to force timeseries predictions lacking historical context please set the `allow_incomplete_history` timeseries setting to `True`, but this might lead to subpar predictions.') # noqa
+            raise Exception(f'Not enough historical context to make a timeseries prediction ({min(group_lengths)} < {tss.window}). Please provide a number of rows greater or equal to the window size. If you can\'t get enough rows, consider lowering your window size. If you want to force timeseries predictions lacking historical context please set the `allow_incomplete_history` timeseries setting to `True`, but this might lead to subpar predictions.') # noqa
```
hakunanatasha, line 165: When this outputs does it show the numbers? If so move as follows:
"
'Not enough historical context to make a timeseries prediction. Please provide a number of rows greater or equal to the window size - currently number of rows to window size = ({min(group_lengths)} < {tss.window}). If you can\'t get enough rows, consider lowering your window size. If you want to force timeseries predictions lacking historical context please set the `allow_incomplete_history` timeseries setting to `True`, but this might lead to subpar predictions.')
"

hakunanatasha: real silly nitpick but it's just to make sense of the numbers that will just automatically pop up.

lightwood/helpers/general.py
```diff
@@ -146,11 +146,6 @@ def evaluate_array_accuracy(
             if ts_analysis['tss'].nr_predictions == 1:
                 preds = np.expand_dims(preds, axis=1)
 
-            # only evaluate accuracy for rows with complete historical context
```
hakunanatasha, line 149: how come this is omitted now?

lightwood/mixer/sktime.py
```diff
@@ -16,24 +21,51 @@ class SkTime(BaseMixer):
...
-        self.cutoff_index = {}  # marks index at which training data stops and forecasting window starts
+        self.fh = ForecastingHorizon(np.arange(1, self.n_ts_predictions + 1), is_relative=True)
         self.grouped_by = ['__default'] if not ts_analysis['tss'].group_by else ts_analysis['tss'].group_by
         self.supports_proba = False
         self.stable = True
         self.prepared = False
 
+        # load sktime forecaster
+        self.model_path = model_class
+        sktime_module = importlib.import_module('.'.join(['sktime', 'forecasting', self.model_path[0]]))
```
hakunanatasha, line 63: this is super strange - any reason why it needs to get imported as such?

lightwood/mixer/sktime.py
```diff
@@ -44,8 +76,11 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
                                for gcol in self.grouped_by} if self.ts_analysis['tss'].group_by else {}}
 
         for group in self.ts_analysis['group_combinations']:
-            # many warnings might be thrown inside of statsmodels during stepwise procedure
-            self.models[group] = self.model_class(suppress_warnings=True)
+            # ignore warnings if possible
```
hakunanatasha, line 79: might want to make suppress warnings optional with the default true

lightwood/mixer/sktime.py
```diff
@@ -95,32 +125,38 @@ def __call__(self, ds: Union[EncodedDs, ConcatedEncodedDs],
...
-            if self.grouped_by == ['__default']:
-                break
+        # apply default model in all remaining novel-group rows
+        if len(pending_idxs) > 0:
+            series = pd.Series(data['data'][list(pending_idxs)].squeeze(), index=sorted(list(pending_idxs)))
+            ydf = self._call_groupmodel(ydf, self.models['__default'], series, offset=args.forecast_offset)
 
         return ydf[['prediction']]
+
+    def _call_groupmodel(self, ydf, model, series, offset=0):
```
hakunanatasha, line 152: can you docstring this for the future maintainer ; )

hakunanatasha: it's a private method, but this signature doesn't have type hints so it will be difficult to trace what needs to go where

tests/integration/advanced/test_timeseries.py
```diff
@@ -165,6 +169,8 @@ def test_3_time_series_sktime_mixer(self):
         temporal timestamps that we need to ensure are being used correctly. In
         particular, given a train-dev-test split, any forecasts coming from a sktime
         mixer should start from the latest observed data in the entire dataset.
+        
+        This test also compares against manual use of sktime to ensure equal results.
```
hakunanatasha, line 173: LOVE this

hakunanatasha (approved):
I don't see explicit issues but I am not the TS expert; please go over comments and I'll be happy.

lightwood/api/types.py
paxcema: It did not exist before, this is new because sktime explicitly needs either an absolute or relative reference to the latest data point seen in training. This PR proposes using a relative reference.

lightwood/mixer/sktime.py
paxcema: Yes, it is so that a user can specify the underlying forecaster, e.g. they can do:

```
{
        "module": "SkTime",
        "args": {
                "stop_after": 60,
                "n_ts_predictions": "$problem_definition.timeseries_settings.nr_predictions",
                "model_class": ['ets', 'AutoETS']
        },
}
```

If they want AutoETS instead of AutoARIMA.

lightwood/helpers/general.py
paxcema: I thought this hadn't made the push, my bad. Regardless, I'm in the process of finishing a change so that ignoring these rows is now a burden on 1) the mixers and 2) the accuracy functions. Why? Because otherwise, we lose valuable data in the transform process that this mixer (and potentially others in the future) will absolutely want to use.

I promise the change isn't very big :D keep you posted

lightwood/mixer/sktime.py
paxcema: And, to be clear, I could do something like:

```
from sktime.forecasting import *
```

But I think this is better practice. Feel free to argue it's not and I'll change it.

lightwood/mixer/sktime.py
paxcema: Will leave it as non-configurable for now (warnings are really verbose, we can reactivate this is users want it).

lightwood/analysis/helpers/acc_stats.py
```diff
@@ -21,7 +21,7 @@ def analyze(self, info: Dict[str, object], **kwargs) -> Dict[str, object]:
 
         # @TODO: maybe pass ts_analysis to trigger group-wise MASE instead of R2 mean, though it wouldn't be 0-1 bounded
         info['score_dict'] = evaluate_accuracy(ns.data, ns.normal_predictions['prediction'],
-                                               ns.target, ns.accuracy_functions)
+                                               ns.target, ns.accuracy_functions, ts_analysis={'tss': ns.ts_cfg})
```
George3d6, line 24: small thing, but ahm, why call it `ts_cfg` and not `tss` or `timeseries_settings`, kinda confusing to use different names

lightwood/helpers/general.py
```diff
@@ -169,17 +167,19 @@ def evaluate_array_r2_accuracy(
...
 
-    fh = 1 if not isinstance(predictions[0], list) else len(predictions[0])
-    if fh == 1:
-        predictions = [[p] for p in predictions]
+    fh = kwargs.get('ts_analysis', {}).get('tss', None)
+    fh = fh.nr_predictions if fh is not None else 1
 
-    # only evaluate accuracy for rows with complete historical context
-    if kwargs.get('ts_analysis', {}).get('tss', False):
+    if kwargs.get('ts_analysis', {}).get('tss', False) and not kwargs['ts_analysis']['tss'].eval_cold_start:
```
George3d6, line 177: Why is `tss` part of something called `ts_analysis` here? Couldn't we be passing this directly?

lightwood/mixer/sktime.py
```diff
@@ -16,25 +21,57 @@ class SkTime(BaseMixer):
...
-        self.forecasting_horizon = np.arange(1, self.n_ts_predictions)
-        self.cutoff_index = {}  # marks index at which training data stops and forecasting window starts
+        self.fh = ForecastingHorizon(np.arange(1, self.n_ts_predictions + 1), is_relative=True)
         self.grouped_by = ['__default'] if not ts_analysis['tss'].group_by else ts_analysis['tss'].group_by
         self.supports_proba = False
         self.stable = True
         self.prepared = False
 
+        # load sktime forecaster
+        self.model_path = model_class
```
George3d6, line 62: Why re-assign `model_class` here?

lightwood/mixer/sktime.py
```diff
@@ -16,25 +21,57 @@ class SkTime(BaseMixer):
...
+        sktime_module = importlib.import_module('.'.join(['sktime', 'forecasting', self.model_path[0]]))
```
George3d6, line 63: I myself am slightly confused by this... I think it might be more intuitive if either we have a finite set of options that we just hardcode as strings (ETS, ARIMA, etc) or at least make it a string ala `ets.AutoETS`, I mean, I guess it's in the docs though and this is already a kind-of very specific mixer so maybe not.

George3d6 (approved):

lightwood/helpers/general.py
paxcema: No, because under certain conditions `ts_analysis` also includes group information and some stats that might be used inside accuracy functions like MASE.

lightwood/mixer/sktime.py
paxcema: I can't spot the reassignment 🤔 

This path is set to load the respective sktime module (by default, `model_class = ['arima', 'AutoARIMA']`).

lightwood/analysis/helpers/acc_stats.py
paxcema: That naming scheme is not part of this PR, but I agree it's confusing. I'll change it to `tss` for consistency.

lightwood/mixer/sktime.py
paxcema: I want to keep it dynamic, as people are already asking about forecasters that I'd rather not support by default (e.g. prophet). I will concede that maybe a string is simpler, will change the format.

lightwood/api/types.py
```diff
@@ -228,6 +231,8 @@ class TimeseriesSettings:
         # @TODO: George: No, I don't think it is, we need to pass this some other way
     )
     allow_incomplete_history: bool = False
+    eval_cold_start: bool = True
```
hakunanatasha, line 234: this will absolutely require a tutorial example, as it's a pretty "advanced" feature

lightwood/data/timeseries_analyzer.py
```diff
@@ -28,12 +28,13 @@ def timeseries_analyzer(data: pd.DataFrame, dtype_dict: Dict[str, str],
...
     """  # noqa
+    tss = timeseries_settings
     info = {
         'original_type': dtype_dict[target],
         'data': data[target].values
     }
-    if timeseries_settings.group_by is not None:
-        info['group_info'] = {gcol: data[gcol].tolist() for gcol in timeseries_settings.group_by}  # group col values
+    if tss.group_by is not None:
+        info['group_info'] = {gcol: data[gcol] for gcol in tss.group_by}  # group col values
```
hakunanatasha, line 37: is pd.Series a permissible format?

lightwood/data/timeseries_analyzer.py
```diff
@@ -131,3 +136,57 @@ def get_grouped_naive_residuals(info: Dict, group_combinations: List) -> Tuple[D
         group_residuals[group] = residuals
         group_scale_factors[group] = scale_factor
     return group_residuals, group_scale_factors
+
+
+def detect_period(deltas: dict, tss: TimeseriesSettings) -> dict:
```
hakunanatasha, line 141: may want to type-hint deltas better:

ex: Dict[str, float]

lightwood/helpers/general.py
```diff
@@ -107,8 +109,8 @@ def evaluate_multilabel_accuracy(true_values, predictions, **kwargs):
 
 
 def evaluate_array_accuracy(
-        true_values: List[List[Union[int, float]]],
-        predictions: List[List[Union[int, float]]],
+        true_values: pd.DataFrame,
```
hakunanatasha, line 112: is this always a pandas structure or is Iterable[float etc.] ok?

lightwood/helpers/general.py
```diff
@@ -120,21 +122,25 @@ def evaluate_array_accuracy(
...
+    wrapped_data = {
+        'data': data.reset_index(drop=True),
+        'group_info': {gcol: data[gcol].tolist()
+                       for gcol in ts_analysis['tss'].group_by} if ts_analysis['tss'].group_by else {}
+    }
+
+    if ts_analysis['tss'].group_by:
+        [true_values.pop(gby_col) for gby_col in ts_analysis['tss'].group_by]
+
+    true_values = np.array(true_values)
```
hakunanatasha, line 136: if its always pandas true_values.values should be acceptable too but nitpick, not a real issue

lightwood/helpers/general.py
```diff
@@ -120,21 +122,25 @@ def evaluate_array_accuracy(
     Scores are computed for each timestep (as determined by `timeseries_settings.nr_predictions`),
     and the final accuracy is the reciprocal of the average score through all timesteps.
     """
-
     ts_analysis = kwargs.get('ts_analysis', {})
     naive_errors = ts_analysis.get('ts_naive_mae', {})
```
hakunanatasha, line 126: this one is tricky; is there a reason naive_errors is always mae?

lightwood/mixer/sktime.py
```diff
@@ -16,25 +21,57 @@ class SkTime(BaseMixer):
...
         super().__init__(stop_after)
         self.target = target
         dtype_dict[target] = dtype.float
-        self.model_class = AutoARIMA
         self.models = {}
         self.n_ts_predictions = n_ts_predictions
         self.ts_analysis = ts_analysis
-        self.forecasting_horizon = np.arange(1, self.n_ts_predictions)
-        self.cutoff_index = {}  # marks index at which training data stops and forecasting window starts
+        self.fh = ForecastingHorizon(np.arange(1, self.n_ts_predictions + 1), is_relative=True)
```
hakunanatasha, line 55: keep as forecasting_horizon if you can or have some place describing the attributes

lightwood/mixer/sktime.py
```diff
@@ -16,25 +21,57 @@ class SkTime(BaseMixer):
...
         self.stable = True
         self.prepared = False
 
+        # load sktime forecaster
+        self.model_path = model_path
+        sktime_module = importlib.import_module('.'.join(['sktime', 'forecasting', self.model_path.split(".")[0]]))
+        try:
+            self.model_class = getattr(sktime_module, self.model_path.split(".")[1])
+        except AttributeError:
+            self.model_class = AutoARIMA
```
hakunanatasha, line 67: maybe create a log.info call here that you defaulted to AutoARIMA

also, do we want this to happen (i.e. "run something as opposed to nothing") as opposed to an error?

hakunanatasha (approved):
I don't understand enough of time-series to give very critical feedback, but a few notes:

- what is the default for forecasting? Can you recapitulate 'baseline' values from old to the new implementation?
- I noticed we switched ts_cfg to tss; is the object preserved across all instances? I know the tss timeseriessettings object passes between a few other modules.

lightwood/data/timeseries_analyzer.py
paxcema: It is, all methods downstream access it with methods from `Iterable`, so all's good.

lightwood/helpers/general.py
paxcema: Changed so that it's always pandas, but good catch -- it should be `pd.Series`, not `pd.DataFrame`

lightwood/helpers/general.py
paxcema: Yep, has to do with the definition of the MASE metric where these errors are used.

lightwood/helpers/general.py
paxcema: Hmh, you are right. I guess I'll keep this one for clarity (arguable, but as you said it's a nitpick)

paxcema:
- Default forecasting behavior for this mixer hasn't changed much. Really, the change is in correctness. Before, there was an issue with the relative index of the produced forecast, as well as no option to introduce an offset (meaning that any context other than the "ideal"/default one) would yield forecasts for an entirely different time period.
- `ts_cfg` -> `tss` is indeed preserved across the entire codebase

PR #760, opened 2021-11-24
Impute missing data inside `cleaner`

### LINKED ISSUE #759: Add imputers in cleaning phase
It should be possible for users to specify imputation methods for missing data at column-level in JsonAI objects. The best place for this is (probably) at the end of the current `lightwood.data.cleaner` logic.

For instance, specifying:

```python
json_ai = json_ai_from_problem(data, problem_definition=pdef)

json_ai.cleaner['args']['imputers'] = {
    "numeric_col": "mean",
    "categorical_col": "most_popular" 
}
```

Would trigger the above imputers for each specified column. Ideally we'd add a minimal set of useful imputers, with the option for users to define their own custom modules as well.

paxcema:
## Why

To enable greater customization in the data cleaning step. 

Closes #759.

## How

After processing each column's data and tagging malformed or missing rows as `None`, there is a new phase where any columns with user-specified imputers will replace said missing data using the logic of these imputers.

The PR introduces standard imputing methods like `mean`, `median`, `zero`, and `mode` for numerical data, as well as `mode` and `UNK` for categorical data. Complex logic should be easy to implement by following the existing `NumericalImputer` and `CategoricalImputer`.

### Notes:
- `_clean_text` has been modified to keep malformed input as `None` rather than a string `"None"`.
- For now, default behavior triggers no imputers.


lightwood/api/json_ai.py
```diff
@@ -578,6 +578,11 @@ def _add_implicit_values(json_ai: JsonAI) -> JsonAI:
         if json_ai.features[name].dependency is None:
             json_ai.features[name].dependency = []
 
+    imputers = {}
+    for col, feat in json_ai.features.items():
+        if feat.imputer is not None:
```
hakunanatasha, line 583: the approach is good but some imputers may require dependency info from other columns, which is not currently accounted for in json-ai

lightwood/api/types.py
```diff
@@ -38,10 +38,12 @@ class Feature:
     :param data_dtype: The type of information within this column (ex.: numerical, categorical, etc.)
     :param dependency: Any custom attributes for this feature that may require non-standard processing. This highly\
     depends on the encoder (ex: Pretrained text may be fine-tuned on the target; time-series requires prior time-steps).
-    """
+    :param imputer: specifies what imputer class should handle missing values. Format: "type.strategy" (e.g. "numerical.mean"). For all possible strategies, see the imputers documentation.
```
hakunanatasha, line 41: imputer type-hint should be some kind of Callable[]

lightwood/data/cleaner.py
```diff
@@ -23,17 +24,19 @@ def cleaner(
     mode: str,
     timeseries_settings: TimeseriesSettings,
     anomaly_detection: bool,
+    imputers: Dict[str, str] = {},
```
hakunanatasha, line 27: imputer feels like it should be Dict[str, callable]

lightwood/data/cleaner.py
```diff
@@ -227,8 +236,13 @@ def _clean_quantity(element: object) -> Optional[float]:
 # ------------------------- #
 # Text
 # ------------------------- #
-def _clean_text(element: object) -> str:
-    return str(element)
+def _clean_text(element: object) -> Union[str, None]:
```
hakunanatasha, line 239: you may actually want to keep the `_clean_text` fix though in a separate PR

hakunanatasha (changes requested):
In theory, this is a good idea but needs work. Here are some notes from discussion + prior review:

(1) Some imputers require the full table/other column dependency information (i.e. SMOTE); a per-feature basis doesn't account for that
(2) We may want to create a separate stackable submodule like `lightwood.imputer` that runs between cleaner/splitter
(3) It may also be fine to do something like `lightwood.cleaner.impute` as a sub-function; I'd almost prefer making the cleaner its own class that can be inherited, with some private methods that handle the internal-col filtering/defaults.
(4) **You may need to make a separate mini-PR for ``_clean_text``**

hakunanatasha:
As discussed by @paxcema this PR introduces a new feature that may require more design consideration. Please feel free to add more details if time permits.

lightwood/api/json_ai.py
```diff
@@ -635,6 +637,8 @@ def _add_implicit_values(json_ai: JsonAI) -> JsonAI:
     }
 
     for field_name, implicit_value in hidden_fields.items():
+        if getattr(json_ai, field_name):
```
paxcema, line 640: While this is not needed, I do think it's good in case a user wants to specify "hidden" module args beforehand.

lightwood/helpers/imputers.py
```diff
@@ -0,0 +1,89 @@
...
+        
+        :param value: The value to impute. One of 'mean', 'median', 'zero', 'mode'.
+        :param typecast: Used to cast the column into either 'int' or 'float' (`None` skips forced casting).
+        """  # noqa
+        self.value = value
+        self.typecast = typecast
+        super().__init__(target, dependencies)
+
+    def impute(self, data: pd.DataFrame) -> pd.DataFrame:
+        data = deepcopy(data)
```
George3d6, line 41: Consider cleaning the data here and potentially importing some primitives from the statistical analysis module for doing so (or make them shared?)

lightwood/helpers/imputers.py
```diff
@@ -0,0 +1,89 @@
...
+                    raise Exception(f'Numerical imputer failed to cast column {col} to float!')
+            else:
+                raise Exception(f'Numerical imputer used in non-numeric column {col} with dtype {data[col].dtype}!')
+
+        if self.value == 'mean':
+            value = data[col].dropna().mean()
+        elif self.value == 'median':
+            value = data[col].dropna().median()
+        elif self.value == 'mode':
+            value = data[col].dropna().mode().iloc[0]  # if there's a tie, this chooses the smallest value
```
George3d6, line 58: instead of individually cleaning it here, I mean

lightwood/helpers/imputers.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            value = 0.0
+
+        data[col] = data[col].fillna(value=value)
```
George3d6, line 62: Are we use `na` values will cover all of the values we want to impute? If not it would be better to loop 1-by-1 at the beginning, remove "incorrect" values, store their index, then impute here.

But maybe that's overkill.

lightwood/helpers/imputers.py
```diff
@@ -0,0 +1,89 @@
...
+        super().__init__(target, dependencies)
+
+    def impute(self, data: pd.DataFrame) -> pd.DataFrame:
+        data = deepcopy(data)
+        col = self.target
+
+        if self.value == 'mode':
+            value = data[col].dropna().mode().iloc[0]
+        else:
+            value = 'UNK'
```
George3d6, line 86: `None` is the default value for Unkown, it gets converted into an unknown string token inside the OHE/CAE, change `'UNK'` to `None`

lightwood/api/json_ai.py
```diff
@@ -750,8 +759,14 @@ def code_from_json_ai(json_ai: JsonAI) -> str:
     # Pre-processing Body
     # ----------------- #
 
+    imputers = ""
+    for target_col, imputer in json_ai.imputers.items():
+        imputers += f"{target_col}: {imputer},"
+    imputers = "{" + imputers + "}"
```
George3d6, line 765: We have an `inline_dict` function that would help here (I think) + I think this and the loop at line 669 can be one and the same, with the imputers being constructed with `call` here

lightwood/data/cleaner.py
```diff
@@ -51,6 +54,10 @@ def cleaner(
         # If you want to customize the cleaner, it's likely you can to modify ``get_cleaning_func``
         data[col] = data[col].apply(get_cleaning_func(dtype_dict[col], custom_cleaning_functions))
 
+    for col, imputer in imputers.items():
+        cols = [col] + [col for col in imputer.dependencies]
+        data[col] = imputer.impute(data[cols])
```
George3d6, line 59: Should we call this after `clean_timeseries` ?

George3d6 (changes requested):
Some minor issues and tests failing, overall looks nice.

lightwood/helpers/imputers.py
paxcema: You may be right. What "incorrect" values do you suggest other than what is captured by `fillna` (e.g. `nan` and `None`)?

paxcema: Ok, I will leave this as-is given this imputer is more of a basic example (indeed, both are disabled by default).

PR #761, opened 2021-11-24
Do not drop single-group-by column (fix #756)

### LINKED ISSUE #751: Don't allow TS predictions if there are historical columns missing
Current behavior for time series forecasting tasks allows predicting with missing historical columns (except for the order-bys). We cannot guarantee mixers will be able to cope with the missing data, so we should put a hard restriction that it's not possible to predict if any of the historical columns is missing (or has `None/nan` values).

### LINKED ISSUE #756: Error if train TS predictor on data with single value in 'group_by'
I tryed train TS predictor on data:
```
df = pd.DataFrame({
            'group': ["A" for _ in range(100, 210)],
            'order': [x for x in range(100, 210)],
            'x1': [x for x in range(100, 210)],
            'x2': [x * 2 for x in range(100, 210)],
            'y': [x * 3 for x in range(100, 210)]
        })
```
I set `group_by` to `group`. I got error:
```
  File "/home/maxs/dev/mdb/venv38/sources/mindsdb/mindsdb/interfaces/model/learn_process.py", line 100, in run_learn
    run_fit(predictor_id, df)
  File "/home/maxs/dev/mdb/venv38/sources/mindsdb/mindsdb/utilities/functions.py", line 56, in wrapper
    return func(*args, **kwargs)
  File "/home/maxs/dev/mdb/venv38/sources/mindsdb/mindsdb/interfaces/model/learn_process.py", line 87, in run_fit
    raise e
  File "/home/maxs/dev/mdb/venv38/sources/mindsdb/mindsdb/interfaces/model/learn_process.py", line 67, in run_fit
    predictor.learn(df)
  File "/tmp/5ebf11ab5e9ba76011bd4d012f7779ef3ad842556160a13916371366856095033.py", line 397, in learn
    data = self.preprocess(data)
  File "/tmp/5ebf11ab5e9ba76011bd4d012f7779ef3ad842556160a13916371366856095033.py", line 137, in preprocess
    data = transform_timeseries(
  File "/home/maxs/dev/mdb/venv38/sources/lightwood/lightwood/data/timeseries_transform.py", line 105, in transform_timeseries
    for _, df in original_df.groupby(gb_arr):
  File "/home/maxs/dev/mdb/venv38/lib/python3.8/site-packages/pandas/core/frame.py", line 7631, in groupby
    return DataFrameGroupBy(
  File "/home/maxs/dev/mdb/venv38/lib/python3.8/site-packages/pandas/core/groupby/groupby.py", line 889, in __init__
    grouper, exclusions, obj = get_grouper(
  File "/home/maxs/dev/mdb/venv38/lib/python3.8/site-packages/pandas/core/groupby/grouper.py", line 862, in get_grouper
    raise KeyError(gpr)
KeyError: 'group'
```
As i see in debugger, `original_df` in `.../timeseries_transform.py", line 105` does not have 'group' column. I guess it was deleted because it have only single value. 
If it so, then i think we should not remove such column if user explicitly use it in 'group_by'

paxcema:
## Why

See #756 for more details.

## How

Adding `tss.group_by` columns to list of exceptions that should not be dropped.

Unrelated, this fixes sktime mixer's behavior so that it outputs a single number (instead of a list wrapping around it) if `nr_predictions == 1`.

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -185,17 +185,18 @@ def generate_json_ai(
     exec(IMPORT_EXTERNAL_DIRS, globals())
     target = problem_definition.target
     input_cols = []
+    tss = problem_definition.timeseries_settings
+
     for col_name, col_dtype in type_information.dtypes.items():
         if (
                 (col_name not in type_information.identifiers
                  and col_dtype not in (dtype.invalid, dtype.empty)
                  and col_name != target)
                 or
-                col_name in problem_definition.timeseries_settings.group_by
+                (tss.group_by is not None and col_name in tss.group_by)
         ):
             input_cols.append(col_name)
 
-    tss = problem_definition.timeseries_settings
     is_target_predicting_encoder = False
     is_ts = problem_definition.timeseries_settings.is_timeseries
     # Single text column classification
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -187,9 +187,11 @@ def generate_json_ai(
     input_cols = []
     for col_name, col_dtype in type_information.dtypes.items():
         if (
-            col_name not in type_information.identifiers
-            and col_dtype not in (dtype.invalid, dtype.empty)
-            and col_name != target
+                (col_name not in type_information.identifiers
+                 and col_dtype not in (dtype.invalid, dtype.empty)
+                 and col_name != target)
+                or
+                col_name in problem_definition.timeseries_settings.group_by
         ):
             input_cols.append(col_name)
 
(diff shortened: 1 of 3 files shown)


lightwood/api/json_ai.py
```diff
@@ -185,15 +185,18 @@ def generate_json_ai(
...
     for col_name, col_dtype in type_information.dtypes.items():
         if (
-            col_name not in type_information.identifiers
-            and col_dtype not in (dtype.invalid, dtype.empty)
-            and col_name != target
+                (col_name not in type_information.identifiers
+                 and col_dtype not in (dtype.invalid, dtype.empty)
+                 and col_name != target)
+                or
+                (tss.group_by is not None and col_name in tss.group_by)
```
hakunanatasha, line 196: how did this get missed before? Isn't it by definition not an : [identifier, invalid, empty, target] type?

hakunanatasha (approved):
minor question on how the tss_groupby got missed to begin with but it seems fine otherwise.

lightwood/api/json_ai.py
paxcema: Nope, it will get tagged as `invalid` (if I recall correctly) when there's a single value in the column.

Issue #719, opened 2021-11-02
Streamline `PredictorInterface.adjust()` signature


paxcema:
We should avoid passing dictionaries with (kind of arbitrary) keys as arguments in the API. Instead, opting for:

```python
def adjust(self, old_data: pd.DataFrame, new_data: pd.DataFrame) -> None:
...
```

may be better.

PR #762, opened 2021-11-25
Streamlined predictor.adjust signature

paxcema:
## Why

For a simplified signature in `predictor.adjust()`. "Fixes" #719. 

## How
Nothing weird going on. Simple modifications to `JsonAI` and `PredictorInterface`.

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -979,8 +979,8 @@ self.mode = 'train'
 # Extract data
 # --------------- #
 # Extract the featurized data
-encoded_old_data = new_data['old']
-encoded_new_data = new_data['new']
+encoded_old_data = old_data if old_data is not None else pd.DataFrame()
+encoded_new_data = new_data
 
 # --------------- #
 # Adjust (Update) Mixers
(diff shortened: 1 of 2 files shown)


lightwood/api/json_ai.py
```diff
@@ -979,8 +979,8 @@ def code_from_json_ai(json_ai: JsonAI) -> str:
 # Extract data
 # --------------- #
 # Extract the featurized data
-encoded_old_data = new_data['old']
```
hakunanatasha, line 982: the reason the signature here was this way was because the output of the splitter is a dict

lightwood/api/json_ai.py
```diff
@@ -1030,9 +1030,7 @@ def code_from_json_ai(json_ai: JsonAI) -> str:
 if self.problem_definition.fit_on_all:
 
     log.info("Adjustment on validation requested.")
-    update_data = {{"new": enc_train_test["test"], "old": ConcatedEncodedDs([enc_train_test["train"], enc_train_test["dev"]])}}  # noqa
-
-    self.adjust(update_data)
+    self.adjust(enc_train_test["test"], ConcatedEncodedDs([enc_train_test["train"], enc_train_test["dev"]]))
```
hakunanatasha, line 1033: we need to revamp concatenatedencodedds

also the data structure is misspelled : /

hakunanatasha: self consistently fine; it's called this in the script that defines the class

hakunanatasha (approved):

PR #763, opened 2021-11-25
Improve time aim

George3d6:
* Fixed the way a predictor's runtime is computed, won't affect Github much but will affect running on multi-core machines a lot.
* Moved the function to time a predictor's training (and fail if it tranins too slowly compared to the aim) in a separate file
* EVERY single call to `.learn` in the tests is now timed using aforementioned function (still have to give it 2.5x slack to pass on github, 1.5x works locally, github machines are really unrepresentative of reality)
* Imporved how time aim is set for the predictor (it tries to take into account the time it took for the type and statistical analysis to run and decrease it by 2x that, assuming that the other untimed steps will take around the same time)
* Fixed time budget computation for encoders, before it would have always assigned an unreasonably huge value ( 66% of training time)
* Fixed types to be `Option[T]` instead of `Union[T, None]` when the variable can be assigned `None`, just for the sake of keeping convention
* Time aim works better in the `Pretrained` Language Encoder and the `Neural` Mixer by breaking *druing* an iteration if the time is expanded, not only when the epoch is done
* Time aim works slightly better for the LightGBM mixer (and hopefully LGBM Array) by discounting "preparation" time (which can be quite significant) from the total time budget
* Regression now takes time aim into account by not running if the time it has seem insufficient
* Removed a test that wasn't being ran (missing `__init__.py` in the directory) and was deprecated
* Removed all usages of `FileDS` from tests, removed `mindsdb_datasources` testing dependency as a result
* Commented out one of the text tests, I think they are both more or less the same and take loads of time, so for now, when our compute is so shitty, I'd say let's just run one
* Increase `time_aim` in some of the tests to slightly more sensible level, time aim of 10, 20 seconds is unrealistic, all of them are now > 30, doesn't seem to have affected runtime much, tests run slightly faster after this PR
* Added a rule that the minimal time aim will be 10 seconds, otherwise we are playing around with numbers that are too small even for the tinniest of datasets and will pointlessly run into bug, might be wise to even make this higher (5 or 10 minutes?)


tests/integration/strange_data/test_nested_dataset.py
```diff
@@ -1,119 +0,0 @@
-from copy import deepcopy
```
paxcema, line 1: Question: is this one deprecated because the .json -> DataFrame burden is on MindsDB? Or is it something else?

tests/integration/advanced/test_timeseries.py
```diff
@@ -209,7 +198,7 @@ def test_3_time_series_sktime_mixer(self):
         code = code_from_json_ai(json_ai)
         predictor = predictor_from_code(code)
 
-        predictor.learn(train)
+        train_and_check_time_aim(predictor, train)
```
paxcema, line 201: The one big worry I have is that I don't see (yet) a way to make `sktime` compliant with time aim... I'll give it some more thought, but maybe for now a disclaimer at the start of its `fit()` method could be good for the user to notice.

lightwood/mixer/neural.py
```diff
@@ -302,14 +305,15 @@ def partial_fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
...
+        self.started = time.time()
         train_dl = DataLoader(train_data, batch_size=self.batch_size, shuffle=True)
         dev_dl = DataLoader(dev_data, batch_size=self.batch_size, shuffle=True)
         optimizer = self._select_optimizer()
         criterion = self._select_criterion()
         scaler = GradScaler()
 
         self.model, _, _ = self._max_fit(train_dl, dev_dl, criterion, optimizer, scaler,
-                                         self.stop_after, max(1, int(self.epochs_to_best / 3)))
+                                         self.stop_after * 0.1, max(1, int(self.epochs_to_best / 3)))
```
paxcema, line 316: To confirm: this implies final tuning uses 10% of the budget?

lightwood/api/types.py
```diff
@@ -351,7 +352,12 @@ def from_dict(obj: Dict):
         unbias_target = obj.get('unbias_target', True)
         seconds_per_mixer = obj.get('seconds_per_mixer', None)
         seconds_per_encoder = obj.get('seconds_per_encoder', None)
+        expected_additional_time = obj.get('expected_additional_time', None)
+
         time_aim = obj.get('time_aim', None)
+        if time_aim is not None and time_aim < 10:
+            log.warning(f'Your specified time aim of {time_aim} is too sort. Setting it to 10 seconds.')
```
paxcema, line 359: Typo: too sHort

lightwood/api/json_ai.py
```diff
@@ -391,19 +392,20 @@ def generate_json_ai(
...
             problem_definition.seconds_per_encoder = 0
         else:
             problem_definition.seconds_per_encoder = int(
-                problem_definition.time_aim
+                (problem_definition.time_aim - problem_definition.expected_additional_time)
                 * (encoder_time_budget_pct / nr_trainable_encoders)
             )
         problem_definition.seconds_per_mixer = int(
-            problem_definition.time_aim * ((1 / encoder_time_budget_pct) / nr_mixers)
+            (problem_definition.time_aim - problem_definition.expected_additional_time)
```
paxcema, line 407: Is this saying that ensembling + analysis are not included in the time aim calculations? If so, add it to the docstring in `ProblemDefinition`.

lightwood/api/high_level.py
```diff
@@ -50,12 +50,24 @@ def json_ai_from_problem(df: pd.DataFrame, problem_definition: Union[ProblemDefi
...
     stats = statistical_analysis(
         df, type_information.dtypes, type_information.identifiers, problem_definition)
+
+    duration = time.time() - started
+    if problem_definition.time_aim is not None:
+        problem_definition.time_aim -= duration
+        if problem_definition.time_aim < 10:
+            problem_definition.time_aim = 10
+
+    # Assume that the stuff besdies encoder and mixers takes about as long as analyzing did... bad, but let's see
```
paxcema, line 68: I think it would be important to add this into the `ProblemDefinition` docstring:

> expected_additional_time: stuff besides encoder and mixers

Also, typo: besides*

hakunanatasha: done

lightwood/api/json_ai.py
```diff
@@ -391,19 +392,20 @@ def generate_json_ai(
             ]
         )
         nr_mixers = len(list(outputs.values())[0].mixers)
-        encoder_time_budget_pct = max(
-            3.3 / 5, 1.5 + np.log(nr_trainable_encoders + 1) / 5
-        )
+        encoder_time_budget_pct = (2 + np.log(nr_trainable_encoders + 2)) / 5
```
hakunanatasha, line 395: Where are these particular numbers coming from? Are they empirical? Empirical is fine, but it's worth documenting the notes of why we choose these and on what datasets did we assume these numbers gave us something useful.

lightwood/api/types.py
hakunanatasha: corrected

lightwood/mixer/neural.py
```diff
@@ -282,8 +284,9 @@ def fit(self, train_data: EncodedDs, dev_data: EncodedDs) -> None:
         criterion = self._select_criterion()
         scaler = GradScaler()
 
-        self.model, epoch_to_best_model, err = self._max_fit(
-            train_dl, dev_dl, criterion, optimizer, scaler, self.stop_after, return_model_after=20000)
+        self.model, epoch_to_best_model, _ = self._max_fit(
+            train_dl, dev_dl, criterion, optimizer, scaler, self.stop_after * 0.8 - (time.time() - self.started),
```
hakunanatasha, line 288: @paxcema mentions 10%; i'm assuming this is within 80% of the budget? It's worth keeping a documentation note

tests/integration/basic/test_categorical.py
```diff
@@ -17,7 +17,7 @@ def setup_predictor(self, df, target):
         predictor = predictor_from_problem(df, ProblemDefinition.from_dict(
             {'target': target, 'time_aim': 20, 'unbias_target': False}))
 
-        predictor.learn(train)
+        train_and_check_time_aim(predictor, train)
```
hakunanatasha, line 20: do we need to do a time-aim check on ALL unit-tests? If there is an issue with time_aim (for whatever reason) downstream, every unit-test will fail. I don't know if it's helpful to have a condition where we can't isolate which problem is there.

hakunanatasha (changes requested):
In theory I'm ok with this PR but a few notes that require fixes:

(1) Please annotate where you budgeted time; ex 80% allocated to mixers or 10% to analysis etc; @paxcema and I had both comments
(2) I'm unsure of whether `train_and_check_time` for every unit-test is a good idea (I'm open to discussion); this leads to a problem where if for some reason there is something wrong with that check, every test in `integration` would fail, when in fact we may want to have reserved tests isolate each component of the end-to-end compiling like `integration` has. We need a unit-test for this, but it may be a separate case? If you can make a strong argument, I'm willing to reconsider.

George3d6:
Note: Only add time aim check for timeseries and one other "normal" integration test.

paxcema:
Regarding time series: add this only to tests where the sktime mixer is not used (as of now it does not stop)

George3d6:
- Every trainable component should get the same time budget + annotate everything better.

lightwood/mixer/neural.py
George3d6: Yes

tests/integration/strange_data/test_nested_dataset.py
George3d6: Yes, exactly.

hakunanatasha (commented):
Did you remove FileDS or simply omit them from the import calls?

Lastly, can you make a condition where `sktime` doesn't trigger time-aim as per @paxcema 's comments?

George3d6:
The FileDS dependency was `mindsdb_datasources`, which is indeed removed from the test requirements.

Not sure what you mean by:

> Lastly, can you make a condition where sktime doesn't trigger time-aim as per @paxcema 's comments?

George3d6:
As discussed, fix for sktime is separate, see here: https://github.com/mindsdb/lightwood/issues/779 Merging this now.

PR #764, opened 2021-11-25
Revamp Notebooks, test docs

George3d6:
-Parse out all the code from the README and run it during testing in order to make sure it's always valid
- Run jupyter notebooks every time we build the docs and make doc building part of the test suite
- Added load_custom_module function to API, takes a file path and create modules dir (if needed) and moves (or replaces) it in that dir
- Removed pre-built notebooks so that they always have to be newly generated in order to be added to the docs
- Added instructions for building new tutorials (based on how current tutorials had to be standardized)
- Revamped every single notebook to: 1) actually work 2) work without any supporting files (those are written as the notebook runs using a `%%` command

New PR since I mangled the history in the previous one.


hakunanatasha (commented):
NS minor notes: some minor grammar fixes in the readme and added the import of load_custom_module to the api init.

I am close to approving this, but I noticed what could be a bug;

the `truth` values from the predicted outputs are not working anymore; in some cases, this may not be possible (as a user can provide some small component of the data), but in cases where a test_df is a subset of the original df, you should be able to see ground-truth v. predicted values from the predict() call. This is most noted in `tutorial_update_model` where originally the truth values were defined. If you can check this, I'm happy to approve and merge.

George3d6:
`truth` should had been removed entirely by this PR but when re-making the changes that bit got messed up

hakunanatasha (approved):
This is fine so far; the truth cols are removed fine and the init now contains the new load_module fxn.

While outside the scope of the PR, we need a few more added details to the custom_mixer tutorial explaining (1) what the confidence estimate means and (2) an example where you compare results of the prediction with a test dataset (something the model has never seen before) and compare between the prediction and test.

hakunanatasha:
@George3d6 before you check this; can you look into why `test_categorical` is passing? There is an assert that explicitly checks if the `truth` category exists in the columns but this was omitted.

hakunanatasha:
JK you removed it in the unittest ALL GOOD

PR #767, opened 2021-11-27
Automated model selection for sktime mixer

paxcema:
## Why

To automate model/forecaster selection so that `sktime` mixer forecasts are improved.

## How

Use an Optuna study that aims at minimizing the sMAPE error over the dev data split. For now, model selection is done based on the global error instead of the group-wise error (this could be changed in the future, but note it is pretty expensive and won't scale for high cardinality problems).

Considered forecasting algorithms are `AutoARIMA`, `AutoETS` and `ThetaForecaster`.


lightwood/mixer/sktime.py
```diff
@@ -21,74 +23,117 @@ class SkTime(BaseMixer):
...
 
-        Forecaster type can be specified by providing the `model_class` argument in `__init__()`.
+        Forecaster type can be specified by providing the `model_class` argument in `__init__()`. It can also be determined by hyperparameter optimization based on dev data validation error.
         """  # noqa
         log.info('Started fitting sktime forecaster for array prediction')
 
-        all_subsets = ConcatedEncodedDs([train_data, dev_data])
-        df = all_subsets.data_frame.sort_values(by=f'__mdb_original_{self.ts_analysis["tss"].order_by[0]}')
+        if self.hyperparam_search:
+            optuna.logging.set_verbosity(0)
```
paxcema, line 88: remove this

hakunanatasha (approved):

PR #769, opened 2021-11-29
OH and Binary Encoders weighting fix

George3d6:
What's the bug here?

The One hot encoder and the binary were **inverting** the target weights argument when creating the weights-per-index tensor (used by the Neural mixer).

Why was this not affecting accuracy?

- Because we never weighted manually, only automatically, and instead of passing the *reverse* occurrence of every class to the encoders we were passing the occurrence of every class ... so double investing => all is good

This fix means that the `target_weight` argument now works with the `Neural` encoder *and* might increase accuracy in the case where implicit weights were passed to LightGBM, since the latest release it takes weight into account.

I'm actually surprised we didn't see more of an accuracy drop due to passing the wrong weights to LGBM in the previous PR, but I guess it shows that Neural is able to pick up the slack... and probably that we need a better benchmarks methodology.

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -119,7 +119,7 @@ def lookup_encoder(
             if problem_defintion.unbias_target:
                 encoder_dict["args"][
                     "target_weights"
-                ] = "$statistical_analysis.target_class_distribution"
+                ] = "$statistical_analysis.assumed_target_class_weight"
             if problem_defintion.target_weights is not None:
                 encoder_dict["args"][
                     "target_weights"
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -119,7 +119,7 @@ def lookup_encoder(
             if problem_defintion.unbias_target:
                 encoder_dict["args"][
                     "target_weights"
-                ] = "$statistical_analysis.assumed_target_class_weight"
+                ] = "$statistical_analysis.suggested_target_class_weight"
             if problem_defintion.target_weights is not None:
                 encoder_dict["args"][
                     "target_weights"
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -119,7 +119,7 @@ def lookup_encoder(
             if problem_defintion.unbias_target:
                 encoder_dict["args"][
                     "target_weights"
-                ] = "$statistical_analysis.suggested_target_class_weight"
+                ] = "$statistical_analysis.target_weights"
             if problem_defintion.target_weights is not None:
                 encoder_dict["args"][
                     "target_weights"
--- a/lightwood/api/types.py
+++ b/lightwood/api/types.py
@@ -174,6 +174,7 @@ class StatisticalAnalysis:
     df_target_stddev: Optional[float]
     train_observed_classes: object  # Union[None, List[str]]
     target_class_distribution: object  # Dict[str, float]
+    assumed_target_class_weight: object  # Dict[str, float]
     histograms: object  # Dict[str, Dict[str, List[object]]]
     buckets: object  # Dict[str, Dict[str, List[object]]]
     missing: object
--- a/lightwood/api/types.py
+++ b/lightwood/api/types.py
@@ -161,7 +161,7 @@ class StatisticalAnalysis:
     :param df_target_stddev: The standard deviation of the target of the dataset
     :param train_observed_classes:
     :param target_class_distribution:
-    :param suggested_target_class_weight: What weight the analysis suggests to assign each class by in the case of classification problems. Note: target_weights in the problem definition overides this
+    :param target_weights: What weight the analysis suggests to assign each class by in the case of classification problems. Note: target_weights in the problem definition overides this
     :param histograms:
     :param buckets:
     :param missing:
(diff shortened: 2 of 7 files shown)


lightwood/encoder/categorical/onehot.py
```diff
@@ -79,7 +79,8 @@ def prepare(self, priming_data, max_dimensions=20000):
                 self.target_weights[UNCOMMON_WORD] = uncommon_weight
             for word in set(priming_data):
                 if self.target_weights is not None:
-                    self.index_weights[self._lang.word2index[str(word)]] = 1 / self.target_weights[word]
+                    self.index_weights[self._lang.word2index[str(word)]] = \
+                        self.target_weights[word] / np.max(list(self.target_weights.values()))
```
hakunanatasha, line 83: see above; you will accidentially down-weight minority classes

hakunanatasha (changes requested):
can you double check your logic around weighting? Index_weights is supposed to represent the inverse weights (Which is fine). You have right now something that will continue to exacerbate minority issues because it will do class_counts_of_class_i / max(class_counts)

lightwood/api/json_ai.py
```diff
@@ -119,7 +119,7 @@ def lookup_encoder(
             if problem_defintion.unbias_target:
                 encoder_dict["args"][
                     "target_weights"
-                ] = "$statistical_analysis.target_class_distribution"
+                ] = "$statistical_analysis.assumed_target_class_weight"
```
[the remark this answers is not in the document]
George3d6: because it's the inverse frequency,  but even then, we might want to have another technique for weighting (e.g. something on a log scale later)

lightwood/api/types.py
```diff
@@ -174,6 +174,7 @@ class StatisticalAnalysis:
     df_target_stddev: Optional[float]
     train_observed_classes: object  # Union[None, List[str]]
     target_class_distribution: object  # Dict[str, float]
```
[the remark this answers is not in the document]
George3d6: See comment above, this is the default weighting for classes, which in the future we might actually want to be something else, at least that was my thinking.

lightwood/api/types.py
```diff
@@ -174,6 +174,7 @@ class StatisticalAnalysis:
     df_target_stddev: Optional[float]
```
paxcema, line 174: I think for this it would be super helpful to add the meaning/purpose of `target_class_distribution` vs `assumed_target_class_weight` to avoid user confusion

lightwood/api/json_ai.py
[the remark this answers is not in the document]
paxcema: If this is customizable by the user, then I think "assumed" is not needed in the name here.

paxcema (approved):
I'd be okay with this fix as long as the actual meaning of the params makes its way into the docstring of the stat analyzer.

George3d6:
@paxcema I renamed that since both you and @hakunanatasha commented on it *and* added a docstring for clarity. Let me know if it makes more sense now.

paxcema:
Does make sense, thanks :+1:

lightwood/api/json_ai.py
```diff
@@ -119,7 +119,7 @@ def lookup_encoder(
             if problem_defintion.unbias_target:
                 encoder_dict["args"][
                     "target_weights"
-                ] = "$statistical_analysis.target_class_distribution"
+                ] = "$statistical_analysis.suggested_target_class_weight"
```
hakunanatasha, line 122: just call it target_class_weights versus target_class_distributions. Target_class_weights will always refer to what is passed to the model henceforth.

lightwood/api/types.py
```diff
@@ -161,19 +161,21 @@ class StatisticalAnalysis:
     :param df_target_stddev: The standard deviation of the target of the dataset
     :param train_observed_classes:
     :param target_class_distribution:
+    :param suggested_target_class_weight: What weight the analysis suggests to assign each class by in the case of classification problems. Note: target_weights in the problem definition overides this
```
hakunanatasha, line 164: please just call it `Target_class_weights`

lightwood/api/types.py
hakunanatasha: call distribution == number of instances of each class; and weights == assigned importance to each class (whether this is related to the weights or not doesnt matter)

lightwood/api/types.py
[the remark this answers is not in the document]
hakunanatasha: in the future we will either change this, or a user will just implement their own approach

lightwood/api/types.py
```diff
@@ -161,19 +161,21 @@ class StatisticalAnalysis:
...
     :param avg_words_per_sentence:
     :param positive_domain:
-    """
+    """ # noqa
 
     nr_rows: int
     df_target_stddev: Optional[float]
     train_observed_classes: object  # Union[None, List[str]]
     target_class_distribution: object  # Dict[str, float]
+    suggested_target_class_weight: object  # Dict[str, float]
```
hakunanatasha, line 178: replace to `Target_class_weights`

lightwood/data/statistical_analysis.py
```diff
@@ -133,8 +133,12 @@ def statistical_analysis(data: pd.DataFrame,
 
     # get observed classes, used in analysis
     target_class_distribution = None
+    suggested_target_class_weight = None
```
hakunanatasha, line 136: replace to `target_class_weights`

hakunanatasha (changes requested):
please change 'suggested_target_weights' to just `target_class_weights` and then I will approve.

hakunanatasha:
One note - please make an edge case/unit-test/assert that handles if a count is 0 for some reason, as this will assuredly crop up sometime.

PR #772, opened 2021-12-03
Revert "Extend default layout to include GTM scripts"; doc builds failed

hakunanatasha:


hakunanatasha:
merged; does not affect unit-tests as it only deletes a file

PR #774, opened 2021-12-03
Modify update model tutorial

hakunanatasha:
Fixes an incorrect call in the update_model tutorial


hakunanatasha:
The tutorial update function does not work as intended; this needs to be revisited when time permits.

paxcema:
I didn't notice the tutorial was scraped. I think going forward we should fix them ASAP or maybe set them somehow as inactive so that we don't forget they are missing in the first place.

I've changed the API slightly so that `adjust()` usage is now internally consistent with the hinted types, and the tutorial seems to work fine (though I'll let tests decide this).

PR #776, opened 2021-12-06
Precision recall and f1

George3d6:
Added the aforementioned scores, they require some assertions for their usage to make sense *and* they require `int` conversion for sklearn interface compatibility.

In theory, we could have a more generic recall_for_{label} score, but for now, we are sticking with the "standard" that defines true/1 as "positive" and false/0 as negative.

### DIFF
--- a/lightwood/helpers/accuracy.py
+++ b/lightwood/helpers/accuracy.py
@@ -7,13 +7,21 @@ from sklearn.metrics import precision_score as sk_precision_score
 
 def to_binary(y: Iterable) -> List[int]:
     try:
-        y = [int(x) for x in y]
+        y_binarized = []
+        for ele in y:
+            if str(ele).lower() == 'true':
+                y_binarized.append(1)
+            elif str(ele).lower() == 'false':
+                y_binarized.append(0)
+            else:
+                y_binarized.append(int(ele))
+
         assert len(set(y)) < 3
         assert 1 in y
         assert 0 in y
     except Exception:
         raise Exception('To use precision, recall or f1 please make sure your target consists only of 1s and 0s')
-    return y
+    return y_binarized
 
 
 def f1_score(y_true, y_pred) -> float:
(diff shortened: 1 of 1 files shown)


lightwood/helpers/accuracy.py
```diff
@@ -1,4 +1,38 @@
...
+        for ele in y:
+            if str(ele).lower() == 'true':
+                y_binarized.append(1)
+            elif str(ele).lower() == 'false':
+                y_binarized.append(0)
+            else:
+                y_binarized.append(int(ele))
+
+        assert len(set(y_binarized)) < 3
+        assert 1 in y_binarized or 0 in y_binarized
```
hakunanatasha, line 20: I feel like this should be "and" but I get the point which is there should be at least one instance of both "1" and "0".

lightwood/helpers/accuracy.py
```diff
@@ -1,4 +1,38 @@
+from typing import Iterable, List
 from sklearn.metrics import r2_score as sk_r2_score
+from sklearn.metrics import f1_score as sk_f1_score
+from sklearn.metrics import recall_score as sk_recall_score
+from sklearn.metrics import precision_score as sk_precision_score
+
+
+def to_binary(y: Iterable) -> List[int]:
```
hakunanatasha, line 8: is "y" always a list of True/False? If so, can you not just do a map like `{False: 0, True: 1}` or is it because lightwood converts categorical (including binary) info into a str type?

hakunanatasha (approved):
The only thing I'm confused about is to_binary being in there since it explicitly looks for str/bool of True/False requiring some other pre-processing step ahead. Is there a reason we introduce this function and don't just call a binary encoder on it?

lightwood/helpers/accuracy.py
George3d6: I think there's cases where we only predict 0 or 1 and this accounts for that.

lightwood/helpers/accuracy.py
George3d6: The issue is that possible values for false are:
`'false'`, `'FALSE'`, `False`, `'False'`,`0` and `'0'`... and due to pd arbitrary type casting we can't guarantee what is what and due to user variability in how they declare false/true string we can't either (same for true labels)

This jus standardized that

We could do a map here, but I think the iteration is more obvious since it's a slightly complex operation, not just a simple map it's basically `lower -> map if -> cast else`

PR #778, opened 2021-12-07
Make image encoder optional

George3d6:


hakunanatasha (commented):
how difficult is it to set a flag in the dtypes where if a data type is detected as audio, lightwood throws and exception saying we cannot model without torchvision + pillow? This can be done if `eval(encoder_dict["module"]) is None` or by setting an encoder to a `None` type.

I wanted to make a unit-test around this anyway where any incorrect "evals" of a module that doesn't work actually throws an error; it should be in the `lookup_encoder` function. I need to think more carefully to design a better-than-hacky solution though.

George3d6:
I think we can set a rule to throw an exception if the encoder for that data type doesn't exist, which is more generic. I will do it in here.

George3d6:
And I can add it to the audio unittest

hakunanatasha (approved):

George3d6:
Ok, made the change, unittesting is a bit cumbersome since it turns out I'd have to add a whole new test for it which would be annoying to say the least as I'd involve downloading a dataset with audio files and running lightwood on that.

Did manually test locally and it produces the expected error: ```Exception: Encoder: MFCCEncoder for data type audio does not exist! You might need to install some optional dependencies (e.g. requirements_audio.txt)```

hakunanatasha:
can you not make an example unit-test that simply calls something that would default to `eval(<something>) == None` ? You can customize the json to a data-type forcing it to be audio...?

George3d6:
> can you not make an example unit-test that simply calls something that would default to eval(<something>) == None ? You can customize the json to a data-type forcing it to be audio...?

That is already happening as part of the audio unit test, we check that the import is either `None` or that it runs.

The fact that the `eval` will return the object associated to the string it evaluates (be it a class definition or a `None`) is a lanugae guarantee.

I see the PR is approved so I will go ahead and merge :)

PR #785, opened 2021-12-09
Fixes

George3d6:
- binary encoder accepts either 1 or 2 classes now
- identifier detection correctly takes `binary` type into account in an if where it was only accounting for categorical (realistically this should make no difference to current datasets, but good to fix)
- returning probabilities for OH and Binary now account for all zero vectors (though this likely makes little sense to ever happen and we should find out why)


George3d6:
Issue here: https://github.com/mindsdb/lightwood/issues/786

PR #788, opened 2021-12-16
Better T+N bounds

paxcema:
## Why 
To improve uncertainty estimation in long-horizon forecasts.

## How
Modified formula for bound width throughout the forecast from static to a constant increase given by the confidence level times the log of the time step (with a corrective factor). This is a conservative approach should increase the coverage rate (i.e. the estimated confidence value could be pessimistically biased).

### Notes
Additionally:
- Fixed bugs in confidence normalizer initialization and calling 
- Added missing bound clipping in forecasts when positive domain flag is active


lightwood/analysis/nc/calibrate.py
```diff
@@ -363,19 +371,34 @@ def explain(self, row_insights: pd.DataFrame, global_insights: Dict[str, object]
             if ns.tss.is_timeseries and ns.tss.nr_predictions > 1 and is_numerical:
                 row_insights = add_tn_conf_bounds(row_insights, ns.tss)
 
+            # clip bounds if necessary
+            if is_numerical:
+                lower_limit = 0.0 if ns.positive_domain else -pow(2, 62)
+                upper_limit = pow(2, 62)
```
hakunanatasha, line 377: phew, 2**62 a safe upper bound?

lightwood/analysis/nc/calibrate.py
```diff
@@ -363,19 +371,34 @@ def explain(self, row_insights: pd.DataFrame, global_insights: Dict[str, object]
             if ns.tss.is_timeseries and ns.tss.nr_predictions > 1 and is_numerical:
                 row_insights = add_tn_conf_bounds(row_insights, ns.tss)
 
+            # clip bounds if necessary
+            if is_numerical:
+                lower_limit = 0.0 if ns.positive_domain else -pow(2, 62)
```
hakunanatasha, line 376: 0 makes sense if positive but -2**64 is pretty intense

lightwood/analysis/nc/calibrate.py
hakunanatasha: question -is there a reason we have clipping?

lightwood/analysis/nc/norm.py
```diff
@@ -37,15 +36,15 @@ def __init__(self, fit_params: dict):
...
+        self.model = Ridge()  # TODO: enable underlying model selection from JsonAI
         self.prepared = False
         self.prediction_cache = None
         self.bounds = (0.5, 1.5)
         self.error_fn = mean_absolute_error
 
     def fit(self, data: EncodedDs) -> None:
         try:
-            data = ConcatedEncodedDs(data)
+            data = ConcatedEncodedDs([data])
```
hakunanatasha, line 47: is there a change to the data type such that it now needs to pass as a list?

lightwood/helpers/ts.py
```diff
@@ -29,18 +30,20 @@ def get_inferred_timestamps(df: pd.DataFrame, col: str, deltas: dict, tss: Times
...
-    @TODO: active research question: how to guarantee 1-e coverage for t+n, n>1
-    for now, we replicate the width and conf obtained for t+1
+    TODO: active research question: how to guarantee 1-e coverage for t+n, n>1
+    For now, conservative approach is to increase width by the confidence times the log of the time step.
     """
     for col in ['confidence', 'lower', 'upper']:
         data[col] = data[col].astype(object)
 
     for idx, row in data.iterrows():
+        error_increase = [row['confidence']] + [row['confidence'] * np.log(np.e + t / 2)
```
hakunanatasha, line 40: what is np.e? can you document why this function is as such?

hakunanatasha (approved):
I don't have enough expertise to critique this review; the standards seem fine. I had a few ignorant comments but overall should be a prioritized PR due to internal impact.

lightwood/analysis/nc/calibrate.py
paxcema: Note: this bound was not introduced in this PR. Having said that, it's in place to avoid faulty overflows if a mixer is faulty.

lightwood/analysis/nc/calibrate.py
paxcema: Yep, though as per above, it's just to avoid overflows.

lightwood/analysis/nc/norm.py
paxcema: There was a change at some point but this method was not updated. I'll add a test to explicitly try the normalizer so we're aware next time.

lightwood/helpers/ts.py
paxcema: The numerical constant `e`. Will improve the reasoning for this in the docstring.

lightwood/analysis/nc/norm.py
paxcema: For the record, this is an example where already existing unit tests did not catch something that an integration test would have caught :eyes:

PR #795, opened 2021-12-22
Temperature scaling block

paxcema:
**NOTE:** Builds upon #800.

## Why
To offer an alternative confidence estimation mechanism to the default one based on inductive conformal prediction. Temperature scaling is a simple and cheap technique that can be applied over a classifier's logits to calibrate its confidence on any given prediction.

## How
Introduce a `TempScaler` analysis block that learns the optimal temperature value for the logit softmax operation, which is then applied at prediction time. The PR includes an integration test that uses this block.


lightwood/analysis/nn_conf/temp_scale.py
```diff
@@ -0,0 +1,100 @@
...
+            output['confidence'] = torch.max(self.softmax(logits), dim=1).values.detach().numpy()
+            info['result_df'] = output
+        self.active = True
+        return info
+
+    def explain(self,
+                row_insights: pd.DataFrame,
+                global_insights: Dict[str, object], **kwargs) -> Tuple[pd.DataFrame, Dict[str, object]]:
+        """ Perform temperature scaling on logits """
+        conf_cols = [col for col in row_insights.columns
```
George3d6, line 89: Shouldn't this be a single column? Also, isn't this the "probability column" rather than the "confidence column" ?
It seems much easier to search `__mdb_proba_{pred_col_name}` (or however we call that column) and drop the whole `.numpy().reshape(...` bit fo the code.

lightwood/analysis/nn_conf/temp_scale.py
```diff
@@ -0,0 +1,100 @@
...
+
+            # collect logits and labels for the validation set
+            logits_list = []
+            labels_list = []
+            with torch.no_grad():
+                prob_cols = [col for col in ns.normal_predictions.columns
+                             if '__mdb_proba' in col and
+                             '__mdb_unknown_cat' not in col]
+                if not prob_cols:
+                    return info  # early stop if no proba info is available
```
George3d6, line 52: I think we ought to instead *not* automatically add this module via json-ai when `predict_proba` is not set for at least 1 mixer.

If the user adds it manually that's fine, it's on them to have understanding that this is needed, more confusing if they add it and it doesn't run.

lightwood/analysis/nn_conf/temp_scale.py
```diff
@@ -0,0 +1,100 @@
...
+import torch
+import pandas as pd
+from torch import nn, optim
+from sklearn.preprocessing import OrdinalEncoder
+
+from lightwood.helpers.log import log
+from lightwood.analysis.base import BaseAnalysisBlock
+
+
+class TempScaler(BaseAnalysisBlock):
```
George3d6, line 14: I honestly spent like 20 mins pouring over this and I don't quite understand why it should work better than just taking the probability and calling it a confidence... but I'm tired, if you think it's worth it feel free to merge :) but see 2 comments below first.

George3d6 (changes requested):

lightwood/analysis/nn_conf/temp_scale.py
paxcema: Here's a good reference (sections 4.2 at the end, then results in table 1): https://arxiv.org/pdf/1706.04599.pdf

TL;DR: the raw output of a softmax in a classifier is rarely well-calibrated. Turns out learning a value for the operation's "temperature" parameter can help quite a bit in terms of calibration.

lightwood/analysis/nn_conf/temp_scale.py
paxcema: I agree, and in fact that's what happens right now, as the default blocks in `JsonAI` remain unchanged.

lightwood/analysis/nn_conf/temp_scale.py
paxcema: You're right they are "probability columns", I'll rename the variable. However, it is more than one column because we yield the probability for each class as a separate column.

PR #800, opened 2021-12-28
ConfStats block

paxcema:
## Why
To offer insights with regards to each model's calibration (or lack thereof).

## How
Additional analysis block (`ConfStats`) is introduced. Gets added to any JSON AI by default but can be overridden with `use_default_analysis`.

Metrics that are currently computed:

- Binned calibration error (default bins: 10)
- Expected calibration error
- Maximum calibration error
- Global calibration score (used previously in our benchmarking suite)

Caveat: accuracy for regression & forecasting is taken as whether predictions fall within bounds.


lightwood/analysis/helpers/acc_stats.py
```diff
@@ -23,6 +24,7 @@ def analyze(self, info: Dict[str, object], **kwargs) -> Dict[str, object]:
         info['score_dict'] = evaluate_accuracy(ns.data, ns.normal_predictions['prediction'],
                                                ns.target, ns.accuracy_functions, ts_analysis={'tss': ns.tss})
         info['normal_accuracy'] = np.mean(list(info['score_dict'].values()))
+        log.info(f'Validation accuracy: {info["normal_accuracy"]}')
```
George3d6, line 27: I'd rather not log this, as stated before, I think users will just find an accuracy number confusing, doubly so a mean of all the scores, if you want to log something log the full dict with a format like this:
```
slog = '\n'.join([f'{k} -> round({v}, 3)' for k, v in info['score_dict'].items()])
log.info('Accuracy scores on validation (held out) data:\n {slog}')
```
... or some such, that's just my proposal

George3d6 (approved):

PR #805, opened 2022-01-12
Leaner TS encoder

paxcema:
Changes:

* Fix a small bug in time-delta computation.
* Replace RNN/Transformer time series encoder with a simple moving-average feature extraction, built on top of the generic array encoder. In a future PR, the RNN/Transformer encoder will be turned into mixer(s).

* ~Faster LightGBM array mixer for time series. Now, the mixer learns to predict next-step forecasts and longer horizons are reconstructed by predicting iteratively over the specified length, as opposed to the previous approach where we train a different LightGBM model for each timestep in the forecast horizon. Preliminary tests and benchmarks suggest both faster and better predictions. EDIT: This is essentially a rehash of #487~ EDIT 2: reverting due to accuracy concerns, will revisit at a later point.


lightwood/api/json_ai.py
```diff
@@ -151,8 +150,9 @@ def lookup_encoder(
                 encoder_dict["args"]["grouped_by"] = f"{gby}"
                 encoder_dict["args"]["timesteps"] = f"{tss.horizon}"
                 encoder_dict["module"] = "TsArrayNumericEncoder"
-        if "__mdb_ts_previous" in col_name:
-            encoder_dict["module"] = "ArrayEncoder"
+
+        if "__mdb_ts_previous" in col_name or col_name in tss.historical_columns:
+            encoder_dict["module"] = "TimeSeriesEncoder"
```
George3d6, line 155: Why use the new timeseries encoder for historical columns and target-history columns?

lightwood/api/json_ai.py
```diff
@@ -134,11 +134,10 @@ def lookup_encoder(
     # Time-series representations require more advanced flags
     if tss.is_timeseries:
         gby = tss.group_by if tss.group_by is not None else []
-        if col_name in tss.order_by + tss.historical_columns:
-            encoder_dict["module"] = "TimeSeriesEncoder"
-            encoder_dict["args"]["original_type"] = f'"{col_dtype}"'
-            encoder_dict["args"]["target"] = "self.target"
-            encoder_dict["args"]["grouped_by"] = f"{gby}"
+        if col_name in tss.order_by:
+            encoder_dict["module"] = "ArrayEncoder"
```
George3d6, line 138: Why not use the ArrayEncoder based timeseries encoder here ?

lightwood/encoder/time_series/ts.py
```diff
@@ -0,0 +1,62 @@
...
+        self.max_mavg_offset = self.output_size
+        self.output_size += self.max_mavg_offset
+
+    def prepare(self, train_priming_data: Iterable[Iterable], dev_priming_data: Iterable[Iterable], **kwargs):
+        """
+        Prepare the time series encoder for sequence data.
+
+        :param priming_data: time series data to train the encoder.
+        """
+        super().prepare(train_priming_data, dev_priming_data)
```
George3d6, line 30: Unless the signature is different I think you can skip this override.

George3d6 (changes requested):
Overall looks good sans the confusion about the default encoders in json.ai. Did you benchmark to make sure no accuracy drops.

Also, the rnn encoder is left in but never used by default, right?

paxcema:
> Did you benchmark to make sure no accuracy drops.

Yes, although not the most recent commit. Will do once again after addressing the review to be sure.

> Also, the rnn encoder is left in but never used by default, right?

Correct. I want to try adapting it into a mixer (especially due to the transformer backend option) and see how it performs on non-TS tasks.

lightwood/api/json_ai.py
paxcema: Because `order_by` columns are timestamps that we assume have (mostly) regularly sampled data points. As such, in theory we won't gain much by computing moving averages over it.

lightwood/api/json_ai.py
paxcema: Both target history and "historical columns" are time series info. that we can featurize with the new TS encoder (by normalizing and then getting a few moving averages).

George3d6 (approved):

paxcema:
After a few in-depth tests, am scrapping the new LGBM mixer until I can get it to work at least as well as the current approach accross the board. It does improve in some scenarios, but is also worse on others.

PR #819, opened 2022-01-25
Pulsar tutorial

chandrevdw31:


paxcema:
This looks great, thanks! We have reviewed the content and will provide a doc for you to approve the suggestions we've made. Once that is done, please include said suggestions in this PR and we can work on merging into the official docs.

George3d6:
Hi, the dataset here is mismatched from the dataset description (target seems to be set as `Development Index`) + it contains a lot of introductory material we have somewhere else in the documentation and doesn't follow the template for our tutorials (code as part of an 'ibpny notebook).

Seems like a great introduction to lightwood, I'd suggest making it a blog article on Medium or some such. But given that the docs of lightwood are a bit more minimalist and technical it's not the best fit for them right now.

PR #824, opened 2022-01-28
Fix 814

paxcema:


paxcema:
Closing for now. The burden to fetch latest context and pass it to the ML backend should probably be on MindsDB proper.

PR #825, opened 2022-01-28
Remove `anomaly_error_rate` arg in favor of `fixed_confidence`

### LINKED ISSUE #813: Merge `fixed_confidence` with `anomaly_error_rate` argument
Both are used basically for the same purpose, they should be merged.

paxcema:
Fixes #813

### DIFF
--- a/tests/integration/advanced/test_timeseries.py
+++ b/tests/integration/advanced/test_timeseries.py
@@ -105,9 +105,8 @@ class TestTimeseries(unittest.TestCase):
 
         # Check custom ICP params
         test.pop('__mdb_make_predictions')
-        preds = pred.predict(test, {'fixed_confidence': 0.01, 'anomaly_cooldown': 100, 'anomaly_error_rate': 1})
+        preds = pred.predict(test, {'fixed_confidence': 0.01, 'anomaly_cooldown': 100})
         assert all([all([v == 0.01 for v in f]) for f in preds['confidence'].values])
-        assert pred.pred_args.anomaly_error_rate == 1
         assert pred.pred_args.anomaly_cooldown == 100
 
     def test_1_time_series_regression(self):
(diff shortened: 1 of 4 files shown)


tests/integration/advanced/test_timeseries.py
```diff
@@ -105,9 +105,8 @@ def test_0_time_series_grouped_regression(self):
 
         # Check custom ICP params
         test.pop('__mdb_make_predictions')
-        preds = pred.predict(test, {'fixed_confidence': 0.01, 'anomaly_cooldown': 100, 'anomaly_error_rate': 1})
+        preds = pred.predict(test, {'fixed_confidence': 0.01, 'anomaly_cooldown': 100})
```
George3d6, line 108: I am /slightly/ confused by what is happening here... was `anomaly_error_rate` essentially the same as `error_rate`, and this PR removed it *then* renamed `error_rate` to `fixed_conf` ?

A short description would help.

Overall this is an API change and I see no issues with it internally, so if you know this is the API we need, go ahead and merge, though as always I would encourage to change the API as infrequently as possible.

George3d6 (approved):

tests/integration/advanced/test_timeseries.py
paxcema: `error_rate` was never a thing.

At some point, `fixed_conf` was used only when fitting the confidence model, and `anomaly_error_rate` at prediction. The backend has changed a bit since then. Now, in the former step there is no need for a `fixed_conf` param, and users might want to set the confidence level independently of whether the problem has anomalies or not (e.g. normal numerical tasks). 

Long story short, there was a duplicate of sorts going on, and the PR fixes the issues by defaulting to a single parameter that only triggers during inference.

paxcema: `error_rate` was never a thing.

At some point, `fixed_conf` was used only when fitting the confidence model, and `anomaly_error_rate` at prediction. The backend has changed a bit since then. Now, in the former step there is no need for a `fixed_conf` param, and users might want to set the confidence level independently of whether the problem has anomalies or not (e.g. normal numerical tasks). 

Long story short, there was a duplicate of sorts going on, and the PR fixes the issues by defaulting to a single parameter that only triggers during inference.

Issue #822, opened 2022-01-26
Json AI changes


George3d6:
Based on discussions with @torrmal we are going to make some changes to Json AI in order to make it easier to read and edit from within SQL.

```
outputs = Dict[
	target_col
	,{
		data_dtype=dtype_str
		encoder=module_dict(containing an encoder)
		mixers=[module_dict(containing a mixer)]
		ensemble=module_dict(cotaninig an ensembler)
	}
]
```

becomes

```
model = module_dict(containing an ensembler)
```

and the ensembler args will now have a `submodels` key containing `[module_dict(containing a mixer)]`

To put it another way, this:

```
    "outputs": {
        "Development Index": {
            "data_dtype": "categorical",
            "encoder": {
                "module": "OneHotEncoder",
                "args": {
                    "is_target": "True",
                    "target_weights": "$statistical_analysis.target_weights"
                }
            },
            "mixers": [
                {
                    "module": "Neural",
                    "args": {
                        "fit_on_dev": true,
                        "stop_after": "$problem_definition.seconds_per_mixer",
                        "search_hyperparameters": true
                    }
                },
                {
                    "module": "LightGBM",
                    "args": {
                        "stop_after": "$problem_definition.seconds_per_mixer",
                        "fit_on_dev": true
                    }
                },
                {
                    "module": "Regression",
                    "args": {
                        "stop_after": "$problem_definition.seconds_per_mixer"
                    }
                }
            ],
            "ensemble": {
                "module": "BestOf",
                "args": {
                    "args": "$pred_args",
                    "accuracy_functions": "$accuracy_functions",
                    "ts_analysis": null
                }
            }
        }
    }
```

Becomes:

```
"model" {
	"module": "BestOf"
                "args": {
                    "args": "$pred_args",
                    "accuracy_functions": "$accuracy_functions",
                    "ts_analysis": null
                    "submodels": [
		        {
		            "module": "Neural",
		            "args": {
		                "fit_on_dev": true,
		                "stop_after": "$problem_definition.seconds_per_mixer",
		                "search_hyperparameters": true
		            }
		        },
		        {
		            "module": "LightGBM",
		            "args": {
		                "stop_after": "$problem_definition.seconds_per_mixer",
		                "fit_on_dev": true
		            }
		        },
		        {
		            "module": "Regression",
		            "args": {
		                "stop_after": "$problem_definition.seconds_per_mixer"
		            }
		        }
		    ]
                }
}
```

[...]

PR #826, opened 2022-01-31
Simpler & better Json AI

George3d6:
This PR implements the changes discussed in #822


lightwood/helpers/templating.py
```diff
@@ -68,6 +67,10 @@ def is_allowed(v):
 
 
 def call(entity: dict) -> str:
+    # Special behavior for ensemble
+    if 'submodels' in entity['args']:
```
paxcema, line 71: Why is this needed?

From what I can gather, it removes all submodels from the JsonAI representation once built (well, `model.args.mixers = $mixers`). Is it intended that we remove mixers from the IR once predictor code is generated? I'd argue not.

lightwood/api/types.py
```diff
@@ -468,8 +376,10 @@ def from_dict(obj: Dict):
         """
         Creates a JSON-AI object from dictionary specifications of the JSON-config.
         """
-        features = {k: Feature.from_dict(v) for k, v in obj["features"].items()}
-        outputs = {k: Output.from_dict(v) for k, v in obj["outputs"].items()}
+        encoders = obj["encoders"]
+        dtype_dict = obj["dtype_dict"]
+        dependency_dict = obj.get('dependency_dict', None)
```
paxcema, line 381: default value `{}` instead?

lightwood/api/json_ai.py
```diff
@@ -664,50 +652,42 @@ def code_from_json_ai(json_ai: JsonAI) -> str:
...
                 None,
             )
         )
-        dependency_dict[col_name] = []
-        data_dtype = list(json_ai.outputs.values())[0].data_dtype
-        dtype_dict[col_name] = f"""'{data_dtype}'"""
-        json_ai.features[col_name] = Feature(encoder=encoder_dict[col_name], data_dtype=data_dtype)
+
+        dtype_dict[col_name] = f"""'{target_type}'"""
+        # @TODO: Is populating the json_ai at this stage even necessary?
```
paxcema, line 682: For time series it is, as we automatically generate encoders for e.g. previous target columns.

lightwood/api/types.py
```diff
@@ -436,8 +340,10 @@ class JsonAI:
     recipe of how to pre-process data, construct features, and train on the target column. To do so, the following \
     specifications are required internally.
 
-    :param features: The corresponding``Feature`` object for each of the column names of the dataset
-    :param outputs: The column name of the target and its ``Output`` object
+    :param encoders: The encoders corresponding to each column
+    :param dtype_dict: The dictionary of types associated to each column
+    :param dependency_dict: The dictionary of dependencies associated to each column
```
paxcema, line 345: typo: "to each encoder"

lightwood/api/types.py
```diff
@@ -436,8 +340,10 @@ class JsonAI:
...
+    :param model: The ensemble and it's submodels
```
paxcema, line 346: typo: "its"

lightwood/api/types.py
paxcema, line 345: typo: "to each encoder"

lightwood/api/json_ai.py
```diff
@@ -890,7 +870,7 @@ def code_from_json_ai(json_ai: JsonAI) -> str:
 # Fit Models
 # --------------- #
 # Assign list of mixers
-self.mixers = [{', '.join([call(x) for x in list(json_ai.outputs.values())[0].mixers])}]
+self.mixers = [{', '.join([call(x) for x in json_ai.model["args"]["submodels"]])}]
```
paxcema, line 873: This jumps to me a bit in terms of consistency. What is the difference between `mixers` and `submodels`? I guess from this PR we call `submodel` an "unrealized" `mixer`.

It may be better to opt for either `mixer` or `submodel` and use it all the way through to the autogenerated code? I'm not too convinced either way, but I can see how this could generate confusion for users.

lightwood/api/json_ai.py
paxcema, line 682: For time series it is, as we automatically generate encoders for e.g. previous target columns.

lightwood/api/types.py
George3d6: actually, I think the rule is "if it's not optional, no default value", in most cases, so I will stick to that

lightwood/api/json_ai.py
George3d6: I agree, I was too lazy to do the change but I shouldn't have been... implementing now (with all incoming bugs :p)... sigh

George3d6: Ok... looking at this, no no no, 278 occurrences, loads of bugs, I will make an issue to rename and get to it later.

George3d6: https://github.com/mindsdb/lightwood/issues/829

paxcema (approved):

PR #827, opened 2022-01-31
Dask for time series preprocessing

paxcema:
Pending actual stress tests.

### DIFF
--- a/lightwood/data/timeseries_transform.py
+++ b/lightwood/data/timeseries_transform.py
@@ -36,7 +36,6 @@ def transform_timeseries(
     
     :return: A dataframe with all the transformations applied.
     """  # noqa
-
     tss = timeseries_settings
     original_df = copy.deepcopy(data)
     gb_arr = tss.group_by if tss.group_by is not None else []
@@ -129,11 +128,10 @@ def transform_timeseries(
                 last_index += 1
 
     if len(original_df) > mp_row_cutoff:
-        # @TODO: restore possibility to override this cutoff with args
         nr_procs = get_nr_procs(original_df)
         log.info(f'Using {nr_procs} processes to reshape.')
 
-        if use_dask:
+        if use_dask and len(original_df) > 50000:  # @TODO either add another arg or rm this
             log.info("Dask active.")
             df = dd.from_pandas(pd.concat(df_arr), npartitions=nr_procs)
             df_1 = dask.delayed(_ts_to_obj)(df, historical_columns=ob_arr + tss.historical_columns)
--- a/lightwood/data/timeseries_transform.py
+++ b/lightwood/data/timeseries_transform.py
@@ -14,7 +14,8 @@ from lightwood.api import dtype
 
 def transform_timeseries(
         data: pd.DataFrame, dtype_dict: Dict[str, str],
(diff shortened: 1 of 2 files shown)


lightwood/data/timeseries_transform.py
```diff
@@ -14,7 +14,8 @@
 
 def transform_timeseries(
         data: pd.DataFrame, dtype_dict: Dict[str, str],
-        timeseries_settings: TimeseriesSettings, target: str, mode: str) -> pd.DataFrame:
+        timeseries_settings: TimeseriesSettings, target: str, mode: str,
+        use_dask: bool = True, mp_row_cutoff: int = 500) -> pd.DataFrame:
```
George3d6, line 18: Much better to default to `False` and require it to be `True` + dask installed. Otherwise, there's a risk in terms of version incompatibility + it's usually preferred to have consistent behavior regardless of user environment with a few critical exceptions (e.g. using GPU when we detect cuda)

lightwood/data/timeseries_transform.py
```diff
@@ -30,16 +31,25 @@ def transform_timeseries(
     :param timeseries_settings: A `TimeseriesSettings` object.
     :param target: The name of the target column to forecast.
     :param mode: Either "train" or "predict", depending on what phase is calling this procedure.
+    :param use_dask: If `True` and Dask is installed, lightwood will use Dask to compute the transformations.
+    :param mp_row_cutoff: if dataset size is bigger than this amount, parallel compute procedures will be used.
```
George3d6, line 35: Probably not worth having `mp_row_cutoff` as a parameter, seems to specific. Tbh I'd consider just canning it and doing mp no matter the number of rows to reduce complexity instead of reverting back to hardcoding 500.

lightwood/data/timeseries_transform.py
```diff
@@ -117,30 +127,47 @@ def transform_timeseries(
...
+            df_2 = dask.delayed(_ts_order_col_to_cell_lists)(df_1, order_cols=ob_arr + tss.historical_columns)
+            df_3 = dask.delayed(_ts_add_previous_rows)(df_2, order_cols=ob_arr + tss.historical_columns, window=window)
+            df_4 = dask.delayed(_ts_add_future_target)(df_3,
+                                                       target=target, horizon=tss.horizon,
+                                                       data_dtype=tss.target_type, mode=mode)
+            if tss.use_previous_target:
+                df_5 = dask.delayed(_ts_add_previous_target)(df_4, target=target, window=tss.window)
+                combined_df = dask.compute(df_5)[0]
+            else:
+                combined_df = dask.compute(df_4)[0]
```
George3d6, line 147: `combined_df` (probably) needs to be cast back to `pd.DataFrame`

requirements.txt
```diff
@@ -23,3 +23,4 @@ pmdarima >=1.8.0,<=1.8.3
 black >=21.9b0
 typing_extensions
 colorlog ==6.5.0
+dask==2022.1.0
```
George3d6, line 26: This will lead to installation issues, make `requirements_dask.txt` instead and add it there. Or maybe a broad `requirements_parallelsim.txt` where we can have things like dask, ray, modin, etc

George3d6 (changes requested):

paxcema:
Closing for now.

PR #845, opened 2022-02-21
Use Ridge Regression

George3d6:
In some cases regression coefficients get really big (like 10^7 kinda big) and you get overflow from small variations in parameters. Using Ridge regression fixed this in the one case where we detected a glaring error.


paxcema (approved):
LGTM

PR #860, opened 2022-04-13
fix: add complete instructions for setting up a dev environment

FedericoGarza:
I followed the instructions for [setting up a dev environment](https://github.com/mindsdb/lightwood#setting-up-a-dev-environment) and found that the installation of `requirements_image.txt` was missing (otherwise, the line `python -m unittest discover tests` fails).

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ However, we recommend creating a python virtual environment.
 #### Setting up a dev environment
 
 - Clone lightwood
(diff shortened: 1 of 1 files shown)


paxcema (approved):

paxcema:
Hi @FedericoGarza! Nice catch, you're right. Let's see if the CI passes, but this looks good to me ✅.

PR #861, opened 2022-04-13
[ENH]Include StatsForecast AutoARIMA

FedericoGarza:
The newest version of `sktime` includes a faster version of AutoARIMA (from [StatsForecast](https://github.com/Nixtla/statsforecast)). This PR changes the previous version of AutoARIMA (pmdarima) to the faster version.


requirements.txt
```diff
@@ -17,7 +17,8 @@ langdetect >=1.0.0,<=1.0.9
 dataclasses_json >=0.5.4
 autopep8 >=1.5.7
 dill ==0.3.4
-sktime >=0.5.0,<=0.8.0
+sktime >=0.11.0
```
paxcema, line 20: Can you upper cap these to avoid (potential) breaking changes without us necessarily releasing a new version? It has happened in the past :) 

I think this

```python
sktime >=0.11.0,<0.12.0
statsforecast >=0.5.3,<0.6.0
```

should work.

requirements.txt
```diff
@@ -17,7 +17,8 @@ langdetect >=1.0.0,<=1.0.9
...
+statsforecast >=0.5.3
 torch_optimizer ==0.1.0
 pmdarima >=1.8.0,<=1.8.3
```
paxcema, line 23: Also, we should remove this dependency.

paxcema (changes requested):
Looking good, just a couple of small changes in the requirements file.

requirements.txt
FedericoGarza: Sure, fixed in [be7b9ae](https://github.com/mindsdb/lightwood/pull/861/commits/be7b9ae003f4690d8b3b7572f2d488550c05cffd). :D

requirements.txt
FedericoGarza: Right, fixed in [9787398](https://github.com/mindsdb/lightwood/pull/861/commits/9787398b5aaace26b50c12f5dbbc3e8b7fc914f2). :)

paxcema (approved):

Issue #746, opened 2021-11-10
Make `EncodedDs` interface simpler


George3d6:
The methods are a bit confusing, especially the `get_{x}` methods, try to take a look at this and see if we can simplify it.

paxcema:
Good progress made in #1150. There's still more we can do, but at least now it's much more efficient and can handle larger datasets.

paxcema:
Closing for now.

PR #907, opened 2022-06-20
STL decomposition

paxcema:
# Why

To improve the set of tools available for time series mixers to use if it is deemed useful.

# How

This PR does a _bunch_ of stuff, namely:

* Big refactor in the TS transform step. We now default to using pandas date/time utilities, meaning that indexes are now proper timestamps (multi-index in the case of grouped tasks). This both simplifies the logic and makes it more robust for things like inferring new rows according to the observed series frequencies.
* Introduces the `NeuralTs` mixer, which inherits from the vanilla `Neural` but adds time-series specific logic. This was done to keep specialized procedures separated from the (simpler) classification and regression tasks. JsonAI dispatch has been modified accordingly to make use of this.
* Available time series groups at training time are now logged and registered inside the `StatisticalAnalysis` object.
* Some simplifications to `ConcatedEncodedDs`, though we should probably remove this abstraction entirely (see #746).
* Moves a bunch of TS utilities to `lightwood.helpers.ts`
* `ts_analysis` phase now finds (and fits) optimal deseasonalizer and detrender objects for each series available at train time, for which we use Optuna.
* The respective methods that mixers can use to leverage these STL blocks are in `mixers.helpers.ts`
* `LightGBMArray`, `SkTime`, `Prophet` have been updated to use STL blocks, controllable by JsonAI args if needed. (NOTE: NeuralTs calls the blocks but there's no effect because it uses encoded values inside the DS, this will be addressed in a new PR)
* Encoder `TsNumericEncoder` has been simplified to avoid storing the sign, instead simply applying the underlying normalizer. This seems to be especially useful for `NeuralTs` to achieve improved forecasts, but formal benchmarks are still pending.
* Modifications to the `analysis.nc` block to work with all the above.


lightwood/data/statistical_analysis.py
```diff
@@ -171,6 +172,18 @@ def statistical_analysis(data: pd.DataFrame,
         else:
             avg_words_per_sentence[col] = None
 
+    if problem_definition.timeseries_settings.is_timeseries:
+        groups = get_ts_groups(data, problem_definition.timeseries_settings)
+        # deltas, periods, freqs = get_delta(data, dtypes, groups, problem_definition.timeseries_settings)
+        ts_stats = {
+            'groups': groups,
+            # 'deltas': deltas,
```
paxcema, line 180: restore

lightwood/data/timeseries_analyzer.py
```diff
@@ -1,22 +1,30 @@
...
+from sktime.transformations.series.detrend import ConditionalDeseasonalizer
 
 from lightwood.api.types import TimeseriesSettings
 from lightwood.api.dtype import dtype
+from lightwood.helpers.ts import get_ts_groups, get_delta
+from lightwood.helpers.log import log
 from lightwood.encoder.time_series.helpers.common import generate_target_group_normalizers
 from lightwood.helpers.ts import Differencer
-from lightwood.helpers.general import get_group_matches
+from lightwood.helpers.ts import get_group_matches
```
paxcema, line 17: move to the above import

lightwood/data/timeseries_analyzer.py
```diff
@@ -30,67 +38,36 @@ def timeseries_analyzer(data: Dict[str, pd.DataFrame], dtype_dict: Dict[str, str
     :return: Dictionary with the aforementioned insights and the `TimeseriesSettings` object for future references.
     """  # noqa
     tss = timeseries_settings
+    groups = get_ts_groups(data['train'], tss)
+    deltas, periods, freqs = get_delta(data['train'], dtype_dict, groups, tss)
```
paxcema, line 42: restore analysis as input and avoid duplicate computation of these

lightwood/mixer/lightgbm_array.py
```diff
@@ -49,50 +50,52 @@ def __init__(
...
-            if timestep > 0:
-                train_data.data_frame[self.target] = train_data.data_frame[f'{self.target}_timestep_{timestep}']
-                dev_data.data_frame[self.target] = dev_data.data_frame[f'{self.target}_timestep_{timestep}']
-
-            self.models[timestep].partial_fit(train_data, dev_data)  # @TODO: this call could be parallelized
-
-        # restore target
-        train_data.data_frame[self.target] = original_target_train
-        dev_data.data_frame[self.target] = original_target_dev
+        pass
```
paxcema, line 75: restore

tests/integration/advanced/test_timeseries.py
```diff
@@ -20,6 +20,7 @@
 np.random.seed(0)
 
 
+# TODO: add tests for STL blocks usage
```
paxcema, line 23: add a test

PR #911, opened 2022-06-28
Use `prophet` 1.1 with wheels

abitrolly:
https://github.com/facebook/prophet/issues/2181#issuecomment-1166057319

### DIFF
--- a/requirements_prophet.txt
+++ b/requirements_prophet.txt
@@ -1,2 +1,2 @@
 pystan==2.19.1.1
-prophet
+prophet==1.1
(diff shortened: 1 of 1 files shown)


requirements_prophet.txt
```diff
@@ -1,2 +1,2 @@
 pystan==2.19.1.1
-fbprophet
\ No newline at end of file
+prophet
```
paxcema, line 2: This looks promising! @abitrolly can you force the version to be `==1.1`?

paxcema (changes requested):

requirements_prophet.txt
abitrolly: Done. This should speed up Docker builds by some 20+ minutes. However, I am not sure if `prophet` is still in MindsDB defaults.

PR #930, opened 2022-07-09
Change supported Python version

ZoranPandovski:
Deprecate the 3.6 version of Python


paxcema:
I think we should be good to go on this one but would like confirmation from @StpMax first.

PR #937, opened 2022-07-15
[ENH] Force repeat datetime encoding

paxcema:


paxcema:
Update: this was not merged, but most of the underlying changes made their way via #938. This one had poor benchmark results so it has been dropped, at least for the time being.

PR #948, opened 2022-07-22
[ENH] Message for missing values

paxcema:
# Why

`StatisticalAnalysis` can be used within MindsDB for data visualization purposes. Short, descriptive messages should accompany all relevant insights, `missing` among them. We currently don't do that, which is what the PR fixes. 

# How

Added a short descriptive message.


Ricram2 (approved):

Ricram2:
Like a rocket.

PR #951, opened 2022-07-25
[ENH] Use Ray.Pool

paxcema:


paxcema:
Closing for now, managed to make M1 work without this. May reopen at a later date.

PR #983, opened 2022-08-29
[ENH] PyOD analysis block

### LINKED ISSUE #955: PyOD as analysis block
[Reference](https://github.com/yzhao062/pyod)

paxcema:
# Why

Fixes #955. Improves the set of tools available at analysis time.

# How

Introduces a wrapper for the PyOD library. It's simple for now, supports only a single "model" variant (SUOD chain with a bunch of underlying methods). We may increase flexibility for this in the future. Currently, the only customizable parameter is the amount of "contamination", or proportion of outlier data that we want to tag in the training dataset.

## Notes
- This block is not enabled by default, as execution time benchmarks should be run prior to making this decision.
- All regression, classification and time series tasks should be supported by this (with a unit test for each).
- The data ingested by the method in time series tasks has no inter-row dependency, which means the anomaly detection will probably be poor. We need to improve upon this, maybe with the `auto_encoder_torch` module or our own `lightwood.encoder.time_series.TimeSeriesEncoder` class.

### Example

```sql
CREATE PREDICTOR 
  mindsdb.home_rentals_model
FROM example_db
  (SELECT * FROM demo_data.home_rentals)
PREDICT rental_price
USING
analysis_blocks=[
  {"module": "PyOD", "args": {"contamination": 0.1, "pyod_window": 1234}},
];
```


paxcema:
Next steps:

- Deaggregate window on new columns @tomshuds
- Use other modules (torch AE)

Issue #884, opened 2022-05-24
Store original amount of columns when generating JsonAI instance


paxcema:
This will be useful for reference because internally columns can be dropped or added.

PR #997, opened 2022-10-03
Added variable to address issue #884

jacktag6:
Added variable to hold original column count when generating JsonAI instance

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -208,7 +208,7 @@ def generate_json_ai(
         ):
             if col_name != problem_definition.target:
                 input_cols.append(col_name)
-
+    original_col_amount = len(input_cols)
     is_target_predicting_encoder = False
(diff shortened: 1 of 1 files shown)


paxcema (changes requested):
Hi @jacktag6, thanks for your contribution!

We've discussed internally with the core team, and we think it will be better to move this inside the `StatisticalAnalyzer` class, which you can find [here](https://github.com/mindsdb/lightwood/blob/a7b02e6cfc54b7fb1844e3b69efbd3695eb6b58f/lightwood/api/types.py#L57). The actual implementation should be very similar to the one for the parameter `nr_rows`. As you can see in this [snippet](https://github.com/mindsdb/lightwood/blob/984e7481aa18446c6f2c2fd49a00a39189654920/lightwood/data/statistical_analysis.py#L206), we create it and then pass it to the class constructor when returning the object.

If you need more help with implementing this, just leave us a comment and we can provide more details.

paxcema:
Closing via #1014

PR #1004, opened 2022-10-05
Argument to override device setting

### LINKED ISSUE #996: Add the possibility to specify the device name at init for NN models
For the moment it seems that the default behaviour is to check if cuda is available and if so, use it (see for example [encoder/text/helpers/rnn_helpers.py](https://github.com/mindsdb/lightwood/tree/staging/lightwood/encoder/text/helpers/rnn_helpers.py) and [helpers/device.py::get_devices](https://github.com/mindsdb/lightwood/tree/staging/lightwood/helpers/device.py), the device that will be used is either the global variable `device` defined in the first file or is the output of `get_devices()`).

It would be good to be able to specify the device manually.

As far as I'm concerned, the issue is that some of the default models cannot be trained on GPU due to memory constraints but could potentially be trained on CPU.

alexandre-dz-oscore:
resolves #996 
- Added a key word argument to objects based on torch modules to specify a device:
  - encoder.categorical.autoencoder
  - encoder.images.img_to_vec
  - encoder.text.pretrained
  - (encoder.text.rnn_helpers, encoder.text.rnn)
  - encoder.text.short
  - encoder.time_series.rnn
  - mixer.helpers.default_net
  - mixer.helpers.residual_net
- added `get_device_from_name` utility function in helpers.device
- added tests for checking device type


paxcema (changes requested):
Looks good in general! There's a couple of minor issues that I've commented below:

lightwood/encoder/text/helpers/rnn_helpers.py
```diff
@@ -509,15 +535,16 @@ def indexesFromSentence(lang, sentence):
...
+def tensorFromSentence(lang, sentence, device=default_device):
     indexes = indexesFromSentence(lang, sentence)
     indexes.append(EOS_token)
     return torch.tensor(indexes, dtype=torch.long, device=device).view(-1, 1)
 
 
-def tensorsFromPair(pair, input_lang, output_lang):
-    input_tensor = tensorFromSentence(input_lang, pair[0])
-    target_tensor = tensorFromSentence(output_lang, pair[1])
+# NOTE: the function below is never used?
```
paxcema, line 544: Yeah, odd, must be a leftover. Can you remove it please?

lightwood/encoder/text/helpers/rnn_helpers.py
```diff
@@ -552,8 +579,10 @@ def tensorsFromPair(pair, input_lang, output_lang):
 
 def train(
         input_tensor, target_tensor, encoder, decoder, encoder_optimizer, decoder_optimizer, criterion,
-        max_length=MAX_LENGTH):
+        max_length=MAX_LENGTH, device=default_device):
+
     encoder_hidden = encoder.initHidden()
+    #device = encoder.device
```
paxcema, line 585: Remove comment

lightwood/encoder/text/pretrained.py
```diff
@@ -52,6 +53,7 @@ def __init__(
         :param epochs: number of epochs to train model with
         :param output_type: Data dtype of the target; if categorical/binary, the option to return logits is possible.
         :param embed_mode: If True, assumes the output of the encode() step is the CLS embedding (this can be trained or not). If False, returns the logits of the tuned task.
+        :param device: name of the device to use (default is an empty string), if is an empty string will use the output of `get_devices()` instead.
```
paxcema, line 56: typo: "`get_device_from_name()`* instead"

lightwood/encoder/text/rnn.py
```diff
@@ -1,14 +1,20 @@
 # flake8: noqa
 from lightwood.encoder.text.helpers.rnn_helpers import *
+#from lightwood.encoder.text.helpers.rnn_helpers import device as default_device
```
paxcema, line 3: remove comment

lightwood/helpers/device.py
```diff
@@ -36,3 +36,11 @@ def get_devices():
         available_devices = 0
 
     return torch.device(device_str), available_devices
+
+
+def get_device_from_name(device_name=''):
```
paxcema, line 41: I think you should modify all modified docstrings a bit.

Change all instances of:

> "name of the device to use (default is an empty string), if is an empty string will use the output of `get_devices()` instead") .

To:

> "name of the device that `get_device_from_name` will attempt to use.

And make the former to be the docstring in this method.

lightwood/mixer/helpers/residual_net.py
```diff
@@ -38,15 +38,22 @@ def __init__(self,
...
         super(ResidualNet, self).__init__()
         self.net = torch.nn.Sequential(
             *
             ([ResidualModule(input_size) for _ in range(1)] +
              [nn.Linear(input_size, max([input_size * 2, output_size * 2, 400])),
               nn.Linear(max([input_size * 2, output_size * 2, 400]),
                         output_size)]))
-        self.to(*get_devices())
+
+        self.to(get_device_from_name(device))
```
paxcema, line 51: Yeah, this will fail. Notice the `self.to()` definition: it takes the `device` but also the amount of `available_devices`.

I think one option here would be to turn `available_devices` in `self.to` into an `Optional[int] = 1` argument.

tests/unit_tests/encoder/categorical/test_autoencoder.py
```diff
@@ -46,3 +46,28 @@ def test_autoencoder(self):
         encoder_accuracy = accuracy_score(list(map(str, test_data)), list(map(str, decoded_data)))
         print(f'Categorical encoder accuracy for: {encoder_accuracy} on testing dataset')
         self.assertTrue(encoder_accuracy > 0.70)
+
+    def check_encoder_on_device(self, device):
+        enc = CategoricalAutoEncoder(stop_after=5, device=device)
```
paxcema, line 51: A lot of this test seems to be copy-pasted from the one above. Could you abstract the data generation procedure into a common function that gets called by both?

(Note: I am ignoring the fact that this way of generating data seems very much unoptimal, but then again it wasn't introduced in this PR and it's also not its objective, so we will tackle it in another issue).

lightwood/mixer/helpers/residual_net.py
alexandre-dz-oscore: Maybe one could have a function that merges the behaviour of `get_devices` and `get_device_from_name`, for example adding an optional `device_str=''` to the former and returns the number of available devices in the cuda case.

paxcema (approved):
Very nice, thanks for your contribution @alexandre-dz-oscore!

Issue #1010, opened 2022-10-07
🔧 Add library version to predictors


paxcema:
## What?

This issue is pretty simple: we should document what was the exact Lightwood version that was used to create every predictor object.

## How?

You can get the `lightwood.__version__` and add it to the predictor generation code in the `code_from_json_ai()` method inside `lightwood/api/json_ai.py`.

adripo:
Hi @paxcema, can you have a look at my pr? I am not sure that my solution is correct and I am not able to test it.

paxcema:
Yes, but let's give @adripo a full week before calling it (so, if he doesn't update in 4 days from now, I will reassign it to you).

adripo:
@paxcema sorry I was was busy for the past few days. I will work this issue today.

PR #1011, opened 2022-10-07
Add library version to predictors

adripo:
fix: #1010

### DIFF
--- a/lightwood/analysis/analyze.py
+++ b/lightwood/analysis/analyze.py
@@ -75,7 +75,8 @@ def model_analyzer(
         'tss': tss,
         'ts_analysis': ts_analysis,
         'accuracy_functions': accuracy_functions,
-        'has_pretrained_text_enc': has_pretrained_text_enc
+        'has_pretrained_text_enc': has_pretrained_text_enc,
+        'library_version': lightwood.__version__
     }
 
     for block in analysis_blocks:


lightwood/analysis/analyze.py
```diff
@@ -75,7 +75,8 @@ def model_analyzer(
         'tss': tss,
         'ts_analysis': ts_analysis,
         'accuracy_functions': accuracy_functions,
-        'has_pretrained_text_enc': has_pretrained_text_enc
+        'has_pretrained_text_enc': has_pretrained_text_enc,
+        'library_version': lightwood.__version__
```
paxcema, line 79: Tests are failing (you can check [here](https://github.com/mindsdb/lightwood/actions/runs/3206717079/jobs/5240808895) for more information). Try to `from lightwood import __version__` instead to avoid this particular exception.

paxcema (changes requested):
Hi @adripo, thanks for contributing!

We've discussed with the team for options, and it's probably better to move this into the code generation phase.

Simply move this line you've added to `lightwood/api/json_ai.py:1188` ([link](https://github.com/mindsdb/lightwood/blob/a7b02e6cfc54b7fb1844e3b69efbd3695eb6b58f/lightwood/api/json_ai.py#L1188)) and assign it to `self.lightwood_version`.

You can also check the implementation is correct if the automated tests pass.

paxcema:
Closing via #1018

PR #1018, opened 2022-10-13
Add lightwood library version to predictor

### LINKED ISSUE #1010: 🔧 Add library version to predictors

adripo:
fixes #1010 

As explained in #1011 the versioning logic will be added in code generation phase.

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -1186,6 +1186,7 @@ class Predictor(PredictorInterface):
         self.accuracy_functions = {json_ai.accuracy_functions}
         self.identifiers = {json_ai.identifiers}
         self.dtype_dict = {inline_dict(dtype_dict)}
+        self.lightwood_version = lightwood_version
 
         # Any feature-column dependencies
         self.dependencies = {inline_dict(json_ai.dependency_dict)}
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -11,7 +11,7 @@ from lightwood.api.types import (
 )
 import inspect
 from lightwood.helpers.log import log
-from lightwood import __version__ as lightwood_version
+from lightwood.__about__ import __version__ as lightwood_version
 
 
 # For custom modules, we create a module loader with necessary imports below
--- a/tests/integration/basic/test_regression.py
+++ b/tests/integration/basic/test_regression.py
@@ -4,6 +4,7 @@ import pandas as pd
 from sklearn.metrics import r2_score
 from lightwood.api.types import ProblemDefinition
 from lightwood.api.high_level import json_ai_from_problem, predictor_from_json_ai
+from lightwood import __version__ as lightwood_version
 
 
 class TestBasic(unittest.TestCase):
@@ -38,6 +39,8 @@ class TestBasic(unittest.TestCase):
         predictor.learn(df)
 
         assert predictor.model_analysis.dtypes[target] == dtype.quantity
+
+        assert predictor.lightwood_version == lightwood_version
 
         predictions = predictor.predict(df)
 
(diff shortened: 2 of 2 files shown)


paxcema (changes requested):
Please modify an existing integration test (e.g. `test_regression.test_0_predict_file_flow`) to assert that the version actually is being stored inside the predictor object with the correct value.

By doing that, you will notice that this implementation -as it currently stands- is not going to work. The `predictor_code` variable is a _string_ that contains code, so you need to inject the actual content (same mechanism than what is being used for e.g. `self.target`).

Additionally, `lightwood_version` has to be defined first in the imports (a simple alias to `lightwood.__version__` is enough).

adripo:
I was in doubt that my change would work, but I tried.
Ok, pretty clear the part of the integration tests.

About the `lightwood_version` I saw that it is already imported [here](https://github.com/mindsdb/lightwood/blob/staging/lightwood/api/json_ai.py#L34) and then recalled [here](https://github.com/mindsdb/lightwood/blob/staging/lightwood/api/json_ai.py#L1171) in the `predictor_code`

So I thought that the version is assigned when `code_from_json_ai` is executed. Am I missing something?

adripo:
If what I was supposing in the last message is not correct I have done the modification that you suggested in a new pr on my fork. If this is the correct solution I will merge it in the branch of this pr.
Please have a look:
https://github.com/adripo/lightwood/pull/1

paxcema:
Hey @adripo!

Yes, the change in that new branch is correctly injecting the version 👍 into the generated code, if you merge that we will be good to go (though I'm pretty sure the test will need to compare against the `str(lightwood_version)` now).

adripo:
@paxcema the changes are done as suggested. Please let me know if there's anything else that could be done.

adripo:
@paxcema after the change the tests are failing with `ImportError: cannot import name '__version__' from partially initialized module 'lightwood'`

adripo:
@paxcema there is a circular import, as we import [lightwood.api](https://github.com/mindsdb/lightwood/blob/f555d5373ded27bc433dcb1959232cce93daf5bb/lightwood/__init__.py#L4) that will import [code_from_json_ai](https://github.com/mindsdb/lightwood/blob/f555d5373ded27bc433dcb1959232cce93daf5bb/lightwood/api/__init__.py#L29) that reference [high level](https://github.com/mindsdb/lightwood/blob/f555d5373ded27bc433dcb1959232cce93daf5bb/lightwood/api/high_level.py#L12)
and lastly json_ai where we added [__version__ import](https://github.com/mindsdb/lightwood/blob/f555d5373ded27bc433dcb1959232cce93daf5bb/lightwood/api/json_ai.py#L14). 

So the version value cannot be referenced when lightwood will be initialized. I still think that my logic as explained here is correct so the value of `lightwood_version` can be referenced when `predictor_code` is recalled. Please let me know how to proceed.

paxcema:
> there is a circular import

Try importing directly from `lightwood.__about__` instead of the top level.

>  value of lightwood_version can be referenced when predictor_code is recalled

It can, but it is not correct, as that version is retrieved dynamically (i.e. once the predictor is loaded in memory and called). This can be done with any lightwood version (even ones that are not compatible with any given predictor) which is why it's important to write the actual version used as a string.

adripo:
@paxcema thank you for the detailed explanation. It worked with this last fix.

paxcema (approved):
Glad to hear! Looks good, so I will merge it. Thank you for your contribution @adripo 🚀

Issue #672, opened 2021-10-22
Add static analysis tools for PRs


George3d6:
We need to add a static analysis tool that triggers on each PR and provides a report, ideally `flake8` style where we can configure its behaviour *and* have the action fail until the PR respects all imposed rules. The "configure behaviour" bit is important since we might have some standards that are not in line with static analysis preferences (e.g. there's certain bits where we use `exec` and `eval` or where we must use `*` imports and so forth)

Multiple people can tackle this one, since the ore tools the merrier as long as they catch actual errors, ideally also fix the errors detected as part of the PR if it's easy enough. However if the tool catch more than 5-10 errors then don't bother, just make the PR, since we might think those are not actual errors and will want to configure them away.

Currently, I use pylance and find it amazing for this, but am open to any tool

George3d6:
Hacktoberfest is over :( But if you want to finish the PR (@vickywane )and/or if anyone else wants to work on this, it's still something we'd very much like

vickywane:
@George3d6 I really would like to finish my work on introducing Prospector here but the issue I faced was that Prospector was not outputting the logs in the Github CI Run. 

I dont know if it’s the way stdio is handled within your GitHub CI but locally the same tool prints out the analysis result. I will however try a few more times to see if i can figure it out.

vickywane:
I'll give this a shot

vickywane:
@paxcema I have gotten Prospector set up within this [pull request](https://github.com/mindsdb/lightwood/pull/1021).

I think we can proceed further to configure its behavior. 

Cc:  @George3d6

chandrevdw31:
@vickywane any progress on the updates in the pr?

vickywane:
Nothing yet as I didn't get feedback from the reviewers. 

I'll be glad to dedicate some time to work on this next week. 

Considering that Hacktoberfest is over, will there be any incentive for working on this? 🌚

PR #1021, opened 2022-10-16
feat: add propsector code analysis to CI run

vickywane:
This pull request will fix this [issue](https://github.com/mindsdb/lightwood/issues/672).

This pull request adds [Prospector](https://prospector.landscape.io/en/master/index.html) as a static code analysis tool to run an analysis each time a PR is opened through the use of GitHub actions.


vickywane:
At this point, Prospector has been set up with a very minimal configuration and it stores the result using GitHub Action Artifacts. We can proceed to adjust its configuration to better suit the project. 

@paxcema Please review.

paxcema (changes requested):
Hey @vickywane,

Looks like an interesting proposal. I think it's still missing a way for us to leverage this in an automated fashion when contributors open new pull requests, because I don't think this configuration goes beyond writing the report internally in the CI machines, is this correct? If so, then I think we need to figure out a way to achieve that before merging this.

paxcema:
> what exactly is being expected

Largely what the original issue states: a way for us to automatically trigger a bot in new PRs that offers static analysis (yes, will be noisy but should be fine as long as the report is actually useful to either/both maintainer and contributors).

Prospector is a good choice. I think we can configure it to check for our current flake8 rules, plus anything else that makes sense and is easy to add. After that, the report should automatically be added to the PR discussion with an easy-to-read markdown format. And I think after the initial report, ideally a contributor would comment some special string to retrigger the check, though this may be too complicated and best left for a second PR.

Does this make sense?

vickywane:
Thank you for explaining @paxcema. 

I'll make an update to the setup.

vickywane:
Thank you for explaining @paxcema. 

I'll make an update to the setup.

vickywane:
@paxcema I have made an update on this PR. 

Here are my updates: 

-  I have re-configured Prospector to generate YAML content for the static analysis. [Outputting](https://prospector.landscape.io/en/master/usage.html#output-format) the analysis in Markdown format is not supported by the prospector tool. 

- Prospector doesn't have support for Flake8. What I have done is to re-configure it to have similar behavior as the `.flake8` file in ignoring certain paths and the `max-line-length`. The flake8 rules are done with Prospector's [strictness levels](https://prospector.landscape.io/en/master/usage.html#strictness). It would be great if you could take a look at the current logs from Prospector and let me know if I should increase or reduce the current strictness

- I added the [thollander/actions-comment-pull-request@v1](https://github.com/marketplace/actions/comment-pull-request) action to paste the YAML content of the analysis here but that seems to have a permissions-related blocker.  [thollander/actions-comment-pull-request@v1](https://github.com/marketplace/actions/comment-pull-request) requires a `GITHUB_TOKEN` in order to be able to make a comment, but I don't have access to the token in as a contributor. From the research done, that token will be available after the PR has been merged into the main branch.  See [this issue](https://github.com/actions/first-interaction/issues/10) on this.

paxcema (approved):
This is looking great @vickywane!

I'd say we can go ahead, merge and check how it behaves. @ZoranPandovski what do you think?

paxcema:
@vickywane while we wait for the other reviewer, I noticed the linting check is now throwing a lot of warnings, as if the flake8 configuration had been change in the linting step, as well. Any clues as to why that may be happening? It is a blocker for merging, I'm afraid.

tests/requirements.txt
```diff
@@ -0,0 +1 @@
+prospector==1.7.7
```
paxcema, line 1: Also, I don't think we need this right? The CI script manually installs it, anyway.

paxcema:
@vickywane any updates? Also, @ZoranPandovski it would be awesome if you can review this

.github/workflows/analysis.yml
```diff
@@ -0,0 +1,35 @@
+name: Static Code Analysis Of Codebase
+on: [push, pull_request]
```
ZoranPandovski, line 2: Maybe for now we can just run this on pull_request

tests/requirements.txt
ZoranPandovski: Yes, it is installed in the gh action env

ZoranPandovski (changes requested):

ZoranPandovski:
Thanks @vickywane. Please just address the above two comments.

paxcema:
Closing this for now. Feel free to reopen.

PR #1022, opened 2022-10-16
fix wrong time-series alignment and add WA for StatsForcastAutoARIMA

bachng2017:
a proposed fix for time-series alignment error and issue of StatsForcastAutoARIMA
1) see https://github.com/mindsdb/mindsdb/issues/3234 for detail about the alignment issue
2) see https://mindsdbcommunity.slack.com/archives/C037482KJ22/p1665911392587569 for more detail about the issue of StatsForcastAutoARIMA

the proposed fix is tested with house-sale tutorial only. It might need more review to consider the impact of the change


ZoranPandovski:
recheckcla

lightwood/mixer/sktime.py
```diff
@@ -291,15 +291,30 @@ def _call_groupmodel(self,
...
         start = max(offset, min_offset)
         end = series.shape[0] + offset + self.horizon
-        all_preds = model.predict(np.arange(start, end)).tolist()
+        # all_preds = model.predict(np.arange(start, end)).tolist()
+        # StatsForcastAutoARIMA return wrong prediction with data point leq zero
+        # Here a WA proposed by getting prediction for ALL trained data and
+        # slice it again from the last data point
+        if start == 0:
+            start = min_offset
+            all_preds = model.predict(np.arange(start, end)).tolist()[-min_offset:]
```
paxcema, line 309: Yeah, this workaround will do for now. It may be slightly more compute, but we can remove it once the behavior is fixed in the dependency.

paxcema, line 309: Yeah, this workaround will do for now. It may be slightly more compute, but we can remove it once the behavior is fixed in the dependency.

paxcema (changes requested):
Thanks for the PR :)

The workaround is okay. There are two other modifications that I'm not 100% convinced on yet. I will manually test a bit more, but if you have any proof it would be awesome to translate that into a `test` that captures the failing behavior, so that we can catch this next time that the alignment fails. If you can't, then give me some time and I will 1) merge this and 2) add tests on top.

paxcema:
Update: the test in #1023 leads me to believe the general alignment behavior is correct (`StatsForecast` bug notwithstanding), and unless we have a test to showcase an edge case that justifies the above changes, I'm inclined to not merge those. I will try to find such a test, but in the meantime @bachng2017, it does seem like the issue is in `mindsdb` rather than `lightwood`, so I will check that now.

bachng2017:
thanks  @paxcema 
did you check the conversation at https://mindsdbcommunity.slack.com/archives/C037482KJ22/p1665911392587569
from that it  looks like an issue of StatsForcastAutoARIMA for us

paxcema:
For now, can you please remove the changes in offset indexes?

Just leave the StatsForecast workaround, and I will merge immediately. We can debug the second problem in a separate PR. Thanks!

bachng2017:
yes we've just left the StatsForcastAutoARIMA wa there and remove other changes

paxcema:
Can you try to merge again please? Somehow GitHub thinks the newly added `RandomForest` mixer will be added again, which should not be the case.

bachng2017:
tried to merge again, not sure is that OK or not

paxcema:
Yes, merge is now fine 👍 

There is a failing test, however. I will have to take a closer look on why before merging.

bachng2017:
thanks, the merge also very helpful for us.
We will do our own test (manual) with the latest and try to submit other issue (because we think the index issue which we've removed from this PR still be there)

PR #1024, opened 2022-10-18
Migrate to type_infer

paxcema:
# Why
Part of a series of functionality extraction out of lightwood into smaller, leaner packages so that they can be called from elsewhere without requiring huge dependencies involved in heavier parts of the pipeline (namely, encoding and mixing).

# How
Replace type inferring procedures with the newly created `mindsdb/type_infer` which, for now, is pretty much a copy-paste of the current implementation, which means this merge should not affect performance whatsoever.


tomhuds (approved):

paxcema:
Note: to merge only after #1027

PR #1026, opened 2022-10-21
Implement test method for PredictorInterface

### LINKED ISSUE #988: `Test` method for PredictorInterface
We should add a new method to the predictor interface for arbitrary metric computation given a dataframe with values for the target column.

```python
pdef = ProblemDefinition.from_dict({...})
json_ai = json_ai_from_problem(train_df, problem_definition=pdef)
predictor_class_code = code_from_json_ai(json_ai)
predictor = predictor_from_code(predictor_class_code)
predictor.learn(train_df)

# existing behavior
predictions = predictor.predict(test_df)

# new behavior

# 1) string for already supported metrics and user-defined modules
metrics = predictor.test(test_df, metrics=['mean_absolute_error', 'my_custom_module.method_name'])

# 2) support for callables and lambdas
from sklearn.metrics import mean_absolute_error
metrics = predictor.test(test_df, metrics=[mean_absolute_error, lambda x, y: abs(x - y)])
```

Returned object should be a Pandas dataframe with the value of each passed metric:

```python
In[1]: metrics
Out[1]:
                             metric_name  value
mean_absolute_error  mean_absolute_error  32433
lambda_0                        lambda_0  45576
```

It may also be worth defining an extra kwarg so that the user can retrieve predictions (as if they were using `predict()`):

```python
metrics, predictions = predictor.test(test_df, metrics=['mean_absolute_error'], return_preds=True)
```

ErdoganSeref:
The change resolves #988.

### DIFF
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -48,7 +48,7 @@ from lightwood.helpers.text import *
 from lightwood.helpers.torch import *
 from lightwood.mixer import *
 import pandas as pd
-from typing import Dict, List, Union
+from typing import Dict, List, Union, Callable
 import os
 from types import ModuleType
 import importlib.machinery
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -1182,7 +1182,7 @@ else:
 
     test_body = """
 predictions = self.predict(data)
-values = [metric(data[self.target], predictions) for metric in metrics]
+values = [metric(data[self.target], predictions[self.target]) for metric in metrics]
 names = [metric.__name__ for metric in metrics]
 return pd.DataFrame(list(zip(names, values)), columns=['Metric', 'Value'])
 """
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -1180,11 +1180,11 @@ else:
 
     predict_body = align(predict_body, 2)
 
-    test_body = f"""
+    test_body = """
 predictions = self.predict(data)
 values = [metric(data[self.target], predictions) for metric in metrics]
 names = [metric.__name__ for metric in metrics]
-return pd.DataFrame(list(zip(names, values)), columns=['Metric', 'Value']) 
+return pd.DataFrame(list(zip(names, values)), columns=['Metric', 'Value'])
 """
 
     test_body = align(test_body, 2)
--- a/lightwood/api/json_ai.py
+++ b/lightwood/api/json_ai.py
@@ -1180,6 +1180,15 @@ else:
 
     predict_body = align(predict_body, 2)
 
+    test_body = f"""
+predictions = self.predict(data)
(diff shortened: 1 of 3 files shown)


lightwood/api/json_ai.py
```diff
@@ -1265,6 +1274,10 @@ def adjust(self, train_data: Union[EncodedDs, ConcatedEncodedDs, pd.DataFrame],
     @timed
     def predict(self, data: pd.DataFrame, args: Dict = {{}}) -> pd.DataFrame:
 {predict_body}
+
+    @timed
+    def test(self, data: pd.DataFrame, metrics: List[Callable]) -> pd.DataFrame:
```
paxcema, line 1279: Add `Callable` in the imports section (see `IMPORTS` variable) near the start of the file.

paxcema (changes requested):
This is looking good @ErdoganSeref! 

Apart from the fix I've suggested above, I would ask to test this method in one of the integration tests. Doesn't have to be anything fancy – getting _any_ value for e.g. R2 score on a regression test, or balanced accuracy score on a classification test would be enough!

tests/integration/basic/test_regression.py
```diff
@@ -41,11 +41,15 @@ def test_0_predict_file_flow(self):
...
         predictions = predictor.predict(df)
 
+        tests = predictor.test(df, [r2_score])
+
         # sanity checks
         self.assertTrue(r2_score([float(x.rstrip('$')) for x in df[target]], predictions['prediction']) > 0.8)
         self.assertTrue(all([0 <= p <= 1 for p in predictions['confidence']]))
         self.assertTrue(all([p['lower'] <= p['prediction'] <= p['upper'] for _, p in predictions.iterrows()]))
 
+        self.assertTrue(tests['r2_score'].dtypes == 'float64')
```
paxcema, line 51: This check doesn't seem correct. `r2_score` would be the value in the `Metric` column, not the actual name of the column.

lightwood/api/json_ai.py
```diff
@@ -1180,6 +1180,15 @@ def code_from_json_ai(json_ai: JsonAI) -> str:
 
     predict_body = align(predict_body, 2)
 
+    test_body = """
+predictions = self.predict(data)
+values = [metric(data[self.target], predictions) for metric in metrics]
```
paxcema, line 1185: `predictions` is a DataFrame that can potentially have many columns. You should pass only the actual predicted values that are contained in the `prediction` column.

ErdoganSeref:
```suggestion
values = [metric(data[self.target], predictions[self.target]) for metric in metrics]
```

tests/integration/basic/test_regression.py
ErdoganSeref:
```suggestion
        self.assertTrue(tests['Value'].dtypes == 'float64')
```

PR #1032, opened 2022-10-25
Simpler TS evaluation

paxcema:
## Why
Improves UX in forecasting scenarios by using a more streamlined accuracy function and mixer evaluation procedure.

## How
- New evaluation protocol for TS predictors: now only the first forecast horizon is considered for accuracy computation. If there are multiple series (i.e. groups), the first data point for each is used.
- New complementary sMAPE metric is now the default metric in forecasting use cases.


tomhuds:
Nice work dude!

PR #1035, opened 2022-10-30
Bug fix: parameters to be optimized

jaredc07:
Parameter optimization is not effective.


jaredc07:
> Looks good! Essentially, this should mean shorter runtimes, correct?

No. It was my fault~. The parameter optimization did not take effect.
`self.model.set_params(**params) `  
`y_pred = cross_val_predict(self.model, X, Y, cv=self.cv, method=predict_method)`
Although the parameters of the model have changed, the model has not been updated.
I forgot to retrain the model, it's updated now.

PR #1044, opened 2022-11-11
MAvg analysis block

paxcema:


lightwood/analysis/helpers/mavg.py
```diff
@@ -0,0 +1,66 @@
...
+from lightwood.analysis.base import BaseAnalysisBlock
+from lightwood.helpers.log import log
+
+
+class MAvg(BaseAnalysisBlock):
+    """
+    Wrapper analysis block for the 'Robbie'...
+    """
+
+    def __init__(self, deps: Optional[Tuple] = ()):
```
George3d6, line 17: Probably intended as `Tuple` not `Optional[Tuple]` given the default arg

lightwood/analysis/helpers/mavg.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__(deps=deps)
+
+    def analyze(self, info: Dict[str, object], **kwargs) -> Dict[str, object]:
```
George3d6, line 20: fun fact, you can now use `dict` here instead of `Dict` (same for `list` and `tuple` instead of `List` and `Tuple`) thus getting rid of the typing imports, most lang servers will support the new type hints

George3d6 (commented):
sorry, was looking around out of curiosity and couldn't resist leaving a small nitpick :p

lightwood/analysis/helpers/mavg.py
paxcema: I'm still unsure whether there's any benefit at all with using these `typing` aliases to the actual type or not. I did some reading on what you mention here and it appears to be supported starting 3.9, so I'm tempted to keep it like this for a bit longer until we move out of 3.7 and 3.8.

paxcema:
Closing this, end model for customer is way too specific to be broadly useful for other Lightwood users. Will import as custom module instead :)

PR #1118, opened 2023-02-22
added supported target datatypes

Sumanth077:
Hi, @paxcema hope you are doing well. Sorry for my delay in working on this issue.

Let's start with the Random Forest Mixer. If everything works well then we can work on the other mixers as well.

I have added the supported target data types in the "base mixer" and also an overridden for that in the Specified Random Forest Class. Kindly take a look at that and let me know if that's good.

Then as a next step, I will work on adding an informative error.

### DIFF
--- a/lightwood/mixer/base.py
+++ b/lightwood/mixer/base.py
@@ -3,6 +3,7 @@ import pandas as pd
 
 from lightwood.data.encoded_ds import EncodedDs
 from lightwood.api.types import PredictionArguments
+from type_infer.dtype import dtype
 
 
 class BaseMixer:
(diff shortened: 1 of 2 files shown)


paxcema:
Hey @Sumanth077, thanks for starting to work on this!

Looking at the code, I think it's generally the right idea. I would be inclined to actually go the other way around and define `self.supported_target_datatypes = None` for `BaseMixer`, so that new mixer contributors actually have to go and think hard about which dtypes are supported or not.

Also, be sure to merge `staging` into this branch, as there have been changes in the base class since.

paxcema:
@Sumanth077 any updates on this?

PR #1130, opened 2023-03-29
deps: Torch 2.0

paxcema:


paxcema:
Closing for now, we either need:

- torch to release 3.7 wheels
- mindsdb to drop support for 3.7

paxcema:
@StpMax reopening so that we can discuss the second approach listed above. I will also look into supporting 3.10, instead.

PR #1131, opened 2023-04-18
update save method on PredictorInterface class: lightwood/lightwood/api/predictor.py[lines: 141-206]

### LINKED ISSUE #852: Add `export` method that saves both `Predictor` object and its code.
Respectively, `import` should load the predictor+code.

noraa-july-stoke:
update save method on PredictorInterface class: lightwood/lightwood/api/predictor.py[lines: 141-206]

The function now optionally takes in a string that that user has gotten from calling code_from_json_ai()
and saves it to a dict. @paxcema when I initially was typing my idea out in the issue thread 
I wasn't sure what the datatype being returned was for code_from_json_ai() but I poked around
and found it's a string so can we just save it like that?

See contributor comments above modification... 
I left some of my thought processes in there so you could follow my logic.
(I'm still figuring out the code base). 

closes #852


noraa-july-stoke:
I have read the CLA Document and I hereby sign the CLA

lightwood/api/predictor.py
```diff
@@ -136,13 +136,71 @@ def predict(self, data: pd.DataFrame, args: Dict[str, object] = {}) -> pd.DataFr
...
         pass
 
-    def save(self, file_path: str) -> None:
+
+
+        """
+            NOTE FROM CONTRIBUTOR - will remove before finalizing
+
+            I spotted this code snippet in the custom_cleaner.ipynb file
+            I figured we could maybe have the user pass it optionally
```
paxcema, line 145: Let's go for an entirely method `export()`. Changing the `save()` method at this point in time is something we'd rather avoid, even if new arguments are optional.

lightwood/api/predictor.py
```diff
@@ -136,13 +136,71 @@ def predict(self, data: pd.DataFrame, args: Dict[str, object] = {}) -> pd.DataFr
...
+    def save(self, file_path: str, json_ai_code: Optional[str] = None) -> None:
         """
         With a provided file path, saves the Predictor instance for later use.
 
         :param file_path: Location to store your Predictor Instance.
+        :param code: The code generated by the user's specification (optional)
 
         :returns: Saves Predictor instance.
         """
+        predictor_dict = {}
```
paxcema, line 199: This idea is overall on the right track.

I think the one missing bit is to actually produce the json_ai internally, otherwise there is a risk in having a mismatch between the passed json_ai and what `self` actually looks like. This is a change that requires a bit more tinkering though, so let's leave it for a second PR.

paxcema (changes requested):
Generally looks good to me! Have just asked to instead move the new logic into a separate `export()` method. Once that's done, let's merge and work on producing the JsonAI internally as a follow-up.

lightwood/api/predictor.py
noraa-july-stoke: you got it! I'll get this done soon.

noraa-july-stoke:
I removed the optional bit from the json_ai_code since
        we changed it to a new function altogether. I should have
        done that in the first place 😅.  I can maybe
        look into producing the AI internally when I finish doing that
        fitbit integration I'm working on.

lightwood/api/predictor.py
```diff
@@ -139,10 +139,25 @@ def predict(self, data: pd.DataFrame, args: Dict[str, object] = {}) -> pd.DataFr
...
         With a provided file path, saves the Predictor instance for later use.
+        :param file_path: Location to store your Predictor Instance.
+        :returns: Saves Predictor instance.
+        """
+        with open(file_path, "wb") as fp:
+            dill.dump(self, fp)
+
+    def export(self, file_path: str, json_ai_code: str) -> None:
+        """
+        With a provided file path, saves the Predictor instance for later use.
```
paxcema, line 150: Suggestion: 

> Exports both the predictor object and its code to a single binary file for later usage.

lightwood/api/predictor.py
```diff
@@ -139,10 +139,25 @@ def predict(self, data: pd.DataFrame, args: Dict[str, object] = {}) -> pd.DataFr
...
 
         :param file_path: Location to store your Predictor Instance.
+        :param json_ai_code: The code generated by the user's specification (optional)
```
paxcema, line 153: Minor nitpick: right now `json_ai_code` is not optional 😁

paxcema (approved):
Left a couple very minor observations, but I'd say we're good to go.

lightwood/api/predictor.py
noraa-july-stoke: Haha sorry! I took it out BC I was doing a new method entirely so I didn’t know if having that was pointless anymore. I’ll put those back in and send another PR this morning!!

noraa-july-stoke: 🤦‍♀️ oops! fixing that now

paxcema (approved):

paxcema:
Thanks for your contribution @noraa-july-stoke 🙌

PR #1145, opened 2023-06-07
[ENH] Top level cache for Predictor.featurize()

paxcema:
## Changelog
- `RandomForestMixer`: reduce number of hyperparameter search trials from 20 to 5
- Rename `filter_ds` to `filter_ts`. Change signature and usage so that it directly interacts with `pd.DataFrame`s instead.
- Setting up a cache in the generated `Predictor` code that maintains three `EncodedDs` objects through the `learn()` pipeline, deleting them at the end to avoid bloated binaries. This helps reduce the number of instantiations, as all invocations are done through `featurize()`, which checks (and potentially uses) the cache first.
- Avoid duplicate instantiation of validation `EncodedDs` object inside `analysis`, reverting to use the predictor's cache instead.

## Effect:
This reduces `learn()` runtime anywhere from 15% to 50% depending on the dataset and predictor properties.

### DIFF
--- a/lightwood/analysis/analyze.py
+++ b/lightwood/analysis/analyze.py
@@ -3,7 +3,6 @@ from typing import Dict, List, Tuple, Optional
 from dataprep_ml import StatisticalAnalysis
 
 from lightwood.helpers.log import log
-from lightwood.helpers.ts import filter_ds
 from type_infer.dtype import dtype
 from lightwood.ensemble import BaseEnsemble
 from lightwood.analysis.base import BaseAnalysisBlock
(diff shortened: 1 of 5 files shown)


pafluxa:
Greetings,

I can confirm that this PR speeds-up the learning by at least 15% in all test datasets. 

The speed-up is less noticeable when dealing with larger datasets (>10K rows) and seems to decrease non-linearly with the number of rows used. In particular, the gain drops from more than 20% to barely above 10% when scaling the number of rows from 10000 to 100000. I interpret this as an indication that the performance issues when dealing with large datasets is not entirely solved by this PR. 

Finally, judging by the results of the unit tests (nicely done!) the changes that were implemented seem to break some of the functionality used for time series analysis/models.

paxcema:
Great! I'll get the tests passing and merge.

PR #1146, opened 2023-06-07
[ENH] faster pretrained text encoder

paxcema:
Note: this builds on top of #1145 so the PR is set against that branch. However, it should be merged into `staging` eventually!

## Changelog
Refactored `PretrainedTextEncoder` training loop to introduce a few things that could help with getting a better runtime-accuracy tradeoff by default:
1. Create and use a validation split.
2. Early stopping strategies based on validation and training loss.
3. Replaced a few operations for vectorized equivalents.

## Effect
Tests indicate a further 15% improvement compared to #1145 in text classification tasks that use the UnitMixer

### DIFF
--- a/lightwood/encoder/text/pretrained.py
+++ b/lightwood/encoder/text/pretrained.py
@@ -1,15 +1,12 @@
-"""
-"""
+import os
 import time
+from typing import Iterable
+from collections import deque
+
+import numpy as np
 import torch
 from torch.utils.data import DataLoader
-import os
 import pandas as pd
-from lightwood.encoder.text.helpers.pretrained_helpers import TextEmbed
-from lightwood.helpers.device import get_device_from_name
-from lightwood.encoder.base import BaseEncoder
-from lightwood.helpers.log import log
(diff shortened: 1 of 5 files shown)


lightwood/encoder/text/pretrained.py
```diff
@@ -90,46 +94,45 @@ def __init__(
...
         encoded_target_values: torch.Tensor,
     ):
         """
         Fine-tunes a transformer on the priming data.
 
-        CURRENTLY WIP; train + dev are placeholders for a validation-based approach. 
-
-        Train + Dev are concatenated together and a transformer is then fine tuned with weight-decay applied on the transformer parameters. The option to freeze the underlying transformer and only train a linear layer exists if `frozen=True`. This trains faster, with the exception that the performance is often lower than fine-tuning on internal benchmarks.
+        Transformer is fine-tuned with weight-decay on training split. 
+        By default, underlying transformer is frozen and only final linear layer is trained. This trains faster, often as tradeoff for performance.
```
paxcema, line 105: Revert this, as `frozen` default value ended up being unchanged instead.

PR #1155, opened 2023-06-21
[fix] Optimized datetime encoder

### LINKED ISSUE #891: Potential bug in datetime encoded value sensitivity
As reported by a community user:

If a LightGBMArray mixer is trained to forecast, querying conditionally and changing the timestamp may not make a difference in the prediction. We should double check that the encoded value is being used.

paxcema:
Fixes #891.

Additionally, this encoder is now able to end-to-end encode and decode 20 million rows in less than 40 seconds with consumer hardware.

### DIFF
--- a/lightwood/data/timeseries_transform.py
+++ b/lightwood/data/timeseries_transform.py
@@ -75,7 +75,7 @@ def transform_timeseries(
                     raise Exception(
                         f"Partition is not valid, faulty group {group}. Please make sure you group by a set of columns that ensures unique measurements for each grouping through time.")  # noqa
 
-                index = pd.to_datetime(subset[oby_col], unit='s')
+                index = pd.to_datetime(subset[oby_col], unit='s', utc=True)
                 subset.index = pd.date_range(start=index.iloc[0],
                                              freq=freqs.get(group, freqs['__default']),
                                              periods=len(subset))
--- a/lightwood/encoder/datetime/datetime.py
+++ b/lightwood/encoder/datetime/datetime.py
@@ -1,4 +1,3 @@
-import datetime
 from typing import Union
 
 import torch
@@ -40,9 +39,8 @@ class DatetimeEncoder(BaseEncoder):
         if isinstance(data, np.ndarray):
             data = pd.Series(data)
 
-        data = data.fillna(pd.Timestamp.max.timestamp())  # TODO: replace with mean once encoded?
-
-        ret = [data.apply(datetime.datetime.fromtimestamp)]
+        data = data.fillna(pd.Timestamp.max.timestamp())  # TODO: replace with mean?
+        ret = [pd.to_datetime(data, unit='s', origin=-1, utc=True)]
         for i, attr in enumerate(self.constant_keys):
             def _get_ts_attr(ts):
                 return getattr(ts, attr)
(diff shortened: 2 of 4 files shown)


lightwood/data/timeseries_transform.py
```diff
@@ -75,7 +75,7 @@ def transform_timeseries(
                     raise Exception(
                         f"Partition is not valid, faulty group {group}. Please make sure you group by a set of columns that ensures unique measurements for each grouping through time.")  # noqa
 
-                index = pd.to_datetime(subset[oby_col], unit='s')
+                index = pd.to_datetime(subset[oby_col], unit='s', utc=True)
```
pafluxa, line 78: UTC is the only time :D 

There might be an edge case where leap seconds don't get properly handled (https://en.wikipedia.org/wiki/Leap_second) but this is very, very unlikely to cause trouble; it did cause trouble to me when working with astronomical data so it might be worth to keep a mental post-it about it.

lightwood/encoder/datetime/datetime.py
```diff
@@ -1,90 +1,82 @@
...
-        return vector
-
-    def decode(self, encoded_data, return_as_datetime=False):
-        ret = []
+        if type(data) not in (np.ndarray, pd.Series):
+            raise Exception(f'Data should be pd.Series or np.ndarray! Got: {type(data)}')
+        if isinstance(data, np.ndarray):
+            data = pd.Series(data)
+
+        data = data.fillna(pd.Timestamp.max.timestamp())  # TODO: replace with mean?
```
pafluxa, line 42: I would an actual interpolation step using a linear spline here, as nothing guarantees that the timestamps are evenly sampled (i.e., the frequency of data acquisition is constant). fillna() is somewhat imprecise and I've seen it producing wrong results when I worked with time series data.

lightwood/encoder/datetime/datetime.py
```diff
@@ -1,90 +1,82 @@
...
-            raise Exception('You need to call "prepare" before calling "encode" or "decode".')
-
-        ret = [self.encode_one(unix_timestamp) for unix_timestamp in data]
-
-        return torch.Tensor(ret)
+    def prepare(self, priming_data):
+        self.is_prepared = True
 
-    def encode_one(self, unix_timestamp: Optional[float]):
+    def encode(self, data: Union[np.ndarray, pd.Series]) -> torch.Tensor:
```
pafluxa, line 32: I like what's being done here, very nice!

My only concern is that that I can see the output tensor being somewhat discontinuous: for instance, when the weekday changes, you will suddenly go from an entry with 'weekday = 6' to 'weekday = 0'; I am not entirely sure this will help the ML side of things (more inclined to think it will not).

Something that worked for me is to use trigonometric encoders for the time. Basically, every component (year, month, day, weekday, hour, mine, second) is encoded with a sine and a cosine function with a period equal to the component's period (365 days per year, 7 days a week, 24 hours a day, etc). This not only puts the numbers in the [-1, 1] range (very nice for ML) but it also avoids nasty discontinuity artifacts, like changing from weekday 6 to 0 from one step to the next one. 

Might be worth giving this a thought if the encoder is not performing as expected ;-)

lightwood/encoder/datetime/datetime.py
pafluxa: fillna() can be used to perform interpolation, by the way, but since we do have access to the "raw" unix timestamps there is no need for it. 

Also, please note that in some cases we will actually _want to keep_ the NaNs around to find gaps in the data.

tests/unit_tests/encoder/date/test_datetime.py
```diff
@@ -1,30 +1,56 @@
...
+            if pd.Timestamp.min.year + 1 < datetime.fromtimestamp(t).year < pd.Timestamp.max.year - 1:
+                if not np.isclose(d, t, atol=5):
+                    assert datetime.fromtimestamp(d) == datetime.fromtimestamp(t)
+                else:
+                    assert np.isclose(d, t, atol=5)
+        assert np.isnan(dec_data[0])
 
+    def test_sinusoidal_encoding(self):
+        data = [self._create_timestamp() for _ in range(100)]
         normalizer = DatetimeNormalizerEncoder(sinusoidal=True)
```
pafluxa, line 54: I see that there is a trigonometric encoder already! Nice. Probably worth to keep this as a default in Lightwood :D

pafluxa (approved):
Looks good to me, I could not spot any obvious errors. My suggestions/notes are mainly for future reference if the spotted "can-be-an-issue-but-probably-not" lines become a problem some day.

I approve this PR.

lightwood/encoder/datetime/datetime.py
paxcema: Great idea! I'll add it as a separate issue because it seems quite important to handle this in a better way.

lightwood/encoder/datetime/datetime.py
paxcema: Yes, I totally agree, we have the `datetime_sin_encoder.py` for this, albeit not as a default option. It's next in line to be vectorized, so more on that later. As for improvements w.r.t. this one, I don't think we've actually ran the benchmark suite on it, would be a nice next step!

tests/unit_tests/encoder/date/test_datetime.py
paxcema: Will add an issue to actually benchmark this and change if the impact is positive (which I agree it _should_)

paxcema: See: #1157

PR #1156, opened 2023-06-22
[ENH] Faster TS transforms

### LINKED ISSUE #1057: Avoid lists as cells in TS transformations
We should avoid nesting arrays inside cells, as it's bad Pandas practice and leads to suboptimal performance.

Instead, we should handle these transformations as separate rows (ideally virtualized to save memory consumption, but that's another issue). If any specific mixer requires this info. to actually be on a single cell, then that transformation should be within its own jurisdiction.

paxcema:
~Note: should be merged into staging, but will compare against `fix_891` while in draft.~

Fixes #1057.

## Changelog
- Removed `get_group_matches` 
- A bunch of methods have been refactored to comply with a faster `df.groupby()`-based approach instead
- Removed `max_pacf()` method for finding maximum coefficients for partial autocorrelation. Was quite slow and had minimal impact on accuracy based on benchmarks
- Removed STL usage in `LightGBM` and `NeuralTS` mixers plus associated helper methods

## Benchmarks
Noticeable runtime speedups, between 2.5x and 16x depending on dataset size and amount of group combinations (if applicable). This is expected given the reduced computational complexity of all modified algorithms.


lightwood/mixer/sktime.py
```diff
@@ -71,7 +69,7 @@ def __init__(
         :param use_stl: Whether to use de-trenders and de-seasonalizers fitted in the timeseries analysis phase.
         """  # noqa
         super().__init__(stop_after)
-        self.stable = False
+        self.stable = True
```
paxcema, line 72: Revert

lightwood/helpers/ts.py
```diff
@@ -15,52 +14,17 @@ def get_ts_groups(df: pd.DataFrame, tss) -> list:
...
     :param df: Dataframe with time series data.
-    :param group_combinations: all tuples with distinct values for `TimeseriesSettings.group_by` columns, defining all available time series.
-    :param target: name of target column
     :param tss: timeseries settings
 
     :return:
     Dictionary with group combination tuples as keys. Values are dictionaries with the inferred delta for each series.
     """  # noqa
-    df = df.copy()
+    df = df.copy()  # TODO: necessary?
```
paxcema, line 27: Double check

lightwood/data/timeseries_transform.py
```diff
@@ -51,37 +50,32 @@ def transform_timeseries(
         if hcol not in data.columns or data[hcol].isna().any():
             raise Exception(f"Cannot transform. Missing values in historical column {hcol}.")
 
-    # infer frequency with get_delta
-    oby_col = tss.order_by
-    groups = get_ts_groups(data, tss)
-
     # initial stable sort and per-partition deduplication
     data = data.sort_values(by=oby_col, kind='mergesort')
+    # TODO evaluate if truly needed (w/o it, test 9 fails)
```
paxcema, line 55: Double check

lightwood/mixer/sktime.py
```diff
@@ -151,12 +150,14 @@ def _fit(self, data):
...
-        for group in self.ts_analysis['group_combinations']:
+        grouped = df.groupby(by=gby) if gby else df.groupby(lambda x: '__default')
+        for group, series_data in grouped:
             kwargs = {}
-            sp = self.sp if self.sp else self.ts_analysis['periods'].get(group, '__default')[0]
+            sp = self.sp if self.sp else self.ts_analysis['periods'].get(group, [1])[0]
 
             options = self.model_kwargs
-            options['sp'] = sp                   # seasonality period
+            if sp > 1:
```
paxcema, line 159: Double check if `if` is needed here

PR #1163, opened 2023-06-26
[ENH] Embedding mode

paxcema:
This feature enables training a Lightwood predictor without any mixers or subsequent analysis if the `embedding_only` flag is enabled in the problem definition. The resulting output is the concatenation of encoded representations learnt by each encoder, a.k.a. the "embedding".

Alternatively, normally trained predictors can also return these "embeddings" if the `return_embedding` argument is set to `True`.


lightwood/api/types.py
```diff
@@ -220,6 +221,7 @@ def from_dict(obj: Dict):
         ignore_features = obj.get('ignore_features', [])
         fit_on_all = obj.get('fit_on_all', True)
         use_default_analysis = obj.get('use_default_analysis', True)
+        embedding_only = obj.get('embedding_only', False)
```
paxcema, line 224: Thoughts on the parameter name @torrmal?

lightwood/api/types.py
```diff
@@ -453,6 +456,7 @@ class PredictionArguments:
     simple_ts_bounds: bool = False
     time_format: str = ''
     force_ts_infer: bool = False
+    return_embedding: bool = False
```
paxcema, line 459: Thoughts on the parameter name @torrmal?

PR #1164, opened 2023-06-27
Improve `transform_timeseries`

StpMax:
Apply `.astype` to part of array is much slower, then to full array. This little change fix it. Depend on count of groups, speedup of model learn in % is about:
```
+------+-------+------+
|  50  |  100  | 500  |
+------+-------+------+
| 53.8 | 7.77  | 4.88 |
+------+-------+------+
```

### DIFF
--- a/lightwood/data/timeseries_transform.py
+++ b/lightwood/data/timeseries_transform.py
@@ -5,14 +5,12 @@ import multiprocessing as mp
 import numpy as np
 import pandas as pd
 from lightwood.helpers.parallelism import get_nr_procs
-import mindsdb.utilities.profiler as profiler
 
 from type_infer.dtype import dtype
 from lightwood.api.types import TimeseriesSettings, PredictionArguments
 from lightwood.helpers.log import log
 
 
-@profiler.profile()
 def transform_timeseries(
         data: pd.DataFrame, dtype_dict: Dict[str, str],
         timeseries_settings: TimeseriesSettings, target: str, mode: str,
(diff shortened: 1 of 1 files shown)


lightwood/data/timeseries_transform.py
```diff
@@ -136,30 +137,29 @@ def transform_timeseries(
...
 
-    if len(original_df) > 500:
+    if len(df_arr) > 4 and len(original_df) > 5000:
         # @TODO: restore possibility to override this with args
-        nr_procs = get_nr_procs(original_df)
+        nr_procs = min(get_nr_procs(original_df), len(original_df))
         log.info(f'Using {nr_procs} processes to reshape.')
-        pool = mp.Pool(processes=nr_procs)
-        # Make type `object` so that dataframe cells can be python lists
-        df_arr = pool.map(partial(_ts_to_obj, historical_columns=[oby] + tss.historical_columns), df_arr)
```
paxcema, line 145: If we're OK without using `_ts_to_obj`, then let's remove its source code, too 👍

paxcema: Nevermind, missed the usage outside of the conditional.

lightwood/data/timeseries_transform.py
```diff
@@ -109,6 +109,7 @@ def transform_timeseries(
         secondary_type_dict[oby] = dtype_dict[oby]
 
     original_df[f'__mdb_original_{oby}'] = original_df[oby]
+    original_df = _ts_to_obj(original_df, [oby])
```
paxcema, line 112: I think the second argument should be `[oby] + tss.historical_columns`, where the latter is a list by default, so adding both should always work.

paxcema (changes requested):

PR #1168, opened 2023-06-27
[ENH] Various improvements

paxcema:
## Faster categorical autoencoder

New approach based on a `SimpleLabelEncoder` (more efficient in terms of training speed and memory consumption) is automatically triggered past a certain amount of observed labels. 

While less accurate, this enables handling of large datasets with categorical features that have huge cardinality.

## Improved templates for ensembles
Ensembles don't need to strictly follow the base signature now. Instead, `JsonAI` inspects the subclass' arguments and adds only the relevant ones to the generated code.

## Identity Ensemble
Introduces a new `IdentityEnsemble` that performs no additional operations apart from storing and calling mixers. Ideal for high performance use cases where a single mixer is used.

## Faster analysis for grouped time series forecasters
This phase now considers the first forecast for every group, which has the added bonus of keeping the amount of known target values rather high, compared to the current approach that may be reporting pessimistic metrics.


lightwood/api/predictor.py
```diff
@@ -145,6 +148,22 @@ def save(self, file_path: str) -> None:
         with open(file_path, "wb") as fp:
             dill.dump(self, fp)
 
+    def __getstate2__(self):
```
paxcema, line 151: remove

lightwood/api/predictor.py
```diff
@@ -4,6 +4,9 @@
 import pandas as pd
 from lightwood.api.types import ModelAnalysis
 
+# these imports need special handling in this class
+from lightwood.mixer import NHitsMixer
```
paxcema, line 8: remove

PR #1178, opened 2023-07-18
Support for hdf5 when featurizing

paxcema:


paxcema:
Closing for now, we've decided Sigmoid will be its own ML engine so there's no need to natively support HDF5 in Lightwood at this time.

PR #1186, opened 2023-08-09
fix args passing to mixers

StpMax:


paxcema:
Closing as won't fix, bug is handled on the mindsdb side.

Issue #1190, opened 2023-09-08
BestOf ensemble: option to specify whether reported metric should be minimized or maximized


paxcema:

paxcema:
https://docs.google.com/document/d/1sqs7qpcwcbThKwJpzackBc6fcrou-Ei5Wfz5JT8A7fE/edit

PR #1191, opened 2023-09-08
[ENH] New ts defaults

paxcema:
Changelog:

- Disable sktime and all related mixers.
- Introduce an array variant for the XGBoost mixer: `XGBoostArrayMixer`.
- Add `XGBoostArrayMixer` as default mixer. Initial tests suggest runtime should improve without too much loss in accuracy (even then, no actual config space is lost for power users, just default configs change)

### DIFF
--- a/lightwood/ensemble/best_of.py
+++ b/lightwood/ensemble/best_of.py
@@ -57,7 +57,7 @@ class BestOf(BaseEnsemble):
 
                 score_list.append(avg_score)
 
-            self.indexes_by_accuracy = list(np.array(score_list).argsort())
+            self.indexes_by_accuracy = list(reversed(np.array(score_list).argsort()))
             self.supports_proba = self.mixers[self.indexes_by_accuracy[0]].supports_proba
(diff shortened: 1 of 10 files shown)


lightwood/ensemble/best_of.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, target, mixers: List[BaseMixer], data: EncodedDs, accuracy_fu
 
                 score_list.append(avg_score)
 
-            self.indexes_by_accuracy = list(reversed(np.array(score_list).argsort()))
+            self.indexes_by_accuracy = list(np.array(score_list).argsort())
```
paxcema, line 60: restore, add as customizable feature (see #1190)

tests/integration/advanced/test_timeseries.py
```diff
@@ -475,7 +475,7 @@ def test_8_time_series_double_grouped_regression(self):
         target = 'MA'
         order_by = 'saledate'
         window = 8
-        for horizon in [1, 4]:
+        for horizon in [4]:
```
paxcema, line 478: restore

PR #1196, opened 2023-10-24
Merged lightwood & release.yml

ddninja:
Merged the lightwood & release.yml, now deploy job in the lightwood yaml will also be executed if the release is published from any other branch.


paxcema (approved):
Amazing work @ddninja, thanks for your contribution!

PR #1199, opened 2023-12-12
Include extras requirements files

hamishfagg:
Fixes issues with torchvision missing from mindsdb docker files

Extras requirements files are not included in the lightwood pip package, so setup.py can't enumerate them, and when users install e.g. `lightwood[all_extras]` there are packages missing


paxcema (approved):
LGTM, good catch

PR #1232, opened 2024-11-13
release 24.5.2.1

dusvyat:


MinuraPunchihewa:
@dusvyat It looks like the workflow for running the benchmarks given here are failing. Do you have an idea of how this works? It looks like the MySQL database used here no longer exists?

ZoranPandovski:
As agreed we release this without the results of the benchmarks and then fix that workflow for the next release

ZoranPandovski (approved):

